PR #2, opened 2014-11-17
Implement diagnostic SA1500 CurlyBracketsForMultiLineStatementsMustNotShareLine

### LINKED ISSUE #101: SA1500: CurlyBracketsForMultiLineStatementsMustNotShareLine
http://www.stylecop.com/docs/SA1500.html

sharwell:
Fixes #101

This implementation is nearly complete, but I'm observing a high rate of reporting for the following situations.

* Lambda expressions with a body

    ```csharp
    Func<int> example =
      () =>
      {
        return 3;
      };
    ```

* `do`/`while` statements

    ```csharp
    do
    {
    } while (expression);
    ```


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CurlyBracketsForMultiLineStatementsMustNotShareLine.cs
```diff
@@ -73,7 +76,82 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, location));
+                    // no need to report more than one instance for this token
+                    return;
+                }
+            }
+
+            SyntaxToken nextToken = token.GetNextToken();
+            if (!nextToken.IsMissing)
+            {
+                if (nextToken.GetLocation().GetLineSpan().StartLinePosition.Line == line)
```
pdelvo: This will produce SA1503.

pdelvo:
Now that @vweijsters implemented unit tests I think this one can be merged.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CurlyBracketsForMultiLineStatementsMustNotShareLine.cs
sharwell: I'm waiting for the code fix for this to be part of the project before I really worry about it (it'll be included when we switch to Alpha 5).

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1500/SA1500UnitTests.Blocks.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+        }
+
+        // valid block #2
+        {
+            Debug.Indent();
+        }
+
+        // valid block #3 (valid only for SA1500)
+        { }
```
sharwell: @vweijsters What diagnostic gets reported instead for this case?

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1500/SA1500UnitTests.Blocks.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        // valid block #4 (valid only for SA1500)
+        { Debug.Indent(); }
```
sharwell: :question: @vweijsters What diagnostic gets reported instead for this case?

sharwell:
It looks like I misinterpreted this diagnostic. I was looking at all curly braces instead of just the ones associated with multi-line statements (`if`, `while`, etc.).

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1500/SA1500UnitTests.Blocks.cs
vweijsters: SA1501

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1500/SA1500UnitTests.Blocks.cs
vweijsters: SA1501

sharwell:
I think I finally figured out most of the rules. I'm working on updating the code.

@vweijsters I noticed that you have several test cases to ensure only one diagnostic is reported per line. I think it's fine to only report one diagnostic for a matched part of `{` and `}` if they appear on the same line, but for other cases (e.g. two separate `}` tokens on the same line) they should not suppress the second message.

sharwell:
@vweijsters I updated this pull request. Please review my changes to the expected test suite in eae4704176ecc93198fa0b722552674b473b8aa4 and ad49f77b962331d367864fe23933496b4adb1aca.

pdelvo:
Otherwise this looks good :+1:

PR #3, opened 2014-11-17
Implement diagnostic SA1300 ElementMustBeginWithUpperCaseLetter

### LINKED ISSUE #76: SA1300: ElementMustBeginWithUpperCaseLetter
http://www.stylecop.com/docs/SA1300.html

sharwell:
Fixes #76

It's hard to tell if the rules used by this implementation are an exact match to the original StyleCop rule. In particular, the following are open questions.

1. :question: What about a class named `fooNativeMethods`? Current behavior: warning unless `fooNativeMethods` is inside a containing `NativeMethods` class.
2. :question: Are private methods included or excluded from analysis? Current behavior: included.
3. :question: ~~How does the `protected internal` accessibility apply to analysis of fields? Current behavior: not considered "internal", and not considered "private".~~ Update: Fields are handled by SA1303-SA1311.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1300UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1300UnitTests.cs
@@ -412,22 +412,8 @@ readonly string test;
 public string test;
 }";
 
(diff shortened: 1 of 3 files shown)


pdelvo:
I think the current behavior is the one we should go for.

PR #261, opened 2014-11-23
Fix errors in firstInLine calculations

### LINKED ISSUE #257: SA1001 incorrectly reported at beginning of line
The diagnostic for SA1001 does not correctly determine whether a token is the first in the line or not. The following code produces an incorrect SA1001 warning for the `,` token. Note that this is only reproduced when the code is not indented.

```csharp
int[] data = { 3
, 2 };
```

### LINKED ISSUE #258: SA1018 incorrectly reported at beginning of line
The diagnostic for SA1018 does not correctly determine whether a token is the first in the line or not. The following code produces an incorrect SA1018 warning for the `?` token. Note that this is only reproduced when the code is not indented.

```csharp
int
? data;
```

sharwell:
Fixes #257 
Fixes #258

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1001CommasMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1001CommasMustBeSpacedCorrectly.cs
@@ -87,9 +87,9 @@
             }
 
             bool hasPrecedingSpace = false;
-            if (!token.HasLeadingTrivia)
+            bool firstInLine = token.HasLeadingTrivia || token.GetLocation()?.GetMappedLineSpan().StartLinePosition.Character == 0;
+            if (!firstInLine)
             {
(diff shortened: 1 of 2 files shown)


heejaechang:
Finding out whether a token is a first token on a line only using Token is not simple since leading and trailing trivia could contain structured trivia, and end of line (or other kinds of trivia that could contain new line) trivia might belong to them rather than regular token.

if you have SourceText, just use previous Token's end position and current token's start position to get line number and compare them to see whether two are on different line.

also, when you say, "token", you need to distinguish whether it only means a normal token or it includes token that belong to structured trivia such as xml document, skipped tokens and etc.

Simplest way is probably using SourceText, otherwise, you probably need to traverse down to every trivia (including trivia inside of structured trivia) between two tokens and see any one of them (since there can be multiple different kinds of trivia that can have EOL) contains new line.

other easiest way (without SourceText) will be go through all trivia between two tokens and then just use ToString to get "string representation" of the trivia and do text search.

PR #306, opened 2014-12-05
Type alias diagnostics

### LINKED ISSUE #54: SA1121: UseBuiltInTypeAlias
http://www.stylecop.com/docs/SA1121.html

### LINKED ISSUE #58: SA1125: UseShorthandForNullableTypes
http://www.stylecop.com/docs/SA1125.html

sharwell:
Fixes #54 
Fixes #58


pdelvo:
The unit tests I implemented in #350 don't report any problems for SA1121 (if the messages are changed to match the ones used here). So I think merging this diagnostic is fine. I don't see any problems in the diagnostic for SA1125 as well but I don't have any tests yet.

PR #327, opened 2014-12-05
Field names diagnostics

### LINKED ISSUE #84: SA1308: VariableNamesMustNotBePrefixed
http://www.stylecop.com/docs/SA1308.html

### LINKED ISSUE #85: SA1309: FieldNamesMustNotBeginWithUnderscore
http://www.stylecop.com/docs/SA1309.html

### LINKED ISSUE #86: SA1310: FieldNamesMustNotContainUnderscore
http://www.stylecop.com/docs/SA1310.html

sharwell:
Fixes #84 
Fixes #85 
Fixes #86

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
 
     internal static class NamedTypeHelpers
     {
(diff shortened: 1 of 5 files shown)


sharwell:
This is blocked on #358.

sharwell:
The issues mentioned above have now been fixed.

@macpak Since this affects the API for your `NamedTypeHelpers` class, can you help review it?

macpak:
Yes, I will do it but probably not today (but maybe).

macpak:
Code looks ok.
I am just wondering if analyzing FieldDeclarationSyntax and "going down" would give us better performance than this solution. The reason is that subscribing to VariableDeclaratorSyntax "catches" all declarations of (for instance) local variables so we actually analyzing redundant nodes.

sharwell:
@macpak I think you made a better point than you even realized. With this code, SA1309 is reported for `_x`:

```csharp
class Class1
{
    int bar =
        ((Func<int>)(() =>
        {
			int _x;
            return 0;
        }))();
}
```

sharwell:
It also fails to report SA1309 for this:

```csharp
int @\u005Fbar2 = 0;
```

sharwell:
I just updated this:

* Used `FieldDeclarationSyntax` instead of `VariableDeclaratorSyntax`
* Used `SyntaxToken.ValueText` instead of `SyntaxToken.Text` for accurate analysis of Unicode identifiers and the `@` prefix

PR #333, opened 2014-12-19
Implement diagnostic SA1100 DoNotPrefixCallsWithBaseUnlessLocalImplementationExists

### LINKED ISSUE #34: SA1100: DoNotPrefixCallsWithBaseUnlessLocalImplementationExists
http://www.stylecop.com/docs/SA1100.html

macpak:
Fixes #34

Please take a closer look because this one is a tricky one :)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1100UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1100UnitTests.cs
@@ -1,0 +1,249 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using StyleCop.Analyzers.ReadabilityRules;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using TestHelper;
+
+namespace StyleCop.Analyzers.Test.ReadabilityRules
+{
+    [TestClass]
+    public class SA1100UnitTests : CodeFixVerifier
+    {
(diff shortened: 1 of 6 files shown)


pdelvo:
This unfortunatly still is not working. You are just comparing the name of the parameter, not the type. And there are named parameters, optional parameters, implicit conversions you need to be aware of. (Roslyn has about 3000 lines of code for overload resolution). There is no public api for this unfortunatly :( Here is an example that should produce SA1101 but doesnt:
```csharp
namespace RoslynTest
{
    internal class Base
    {
        public virtual void Bar(long x)
        {

        }
    }

    class Class : Base
    {
        public void Foo()
        {
            base.Bar((long)0);
        }

        public void Bar(int x)
        {
        }
    }
}

```

macpak:
Yeah, it is a cery hard to solve. I was working on it for a few hours and found the Solution added here in Roslyn. I see that it was no Full and I must have missed sth. Ok, I will analyze Roslyn again and gry to repeat the process they implemented. 
Thank you Dennis for the review.

sharwell:
@macpak I reimplemented this in #382, but started with the code from this pull request since you have so many tests implemented. You can see the first commit of #382 is your work from #333. Let me know if you approve of closing this pull request in favor of the other one.

macpak:
Yeah we can close it.
I am just wondering about the performance of the solution you proposed. Have you tested it on a bigger codebase ?
I digged into Rolsyn and found code that actually checks for overldes/hidden members but the code is internal. I [asked](https://roslyn.codeplex.com/discussions/576419) Roslyn team why they decided to hide that, but hasn't got any responses yet.

pdelvo:
You might want to ask [here](https://github.com/dotnet/roslyn). I think @sharwell's solution is the only one that is really working correctly. What we could do is to handle some special cases were we can  decide if the base keyword is necessary without getting this new semantic model. E.g. for methods without arguments or properties and then fall back if we can't decide it for sure.

sharwell:
Closing this. We can continue discussion in #382. @macpak your concern is *absolutely* valid and warrants further discussion. Feel free to post it again in the other issue.

PR #338, opened 2014-12-24
Implemented SA1409 RemoveUnnecessaryCode

### LINKED ISSUE #98: SA1409: RemoveUnnecessaryCode
http://www.stylecop.com/docs/SA1409.html

pdelvo:
Fixes #98. I might add a code fix later

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1400UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1400UnitTests.cs
@@ -577,7 +577,6 @@
             await this.TestNestedDeclarationWithDirectivesAsync("private", "MemberName", "static System.EventHandler MemberName", ", AnotherMemberName;").ConfigureAwait(false);
         }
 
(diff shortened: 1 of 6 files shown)


macpak:
Pretty damn good job :)
I am only wondering if we should keep a parameterless constructor with empty body. 
In my opinion it makes sense in 2 cases:
- if you have another constructor(s) which takes some parameters, and you keep the parameterless constructor in order to avoid breaking code that depends on it . 
- if the constructor calls a base constructor.
Stylecop doesn't report this issue as well.
What do you think Dennis ?

pdelvo:
Thanks ;)
Im only reporting static constructors which get called only once when a class gets used for the first time. You can't explicitly call it and it cant have any parameters or have a base call

macpak:
Yeah, my whole story was about nonstatic constructors :)

sharwell:
Also if it's not public or (for abstract classes) protected. 

Also if it has XML documentation comments. 

The base call doesn't count as necessary if it doesn't pass any arguments.

sharwell:
Marking **wontfix** based on #1058.

PR #341, opened 2014-12-27
Fixed SA1000 is reported for constructor constraints

### LINKED ISSUE #340: SA1000: Generic constraint new()
Currently SA1000 is reporting a diagnostic for this:
```csharp
public void Foo<T>() where T : IInterface, new()
{
    // ...
}
```
SA1000 is reporting, that new has to be followed by a space. If you add this space (new ()), then you get a SA1008. So there is currently no way to get a generic ctor constraint without violating a rule.

pdelvo:
Fixes #340

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
@@ -94,66 +94,70 @@
                 precededBySpace = precedingToken.HasTrailingTrivia;
                 switch (precedingToken.CSharpKind())
                 {
-                case SyntaxKind.AwaitKeyword:
-                case SyntaxKind.CaseKeyword:
-                case SyntaxKind.CatchKeyword:
-                case SyntaxKind.FixedKeyword:
-                case SyntaxKind.ForKeyword:
-                case SyntaxKind.ForEachKeyword:
-                //case SyntaxKind.FromKeyword: // ?
-                case SyntaxKind.GroupKeyword: // ?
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -94,66 +94,70 @@ private void HandleOpenParenToken(SyntaxTreeAnalysisContext context, SyntaxToken
...
-                    {
+                    case SyntaxKind.EqualsGreaterThanToken: // lambda containing a cast
+                    case SyntaxKind.CommaToken:
                         allowLeadingNoSpace = false;
                         allowLeadingSpace = true;
                         break;
-                    }
-                    else
-                    {
+                    case SyntaxKind.NewKeyword:
```
sharwell: This entire case statement case be removed by adding `case SyntaxKind.NewKeyword:` after one of the following (and removing the contents of this case), depending on the desired behavior. See the comment on the pull request for more information.

* After `case SyntaxKind.LockKeyword:`, if the desired spacing is `new ()`
* After `case SyntaxKind.DefaultKeyword:` if the desired spacing is `new()`

sharwell:
As of commit e025fe2, this is implemented so the desired spacing is `new ()`. I believe the spacing users expect is `new()`, despite the manner in which StyleCop documents the SA1000 rule (and indicated by StyleCop ignoring this keyword altogether).

sharwell:
@pdelvo I squashed your commits into a single commit which addresses #340 by disabling the diagnostics. I then added an additional commit to re-enable only SA1000 in this case and not allow a space between `new` and `()` for a default constructor constraint (with several additional tests).

PR #342, opened 2014-12-29
Implemented SA1404 CodeAnalysisSuppressionMustHaveJustification

### LINKED ISSUE #93: SA1404: CodeAnalysisSuppressionMustHaveJustification
http://www.stylecop.com/docs/SA1404.html

pdelvo:
Fixes #93

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeAnalysisSuppressionMustHaveJustification.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeAnalysisSuppressionMustHaveJustification.cs
@@ -79,7 +79,7 @@
                                 var value = context.SemanticModel.GetConstantValue(attributeArgument.Expression);
 
                                 // If value does not have a value the expression is not constant -> Compilation error
-                                if (!value.HasValue || !string.IsNullOrWhiteSpace((string)value.Value))
+                                if (!value.HasValue || !string.IsNullOrWhiteSpace(value.Value as string))
                                     return;
 
                                 // Empty, Whitespace or null justification provided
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeAnalysisSuppressionMustHaveJustification.cs
```diff
@@ -49,7 +53,45 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                        foreach (var argument in attribute.ArgumentList.ChildNodes())
+                        {
+                            var attributeArgument = argument as AttributeArgumentSyntax;
+                            if (attributeArgument?.NameEquals?.Name?.ToString() == "Justification")
+                            {
+                                // Check if the justification is not empty
+
+                                var value = context.SemanticModel.GetConstantValue(attributeArgument.Expression);
+
+                                // If value does not have a value the expression is not constant -> Compilation error
```
sharwell: What happens if the user passes a constant which is not a string? The code won't compile, but would it cause an exception for the cast to `string`?

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeAnalysisSuppressionMustHaveJustification.cs
```diff
@@ -49,7 +53,45 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                if(symbol != null)
+                {
+                    var suppressMessageType = context.SemanticModel.Compilation.GetTypeByMetadataName(typeof(SuppressMessageAttribute).FullName);
+                    if (symbol.ContainingType == suppressMessageType)
+                    {
+
+                        foreach (var argument in attribute.ArgumentList.ChildNodes())
+                        {
+                            var attributeArgument = argument as AttributeArgumentSyntax;
+                            if (attributeArgument?.NameEquals?.Name?.ToString() == "Justification")
```
sharwell: This can be `nameof(SuppressMessageAttribute.Justification)`. :sunglasses:

PR #344, opened 2015-01-01
Handle cases where INamespaceSymbol is not available for a using directive

### LINKED ISSUE #343: SA1210 Diagnostic can throw NullReferenceException
In SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs, Line 62 namespaceSymbol can be null if the document is not valid at the moment. (Happened to me when I typed '.' After System when writing a using directive)

sharwell:
Fixes #343


sharwell:
@pdelvo Can you review this?

pdelvo:
I think this fixes it.

PR #345, opened 2015-01-03
Implemented SA1405 and SA1406

### LINKED ISSUE #94: SA1405: DebugAssertMustProvideMessageText
http://www.stylecop.com/docs/SA1405.html

### LINKED ISSUE #95: SA1406: DebugFailMustProvideMessageText
http://www.stylecop.com/docs/SA1406.html

pdelvo:
Fixes #94 and fixes #95.

I had to add a reference to System.dll to the tests to make the System.Diagnostics.Debug class available. A code fix doesn't make much sense here. It would be nice if someone could check my comments. English is not my first language ;)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1406UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1406UnitTests.cs
@@ -1,0 +1,158 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using StyleCop.Analyzers.MaintainabilityRules;
+using TestHelper;
+
+namespace StyleCop.Analyzers.Test.MaintainabilityRules
+{
+    [TestClass]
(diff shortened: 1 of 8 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1406UnitTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        [TestMethod]
+        public async Task TestWrongDebugClass()
+        {
+            var testCode = @"
+public class Foo
+{
+    public void Bar()
+    {
+        string message = ""A message"";
+        Debug.Fail(message);
```
sharwell: Nice catch to think of this test. :+1: However, the particular input you chose would not produce a diagnostic message even if it was analyzed as the normal `System.Diagnostics.Debug.Fail` method. You should add a test that would normally produce a diagnostic, but does not here only because the `Debug` class is not the expected one.

sharwell:
You've done a great job here. I ran some additional tests locally and everything seems to work like it's supposed to. However, before merging it I'm going to include a few additional tests. You don't need to do any work.

sharwell:
I added a bunch of additional tests in 7b5f7fb94ad195364cf2c58864f075bd0a5112de and 4be27a81cd04e4f0efd505eae57f7882c1c03706. Everything passed.

PR #349, opened 2015-01-11
Implement SA1122 diagnostic and code fix

### LINKED ISSUE #55: SA1122: UseStringEmptyForEmptyStrings
http://www.stylecop.com/docs/SA1122.html

pdelvo:
Fixes #55

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1122UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1122UnitTests.cs
@@ -54,8 +54,37 @@
                     }
                 };
 
-            await VerifyCSharpDiagnosticAsync(string.Format(testCode, useVerbatimLiteral ? "@" : ""), expected, CancellationToken.None);
+            await VerifyCSharpDiagnosticAsync(string.Format(testCode, useVerbatimLiteral ? "@" : string.Empty), expected, CancellationToken.None);
         }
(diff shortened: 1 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1122UnitTests.cs
```diff
@@ -0,0 +1,159 @@
...
+    }
+}";
+
+            await VerifyCSharpFixAsync(string.Format(oldSource, useVerbatimLiteral ? "@" : ""), newSource);
+        }
+
+        [TestMethod]
+        public async Task TestCodeFixMultipleNodes()
+        {
+            // Tests if the code fix works if the SourceSpan of the diagnostic has more then one SynatxNode associated with it
```
sharwell: Nice, I wouldn't have thought of this.

sharwell:
Can you add the following tests? I believe they would fail with the current code.

1. A string literal passed as the argument to an attribute.
2. A string literal which is assigned to a `const` field or local variable.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1122UnitTests.cs
pdelvo: I wouldnt have thought about this too if i would not have tested this with a small project I have with a lot of weird syntax in it ;)

sharwell:
Thanks! By the way, make sure to comment after updating or it won't notify me. :+1:

PR #350, opened 2015-01-11
Implemented SA1121

### LINKED ISSUE #54: SA1121: UseBuiltInTypeAlias

pdelvo:
Fixes #54. Unfortunatly the diagnostic itself was already implemented in #306. But this one contains a bunch of unit tests so it might be a good idea to use them

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
@@ -1,0 +1,724 @@
+﻿namespace StyleCop.Analyzers.Test.ReadabilityRules
+{
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using StyleCop.Analyzers.ReadabilityRules;
+    using TestHelper;
+    using System;
+
+
+    /// <summary>
+    /// This class contains unit tests for <see cref="SA1121UseBuiltInTypeAlias"/>
+    /// </summary>
+    [TestClass]
+    public class SA1121UnitTests : CodeFixVerifier
+    {
+        static readonly Tuple<string, string>[] _referenceTypes = new Tuple<string, string>[]
+        {
+            new Tuple<string,string>("object", "Object"),
+            new Tuple<string,string>("string", "String")
+        };
+        static readonly Tuple<string, string>[] _valueTypes = new Tuple<string, string>[]
+        {
+            new Tuple<string,string>("bool", "Boolean"),
+            new Tuple<string,string>("byte", "Byte"),
+            new Tuple<string,string>("char", "Char"),
+            new Tuple<string,string>("decimal", "Decimal"),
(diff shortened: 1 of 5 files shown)


sharwell:
This is currently blocked on #306. See #54 for additional information.

sharwell:
Now that #306 is merged, can you rebase this and reduce it to a nice set of tests (plus enabling the diagnostic)? :+1:

pdelvo:
This should bring in the tests. Im working on the code fix

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
```diff
@@ -0,0 +1,724 @@
...
+
+    /// <summary>
+    /// This class contains unit tests for <see cref="SA1121UseBuiltInTypeAlias"/>
+    /// </summary>
+    [TestClass]
+    public class SA1121UnitTests : CodeFixVerifier
+    {
+        static readonly Tuple<string, string>[] _referenceTypes = new Tuple<string, string>[]
+        {
+            new Tuple<string,string>("object", "Object"),
```
sharwell: `nameof(Object)` for the right hand side here. :wink: The same for the items below.

sharwell:
Can you add this test?

```csharp
namespace Foo
{
  using Int32 = System.UInt32;
  class Bar
  {
    Int32 value = 3;
  }
}
```

A diagnostic should be reported, but the target built-in type is actually `uint` due to the misleading using alias directive.

sharwell:
Also can you add this test?

```csharp
namespace Foo
{
  using MyInt = System.Int32;
  class Bar
  {
    MyInt value = 3;
  }
}
```

pdelvo:
What are the desired results here? Should using Int32 = System.UInt32; change to using Int32 = uint; or should the Int32 value = 3; change to uint value = 3; ? Or both? Currently it will report 2 diagnostics at MyInt value = 3; and the code fix im implementing is replacing it there

sharwell:
1. Using alias directives are not allowed to use the language-specific keywords to refer to types. No diagnostic should be reported for the following line:

    ```csharp
    using Int32 = System.Int32;
    ```

2. In the first example I posted (the misleading using alias directive), the using alias directive would not be altered by the code fix. The reference to the alias would be changed from `Int32` to `uint`.

3. In the second example I posted, the using alias directive again would not change, and the reference to `MyInt` would be changed to `int`.

pdelvo:
These new tests have some issues. The diagnostic analyzer is called twice for the same IdentifierToken. Do you know why this might happen?

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
```diff
@@ -0,0 +1,1114 @@
...
+                        Id = DiagnosticId,
+                        Message = "Use built-in type alias",
+                        Severity = DiagnosticSeverity.Warning,
+                        Locations =
+                            new[]
+                            {
+                                new DiagnosticResultLocation("Test0.cs", 6, 5)
+                            }
+                    },
+                // Workaround because the diagnostic is called twice for the SyntaxNode
```
pdelvo: I could not figure out why this happens. Maybe a roslyn bug?

sharwell:
It's a bug somewhere in the unit test framework for this extension (modified from the one in the Roslyn templates). The diagnostics are only reported once for actual code, so I merged this and created an issue to find the bug in the unit test framework.

PR #357, opened 2015-01-13
Implemented SA1302: InterfaceNamesMustBeginWithI

macpak:


pdelvo:
The exception for NativeMethods classes is missing (same as in #362)

sharwell:
This is blocked on #358, where a shared `NativeMethods` check is being implemented.

sharwell:
@macpak This is ready to be updated to use the new helper method.

macpak:
NativeMethods case is handled right now + I fixed IsContainedInNativeMethodsClass because it had a small bug.

sharwell:
I updated this to use the syntax tree instead of the semantic model (bf0050d827418546d3c74bc01e392c6c1d63dbbc) and enabled it by default (417445377dd1cd05f4bb63e82b116e1ff4cb78c6). Let me know if you see any issues with that! :+1:

Issue #79, opened 2014-11-18
SA1303: ConstFieldNamesMustBeginWithUpperCaseLetter


sharwell:
http://www.stylecop.com/docs/SA1303.html

macpak:
grabbed

sharwell:
Fixed in #358.

macpak:
Can it be closed ? The code is implemented, we just need to add code fix.

PR #358, opened 2015-01-13
Implemented SA1303: ConstFieldNamesMustBeginWithUpperCaseLetter

macpak:


pdelvo:
The exception for NativeMethods classes is missing (same as in #362)

macpak:
Ok, exception added. I will add this exception to other analyzers when we merge this pull request, because I'd like to use the helper (to check if the class is NativeMethods) from this commit.

pdelvo:
This looks good to me

sharwell:
It also applies to #3. Since you implemented it originally here, I'm going to treat #3 as blocked on *this* issue. Then after you apply this to #362 I'll update #3 to use whatever that pattern is.

There's going to be a slight gap in my reviews, but I'll try to get as many merged on **Saturday/Sunday** as possible. If either of you send a pull request which you want reviewed earlier than other pull requests, please include a comment that states that it's an issue you're waiting on and I'll try to prioritize those during the review process.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1303ConstFieldNamesMustBeginWithUpperCaseLetter.cs
```diff
@@ -45,7 +49,29 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+        private void HandleFieldDeclaration(SymbolAnalysisContext context)
+        {
+            var symbol = context.Symbol as IFieldSymbol;
+
+            if (symbol == null || !symbol.IsConst)
+            {
+                return;
+            }
+
+            if(namedTypeHelpers.IsNativeMethodsClass(symbol.ContainingType))
```
sharwell: This should check all containing types. For example:

```csharp
class NativeMethods
{
  class Anything
  {
    const int this_field_is_fine;
  }
}
```

sharwell:
@macpak :exclamation: In the future can you include the text "Fixes #79" (or whatever the specific issue number(s) are) in the pull request description so I can track the pull request back to the original issue?

PR #362, opened 2015-01-13
Implemented SA1304: NonPrivateReadonlyFieldsMustBeginWithUpperCaseLetter

macpak:


pdelvo:
In http://www.stylecop.com/docs/SA1304.html there is an exception mentioned which is missing

> If the field or variable name is intended to match the name of an item associated with Win32 or COM, and thus needs to begin with a lower-case letter, place the field or variable within a special NativeMethods class. A NativeMethods class is any class which contains a name ending in NativeMethods, and is intended as a placeholder for Win32 or COM wrappers. StyleCop will ignore this violation if the item is placed within a NativeMethods class.

pdelvo:
@macpak Could you review my pull requests too?

macpak:
@pdelvo yes I will do it soon.
About this NativeMethods, sorry for the problem with this, I did all this "upper/lowercase names" in one go and that's why I've missed this requirement. 
Thank you for your help.

pdelvo:
No Problem ;) Could you habe a look at my pull requests too?

macpak:
Of course :)

sharwell:
This is currently blocked on #358.

sharwell:
@macpak This is ready to be updated to use the new helper method.

sharwell:
@macpak I updated this to use the syntax tree instead of the semantic model (cc14922de1600d963a95acadc9f8366d52494eb0), enabled it by default (81f5d8eb812ccac775058dcbf7cbaaa7e0a539c1), and then merged it. If you have any issues with this let me know. :+1:

macpak:
Looks ok, just tell me why using syntaxTree analyzis instead of symbol analyzis is better ? Better performance ?

PR #363, opened 2015-01-14
Implement SA1119

### LINKED ISSUE #88: SA1119: StatementMustNotUseUnnecessaryParenthesis
http://www.stylecop.com/docs/SA1119.html

pdelvo:
Fixes #88.
This is a very tricky one. I tried it on Json.Net. It found a lot of violations but they all seem correct. The Code Fix was able to fix them all at once.
I am mostly compatible with StyleCop I think except for the one thing mentioned in #88. because of readability I don't report diagnostics for parenthised expressions in BinaryExpresions.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1119UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1119UnitTests.cs
@@ -1,0 +1,1855 @@
+﻿namespace StyleCop.Analyzers.Test.MaintainabilityRules
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Analyzers.MaintainabilityRules;
+    using TestHelper;
+
+    [TestClass]
+    public class SA1119UnitTests : CodeFixVerifier
+    {
+        private const string DiagnosticId = SA1119StatementMustNotUseUnnecessaryParenthesis.DiagnosticId;
+        protected static readonly DiagnosticResult[] EmptyDiagnosticResults = { };
+
+        [TestMethod]
+        public async Task TestEmptySource()
+        {
+            var testCode = string.Empty;
+            await VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        [TestMethod]
+        public async Task TestLiteral()
+        {
+            var testCode = @"public class Foo
+{
+    public void Bar()
+    {
+        int x = 1;
+    }
+}";
+            await VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        [TestMethod]
+        public async Task TestLiteralParenthesis()
+        {
+            var testCode = @"public class Foo
+{
+    public void Bar()
+    {
+        int x = (1);
+    }
+}";
+            var expected = new[]
+            {
+                new DiagnosticResult
+                {
+                    Id = DiagnosticId,
+                    Message = "Statement must not use unnecessary parenthesis",
+                    Severity = DiagnosticSeverity.Warning,
+                    Locations =
+                        new[]
+                        {
+                            new DiagnosticResultLocation("Test0.cs", 5, 17)
+                        }
+                },
+                new DiagnosticResult
+                {
+                    Id = DiagnosticId + "_p",
+                    Message = "Statement must not use unnecessary parenthesis",
+                    Severity = DiagnosticSeverity.Hidden,
+                    Locations =
+                        new[]
+                        {
+                            new DiagnosticResultLocation("Test0.cs", 5, 17)
+                        }
+                },
+                new DiagnosticResult
+                {
+                    Id = DiagnosticId + "_p",
+                    Message = "Statement must not use unnecessary parenthesis",
+                    Severity = DiagnosticSeverity.Hidden,
+                    Locations =
+                        new[]
+                        {
+                            new DiagnosticResultLocation("Test0.cs", 5, 19)
+                        }
+                }
+            };
+
+            await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None);
+
+            var fixedCode = @"public class Foo
+{
+    public void Bar()
+    {
+        int x = 1;
+    }
+}";
+            await VerifyCSharpFixAsync(testCode, fixedCode);
+        }
+
+        [TestMethod]
+        public async Task TestLiteralDoubleParenthesis()
+        {
+            var testCode = @"public class Foo
+{
+    public void Bar()
+    {
+        int x = ((1));
+    }
+}";
+            var expected = new[]
+            {
+                new DiagnosticResult
+                {
+                    Id = DiagnosticId,
+                    Message = "Statement must not use unnecessary parenthesis",
+                    Severity = DiagnosticSeverity.Warning,
+                    Locations =
+                        new[]
+                        {
+                            new DiagnosticResultLocation("Test0.cs", 5, 17)
+                        }
+                },
+                new DiagnosticResult
+                {
+                    Id = DiagnosticId + "_p",
+                    Message = "Statement must not use unnecessary parenthesis",
+                    Severity = DiagnosticSeverity.Hidden,
+                    Locations =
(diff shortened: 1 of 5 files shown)


pdelvo:
I rebased a few improvements ans bug fixes

sharwell:
I'm not merging this for now because I'd like to see this issue resolved. I'll post some additional comments in the code regarding less important "tweaks" that I would have just done myself during the merge, in case you want to fix them.

:exclamation: This diagnostic uses `WellKnownDiagnosticTags.Unnecessary`, which results in the diagnostic span getting grayed out in the editor. In the current implementation, the "location" for this diagnostic is the *complete expression*, where the actual unnecessary region is just the parentheses. During my experimentation, I attempted to use the `additionalLocations` parameter to `Diagnostic.Create` in order to specify two regions which are both unnecessary but belong to the same diagnostic instance:

```csharp
private Diagnostic CreateDiagnostic(ParenthesizedExpressionSyntax syntax)
{
    Location primaryLocation = syntax.OpenParenToken.GetLocation();
    Location[] secondaryLocations = { syntax.CloseParenToken.GetLocation() };

    return Diagnostic.Create(Descriptor, primaryLocation, additionalLocations: secondaryLocations);
}
```

Unfortunately this did not have the expected result. Instead, I believe you will need to do the following:

1. Report SA1119 for only the open parenthesis.
2. Define a new diagnostic in the same analyzer:

    * Use `DiagnosticSeverity.Hidden`
    * Use the tags `WellKnownDiagnosticTags.NotConfigurable` *and* `WellKnownDiagnosticTags.Unnecessary`
    * Do not include the diagnostic in the array returned by `SupportedDiagnostics`. Hopefully this will cause Roslyn to not call the code at all when SA1119 is disabled. Since a `NotConfigurable` diagnostic is always enabled, it would override the setting for SA1119 and run this code anyway, even if the SA1119 portion was not shown in the editor UI.

3. Report the new hidden diagnostic for the closing parenthesis.
4. Update the code fix provider to trigger on just the SA1119 open parenthesis. The updated code might look like this (replacing the current call to `FindNode` through the declaration of `syntax`):

    ```csharp
    SyntaxToken openParenToken = root.FindToken(diagnostic.Location.SourceSpan.Start);
    if (openParenToken.IsMissing || !openParenToken.IsKind(SyntaxKind.OpenParenToken))
        continue;

    ParenthesizedExpressionSyntax syntax = openParenToken.Parent as ParenthesizedExpressionSyntax;
    ```

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119CodeFixProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Microsoft.CodeAnalysis.Formatting;
+
+
```
sharwell: These blank lines are getting inserted due to a bug in the Remove and Sort Usings feature in Visual Studio 2015 preview. After running this command, you should look at the result and see if it added the extra lines.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119StatementMustNotUseUnnecessaryParenthesis.cs
```diff
@@ -60,7 +66,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+            context.RegisterSyntaxNodeAction(HandleParenthesizedExpressionn, SyntaxKind.ParenthesizedExpression);
+        }
+
+        private void HandleParenthesizedExpressionn(SyntaxNodeAnalysisContext context)
+        {
+            var node = context.Node as ParenthesizedExpressionSyntax;
+
+            if (node != null && node.Expression != null)
+            {
+                if (!(node.Expression is BinaryExpressionSyntax) && !(node.Expression is AssignmentExpressionSyntax)
```
sharwell: If you are going to separate this expression on the `&&` operator (which you should here due to length), can you go ahead and separate it on every `&&` operator at the same precedence within the expression?

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119StatementMustNotUseUnnecessaryParenthesis.cs
```diff
@@ -60,7 +66,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                    && !node.Expression.IsKind(SyntaxKind.IsExpression) && !node.Expression.IsKind(SyntaxKind.SimpleLambdaExpression)
+                    && !node.Expression.IsKind(SyntaxKind.ParenthesizedLambdaExpression) && !node.IsKind(SyntaxKind.ConstructorDeclaration)
+                    && !node.Expression.IsKind(SyntaxKind.ArrayCreationExpression) && !node.Expression.IsKind(SyntaxKind.CoalesceExpression) 
+                    && !node.Expression.IsKind(SyntaxKind.QueryExpression))
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, node.GetLocation()));
+                }
+                else
+                {
+                    if (!(node.Parent is ExpressionSyntax) || node.Parent.IsKind(SyntaxKind.VariableDeclaration)
```
sharwell: The same applies to this `||` operator as the `&&` operator above.

sharwell: :question: Also, is a `VariableDeclaration` an `ExpressionSyntax`? If not I believe the test for `VariableDeclaration` is unnecessary.

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1119UnitTests.cs
```diff
@@ -0,0 +1,1239 @@
...
+                ;
+        }
+}";
+            await VerifyCSharpFixAsync(testCode, fixedCode);
+        }
+
+        [TestMethod]
+        public async Task TestCodeFixWorksInTrivia()
+        {
+            var testCode = @"#if (NET20 || NET35)
```
sharwell: Nice catch. :sunglasses:

pdelvo: I was trying it on Json.net which has a lot of this and had a few issues before. ;)

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119StatementMustNotUseUnnecessaryParenthesis.cs
pdelvo: I will remove that

pdelvo:
This is what it looks like with your changes:
sa1119
You can barely see that the closing parenthesis is faded out. Im not sure if I like that better.
The code fix actually works without changing anything.

sharwell:
I believe that is a more accurate representation. *However*, if you want to do even more (shouldn't be much more than you just did), you could do the following:

1. Leave the hidden diagnostic you defined as it is.
2. Remove the `WellKnownDiagnosticTags.Unnecessary` tag from the main SA1119.
3. Report *3* diagnostics:
    1. SA1119 is reported for the entire expression. Nothing is grayed out, and the green squiggle goes under the entire expression.
    2. Report the hidden diagnostic for the close parenthesis, like you do now.
    3. Report an additional instance of the same hidden diagnostic for the open parenthesis.

I believe you will like the UI of this much more.

pdelvo:
I will do that. Actually it is not as trivial because I have to change a lot of unit tests too

sharwell:
:question: I thought you shouldn't need to change the tests, because the first character of the location where the SA1119 diagnostic is reported stays the same.

pdelvo:
But the diagnostic now reports 3 diagnostics. 2 start at the same position but they are different and one is at the closing parenthesis

sharwell:
:frowning: Sorry about that.

Do you agree this is the right approach for the user experience?

pdelvo:
Done.

sharwell:
Unfortunately I'm out of time today :disappointed: , but it shouldn't take me too long to review this again when I have more time. :+1:

PR #373, opened 2015-01-16
Fix SA1408 is reporting for wrong Binary Expressions

pdelvo:
```csharp
bool b = true || 1 == 2;
```
was reporting SA1408


pdelvo:
This pull request needs some more work. Found some more problems with it (is operator)

pdelvo:
This is fixed now

PR #376, opened 2015-01-17
Enhance SA1303 ConstFieldNamesMustBeginWithUpperCaseLetter

### LINKED ISSUE #369: Fix remaining issues with SA1303
Pull request #358 is working and contains a new extension method which is blocking several other pull requests. Therefore, I'm merging the pull request prior to the completion of certain "less important" details.

1. SA1303 needs to be enabled by default
2. The implementation currently uses `char.IsLower` to check the first letter. Should this be using `char.IsUpper` instead? Considerations include non-ASCII identifier characters as well as potential conflicts of `_` with [SA1309](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/85).

macpak:
Fixes #369.
As we discussed, we keep the old behaviour. Additionally I added a test to make sure we don't report SA1303 for fields starting with `_`.
The diagnostic is now enabled by default.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1303UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1303UnitTests.cs
@@ -174,6 +174,17 @@ namespace Test
         }
 
         [TestMethod]
+        public async Task TestConstFieldStatingWithUnderscore()
+        {
+            var testCode = @"public class Foo
+{
+    public const string _Bar = ""baz"";
+}";
+
+            await VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        [TestMethod]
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1303UnitTests.cs
```diff
@@ -174,6 +174,17 @@ public class FooInner
         }
 
         [TestMethod]
+        public async Task TestConstFieldStatingWithUnderscore()
```
sharwell: * :exclamation: This should read `Starting`
* :exclamation: This should include a comment in the code that SA1303 should not be reported because this issue is reported by SA1309

sharwell:
Needs two minor tweaks but otherwise looks good. :+1:

sharwell:
I fixed those items in 977a3ba7c21f258a692f8bb593819c0b8b289165 and merged this. :+1:

PR #377, opened 2015-01-17
Update to Visual Studio 2015 CTP 5

sharwell:
This pull request updates the project to Visual Studio 2015 CTP 5.

It is currently blocked on dotnet/roslyn#23, which is causing a total of 10 tests to fail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ obj/
 
 # Artifacts of the IDE and build
 *.sln.ide/
+.vs/
 packages/
 *.suo
 *.user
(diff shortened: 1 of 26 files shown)


sharwell:
The issue with formatting has been fixed by the introduction of a `WithoutElasticTrivia()` extension method for `SyntaxNode`, `SyntaxToken`, and `SyntaxTrivia`.

pdelvo:
looks good

sharwell:
@pdelvo Thanks for reviewing this so quickly. :smile: Can you look at the changes made in the last commit as well?

pdelvo:
Looks good too. Why did you put the SpacingExtensions.cs under StyleCop.Analyzers/SpacingRules/? it could be usefull for other diagnostics as well so I think StyleCop.Analyzers/Helpers/ would be better

sharwell:
It was added back when the only code fixes were for spacing rules. I left it there for now; I'm torn on the name **/Helpers/** so I think they may both get refactored in the future (and probably together).

Issue #334, opened 2014-12-19
Code fix: DoNotPrefixCallsWithBaseUnlessLocalImplementationExists


macpak:
Implement a code fix for #34.

sharwell:
This should be pretty straightforward - just replace `base` with `this`.

macpak:
Yes, I will implement IT today.

macpak:
Ok, I thought that the diagnostic was merged. I am gonna wait with this until the diatnostic is in master.

sharwell:
@macpak I merged it. Thanks for all the help on it. :+1:

PR #382, opened 2015-01-18
Implement diagnostic SA1100 DoNotPrefixCallsWithBaseUnlessLocalImplementationExists

### LINKED ISSUE #34: SA1100: DoNotPrefixCallsWithBaseUnlessLocalImplementationExists

sharwell:
Fixes #34 

This implementation supersedes #333. I used the tests implemented by that pull request, but replaced the implementation with one based on "experimental compilations" to determine whether or not the `base.` expression is necessary on a case-by-case basis.

There is quite a bit of room available for optimizing this implementation (essentially by various forms of pruning), but this should allow us to start with a correct initial implementation.


macpak:
This pull request blocks #334.

pdelvo:
As I already said in #333: This PR looks good and I don't think we can do it better without putting much work into it

macpak:
Ok.
When it's merged I will add the codefix.

sharwell:
Thanks y'all. :+1: This was certainly a challenging one!

macpak:
[I am talking to Microsoft](https://github.com/dotnet/roslyn/issues/200) if the problem we have here can be solved in a different way.

PR #394, opened 2015-01-18
Fix SA1121 incorrectly reported in nameof

### LINKED ISSUE #387: SA1121 UseBuiltInTypeAlias incorrectly reported in nameof
The following statement incorrectly results in SA1121.

```csharp
string objectName = nameof(System.Object);
```

pdelvo:
Fixes #387

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1121UseBuiltInTypeAlias.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1121UseBuiltInTypeAlias.cs
@@ -207,13 +207,58 @@
             default:
                 return;
             }
-
+            
             SyntaxNode locationNode = identifierNameSyntax;
             if (identifierNameSyntax.Parent is QualifiedNameSyntax)
                 locationNode = identifierNameSyntax.Parent;
 
+            // Allow nameof
+            if(IsInNameOfExpression(identifierNameSyntax, semanticModel))
+            {
+                return;
+            }
+
             // Use built-in type alias
             context.ReportDiagnostic(Diagnostic.Create(Descriptor, locationNode.GetLocation()));
+        }
+
+        private bool IsInNameOfExpression(IdentifierNameSyntax identifierNameSyntax, SemanticModel semanticModel)
+        {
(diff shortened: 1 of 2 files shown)


pdelvo:
Don't merge this yet. I just found a bug

pdelvo:
fixed

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1121UseBuiltInTypeAlias.cs
```diff
@@ -207,13 +207,67 @@ private void HandleIdentifierNameSyntax(SyntaxNodeAnalysisContext context)
...
+            if(IsInNameOfExpression(identifierNameSyntax, semanticModel))
+            {
+                return;
+            }
+
             // Use built-in type alias
             context.ReportDiagnostic(Diagnostic.Create(Descriptor, locationNode.GetLocation()));
         }
+
+        private bool IsInNameOfExpression(IdentifierNameSyntax identifierNameSyntax, SemanticModel semanticModel)
```
sharwell: It seems this entire method could be simplified to the following:

```csharp
private bool IsNameInNameOfExpression(IdentifierNameSyntax identifierNameSyntax)
{
    // The only time a type name can appear as an argument is for the invocation expression created for the
    // nameof keyword. This assumption is the foundation of the following simple analysis algorithm.

    if (identifierNameSyntax.Parent == null)
        return false;

    // This covers the case nameof(Int32)
    if (identifierNameSyntax.Parent is ArgumentSyntax)
        return true;

    // This covers the case nameof(System.Int32)
    if (identifierNameSyntax.Parent.IsKind(SyntaxKind.SimpleMemberAccessExpression))
    {
        // This final check ensures that we don't consider nameof(System.Int32.ToString) the same as
        // nameof(System.Int32)
        if (identifierNameSyntax.Parent.Parent is ArgumentSyntax)
            return true;
    }

    return false;
}
```

pdelvo: You're right. I was thinking too complicated here

pdelvo:
I replaced my method wouth yours @sharwell.

sharwell:
:+1:

PR #398, opened 2015-01-18
SA1401 - do not report this diagnostic for const fields

### LINKED ISSUE #384: Consider disabling SA1401 FieldsMustBePrivate for constants
Currently SA1401 is reported for `const` fields. Considering these fields cannot be encapsulated in a property without changing their semantics, it seems SA1401 should be disabled for this case.

macpak:
Fixes #384.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1401UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1401UnitTests.cs
@@ -95,7 +95,18 @@ namespace StyleCop.Analyzers.Test.MaintainabilityRules
             await VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
         }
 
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1401UnitTests.cs
```diff
@@ -95,7 +95,18 @@ public class SA1401UnitTests : CodeFixVerifier
             await VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
         }
 
-        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
+		[TestMethod]
```
pdelvo: The indentation is off here

macpak:
Seems like github problem, because when I look at the code in VS it looks ok.

pdelvo:
You are using Tabs instead of spaces here this is why it looks good in your editor

macpak:
Seems like update of VS changed it. I will fix it thanks !

PR #399, opened 2015-01-18
Enable SA1601 and SA1602 by default

pdelvo:


sharwell:
I just saw this on your progress report :sunglasses:

pdelvo:
This is where I saw that too ;)

PR #400, opened 2015-01-19
Implement codefix SA1100 DoNotPrefixCallsWithBaseUnlessLocalImplementationExists

### LINKED ISSUE #334: Code fix: DoNotPrefixCallsWithBaseUnlessLocalImplementationExists

macpak:
Fixes #334.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1100UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1100UnitTests.cs
@@ -1,4 +1,6 @@
-﻿namespace StyleCop.Analyzers.Test.ReadabilityRules
+﻿using Microsoft.CodeAnalysis.CodeFixes;
+
+namespace StyleCop.Analyzers.Test.ReadabilityRules
 {
     using System.Threading;
     using System.Threading.Tasks;
@@ -50,6 +52,24 @@ public class FooChild : Foo
             };
 
             await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None);
+
+            var fixedTest = @"
+public class Foo
+{
+    protected virtual void Bar()
+    {
+
+    }
+}
+
+public class FooChild : Foo
+{
+    protected override void Baz()
+    {
+        this.Bar();
+    }
+}";
+            await VerifyCSharpFixAsync(testCode, fixedTest, cancellationToken: CancellationToken.None);
         }
 
         [TestMethod]
@@ -88,6 +108,24 @@ public class FooChild : Foo
             };
 
             await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None);
+
+            var fixedTest = @"
+public class Foo
+{
+    protected void Bar()
+    {
+
+    }
+}
+
+public class FooChild : Foo
+{
+    protected override void Baz()
+    {
+        this.Bar();
+    }
+}";
+            await VerifyCSharpFixAsync(testCode, fixedTest, cancellationToken: CancellationToken.None);
         }
 
         [TestMethod]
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1100CodeFixProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+
+                var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+                var node = root.FindNode(diagnostic.Location.SourceSpan) as BaseExpressionSyntax;
+                if (node == null)
+                {
+                    return;
+                }
+
+                var thisExpressionSyntax = SyntaxFactory.ThisExpression();
```
sharwell: Due to [recent changes in the compiler](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/377), you should use this instead:

```csharp
var thisExpressionSyntax = SyntaxFactory.ThisExpression().WithoutFormatting();
```

sharwell: You also need to copy the leading and trailing trivia from the original `BaseExpressionSyntax` over to the new `ThisExpressionSyntax`.

macpak: Ok. Could you provide a test that breaks the current code ?

pdelvo: Example:
```
#if true
        base
#else
    this
#endif
.ToString();
```
When I look at the syntax tree for this would I think will happen is that your code fix will remove the #if directive, some whitespaces and the end of line token. If you think that no one will ever do something incredible like this have a look at Json.Net. They love their #ifs and #elses

sharwell: Nice example (once again) @pdelvo 

I was thinking of comments and you break out the preprocessor directives.

pdelvo: Haha thanks :D They are my new favorite way to break stuff

macpak:
Allright, thank you for your input. I will fix this :)

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1100CodeFixProvider.cs
sharwell: @macpak After #404 is merged, you could use the following to cover everything:

```csharp
var thisExpressionSyntax = SyntaxFactory.ThisExpression()
    .WithTriviaFrom(node)
    .WithoutFormatting();
```

macpak:
I corrected the codefix. Additionally I did a small cleanup of tests's code.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1100UnitTests.cs
```diff
@@ -790,11 +1054,223 @@ public string Baz()
...
+        this
+
+        .Bar();
+    }
+}";
+            await VerifyCSharpFixAsync(testCode, fixedTest, cancellationToken: CancellationToken.None);
+        }
+
+        [TestMethod]
+        public async Task TestChildClassUsesBaseWithPreprocessorDirectivesButNoOverrideTwoIssues()
```
sharwell: :+1:

sharwell:
I enabled SA1100 by default as well in 9ddbae3676e71b85d67032cf68c814af305372b3.

macpak:
Cool :)

Issue #359, opened 2015-01-13
Suggestion: Known Changes Document


pdelvo:
I think it would be a good idea to have a list of changes from StyleCop to this project with the reason why it is different. Im talking about known changes (not bugs) where we report a diagnostic where stylecop does not or the other way around. So that people that migrate from stylecop know if they encountered a bug or not.

sharwell:
A very good idea.

sharwell:
For reference, SA1001 includes an intentional deviation, where `Func<,>` is preferred over the form `Func<, >`, even though the StyleCop documentation suggests the latter would be correct. See #475 for additional information.

sharwell:
SA1642 includes an intentional deviation as described in #413.

sharwell:
SA1000 includes modifications to the original description, as described in #188, #189.

It also includes an intentional deviation as described in #340.

sharwell:
SA1119 currently includes a deviation which was reported in #1450.

sharwell:
SA1025 currently includes a deviation which was reported in #1424. The rule proposed in #993 would move this violation to its own rule, which could be disabled for improved compatibility with StyleCop "classic".

jackwhelpton:
SA1515 currently includes a deviation which was reported in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1490. I think our code included another example that was less justifiable; I'll find that now.

stylpe:
SA1649 now also checks the actual file name rather than just the file attribute in the file header, ref http://www.stylecop.com/docs/SA1649.html

I saw you asked for this explicitly in #169, otherwise I might have filed this as an issue :)

Issue #48, opened 2014-11-18
SA1114: ParameterListMustFollowDeclaration


sharwell:
http://www.stylecop.com/docs/SA1114.html

macpak:
grabbed

Issue #47, opened 2014-11-18
SA1113: CommaMustBeOnSameLineAsPreviousParameter


sharwell:
http://www.stylecop.com/docs/SA1113.html

macpak:
grabbed

Issue #44, opened 2014-11-18
SA1110: OpeningParenthesisMustBeOnDeclarationLine


sharwell:
http://www.stylecop.com/docs/SA1110.html

macpak:
grabbed

macpak:
I am wondering about this diagnostic. Maybe you can help.
Let's consider an example:

    class Program
    {
        private List<int> items;

        static void Main(string[] args)
        {
        }

        public string JoinName(string first, string last)
        {
            return JoinStrings(
                first, last);
        }

        private string JoinStrings(string first, string last)
        {
            return first + last;
        }

        public int this[int x]
        {
            get 
            { 
                return this.items
                [x]; 
            }
        }
    }

According to the description of this diagnostic, we should report SA1110 violation on the line:
    
    return this.items
                [x]; 

However, Stylecop doesn't report this.
Moreover, if I test a method:
        
        public int TestIndexer()
        {
            return this
                [0];
        }

Stylecop reports SA1110 on indexer call.
So I am not sure whether this is a bug or description is incorrect/misleading.

HowardvanRooijen:
Not a direct response to you question - but a wider question about the project? Have you thought about migrating the existing tests (or at lease test examples) from the stylecop project?

macpak:
Good point! 
I think it'd be very valuable.

HowardvanRooijen:
I spent a huge amount of time building StyleCop for ReSharper, before StyleCop was OSS'd - I wish I had access to the tests. Figuring out how to test that a quick fix alters the code structure and testing it against a "gold" image was the approach I used - so essentially you have a before and after snapshot of the code. I'm not sure with VS if the dev team have figured out a better approach for testing this API - but that certainly worked for me.

sharwell:
@macpak In your example, SA1110 should be reported regardless of where the call appears (you showed cases within a property accessor and a method body, but other locations are possible as well).

macpak:
Of course, but my question is different.
One of the violations of SA1110 is when an open bracket of indexer call is not placed at the same line as the indexer. My example contains code:

    return this.items
            [x];

where the open bracket is not placed at the same line as indexer call. However, in my opinion it seems that in this example, we actually don't see "the indexer" because indexer call is a simple usage of brackets. So maybe, SA1110 should only be reported for incorrect indexer declarations or incorrect indexer calls that use "this" like:

      public int TestIndexer()
        {
            return this
                [0];
[...]

PR #401, opened 2015-01-19
Implement diagnostic SA1111: ClosingParenthesisMustBeOnLineOfLastParameter

### LINKED ISSUE #45: SA1111: ClosingParenthesisMustBeOnLineOfLastParameter
http://www.stylecop.com/docs/SA1111.html

macpak:
Fixes #45.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
@@ -66,6 +66,29 @@ namespace StyleCop.Analyzers.ReadabilityRules
             context.RegisterSyntaxNodeAction(HandleElementAccessExpression, SyntaxKind.ElementAccessExpression);
             context.RegisterSyntaxNodeAction(HandleDelegateDeclarationExpression, SyntaxKind.DelegateDeclaration);
             context.RegisterSyntaxNodeAction(HandleAttribute, SyntaxKind.Attribute);
+            context.RegisterSyntaxNodeAction(HandleAnonymousMethod, SyntaxKind.AnonymousMethodExpression);
+        }
+
+        private void HandleAnonymousMethod(SyntaxNodeAnalysisContext context)
+        {
+            var anonymousMethod = (AnonymousMethodExpressionSyntax) context.Node;
+
+            if (anonymousMethod.ParameterList == null ||
+                !anonymousMethod.ParameterList.IsMissing &&
+                !anonymousMethod.ParameterList.Parameters.Any())
+            {
+                return;
+            }
+
+            var lastParameter = anonymousMethod.ParameterList
+                .Parameters
+                .Last();
+
+            if (!anonymousMethod.ParameterList.CloseParenToken.IsMissing)
+            {
+                CheckIfLocationOfLastArgumentOrParameterAndCloseTokenAreTheSame(context, lastParameter,
+                    anonymousMethod.ParameterList.CloseParenToken);
+            }
         }
 
         private void HandleAttribute(SyntaxNodeAnalysisContext context)
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
@@ -1,11 +1,11 @@
-﻿using System.Linq;
-using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
+﻿using System;
 
 namespace StyleCop.Analyzers.ReadabilityRules
 {
     using System.Collections.Immutable;
     using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
```diff
@@ -54,7 +58,154 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+            context.RegisterSyntaxNodeAction(HandleConstructorDeclaration, SyntaxKind.ConstructorDeclaration);
+            context.RegisterSyntaxNodeAction(HandleInvocationExpression, SyntaxKind.InvocationExpression);
+            context.RegisterSyntaxNodeAction(HandleObjectCreationExpression, SyntaxKind.ObjectCreationExpression);
+            context.RegisterSyntaxNodeAction(HandleIndexerDeclaration, SyntaxKind.IndexerDeclaration);
+            context.RegisterSyntaxNodeAction(HandleElementAccessExpression, SyntaxKind.ElementAccessExpression);
+        }
+
+        private void HandleElementAccessExpression(SyntaxNodeAnalysisContext context)
+        {
+            var tyleSymbol = context.SemanticModel.GetTypeInfo(((ElementAccessExpressionSyntax)context.Node).Expression);
```
pdelvo: A spelling mistake here

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
```diff
@@ -54,7 +58,154 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+
+            if ((tyleSymbol.Type != null && tyleSymbol.Type.TypeKind != TypeKind.Array) )
```
pdelvo: Your if-statement has an extra pair of parenthesis here

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
```diff
@@ -54,7 +58,154 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+
+        private static void CheckIfLocationOfLastArgumentOrParameterAndCloseTokenAreTheSame(SyntaxNodeAnalysisContext context,
+       CSharpSyntaxNode parameterOrArgument, SyntaxToken closeToken)
+        {
+            var lastParameterLine = parameterOrArgument.GetLocation().GetLineSpan();
+            var closeParenLocation = closeToken.GetLocation();
+            var closeParenLine = closeParenLocation.GetLineSpan();
+            if (lastParameterLine.IsValid &&
+                closeParenLine.IsValid &&
+                closeParenLine.StartLinePosition.Line != lastParameterLine.StartLinePosition.Line)
```
pdelvo: What happens if I write
```csharp
this.Foo("bar".
ToString())
```
Then the StartLinePositions would not match

pdelvo:
What about Delegate declarations, anonymous delegates, lambdas? Should these also be handled by this diagnostic?

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
macpak: Hmm, I'd report this. What's your opinion ?

macpak:
StyleCop description doesn't include them. The question is if WE want to include them here.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs
sharwell: I wouldn't report this. I read this as the closing parenthesis must be on the same line as the *last character* of the last argument.

sharwell:
Well, since we are already considering invocations of each of these items, it would make sense to include their declarations as well. Make sure attributes get included as well.

macpak:
Ok. We have a family of SAs which are related to one another (like #44, #47, #48). They all mention method/indexer call/declaration, the same way this one does. So I guess the same rule should apply to those issues as well - we check additonal constructs, like delegate declarations, anonymous delegates, lambdas etc.

pdelvo:
I think so. These rules should apply to everything that Roslyn calls ParameterList or ArgumentList to be consistent.

macpak:
Ok. So we actually make it a little bit different than in original StyleCop, I think that rule should apply to other issues - if we think that sth should be reported but is not in original StyleCop, we should add this.

sharwell:
@macpak For now you can leave a comment in #359 if you observe a difference. Eventually we'll have a document in the repository where notes can be added.

macpak:
Allright. Thanks for info.

macpak:
I added code to analyze missing parts. Let me know if you think that something else should be also analyzed by this analyzer.

sharwell:
Sweet! :sparkles: 

This one is next up on my plate for review. Plus it should lay the groundwork for many of the other readability rules. :+1:

macpak:
Yes :) :+1:

macpak:
I found that tests for attribute use an attribute that is not supposed to be applied on a class. I applied it on a method to make code "more correct" - actually it doesn't change anything but it's more correct :)

sharwell:
I went ahead and merged this. So you're aware in the future, I did make a few small changes to the implementation.

1. To keep the diagnostic messages clear and short, you generally want to either duplicate the `Title` value or write a short custom message like you see in SA1310. The description is longer than users will want to read.
2. I placed the `using` directives inside the `namespace` declaration to meet the requirements of SA1200 (not yet enforced, but hopefully will be enabled by default starting with the Alpha 2 release).
3. I changed the layout of a few conditional expressions in **SA1111ClosingParenthesisMustBeOnLineOfLastParameter.cs**, which you can see by using the [blame tool on GitHub]().
4. I enabled the diagnostic by default, because once again you took the time to implement a nice set of tests. :+1:

macpak:
Ok. Thanks.
Just one case - found that conditional expressions had bug (grrrrr) because they should've had ORs instead of combination of OR's and ANDs. For instance, these lines:

     if (lambdaExpressionSyntax.ParameterList == null ||
                !lambdaExpressionSyntax.ParameterList.IsMissing &&
                !lambdaExpressionSyntax.ParameterList.Parameters.Any())
            {
                return;
           }
should be 

    if (lambdaExpressionSyntax.ParameterList == null ||
                !lambdaExpressionSyntax.ParameterList.IsMissing ||
                !lambdaExpressionSyntax.ParameterList.Parameters.Any())
            {
                return;
           }
. I reported it on #423

PR #404, opened 2015-01-19
Trivia extensions

sharwell:
* `WithTriviaFrom`: Copy leading and trailing trivia from one `SyntaxNode` or `SyntaxToken` to another.
* `WithoutTrivia`: Remove the leading and trailing trivia from a `SyntaxNode` or `SyntaxToken`.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SpacingExtensions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SpacingExtensions.cs
@@ -44,6 +44,32 @@
         }
 
         /// <summary>
+        /// Removes the leading and trailing trivia associated with a syntax node.
+        /// </summary>
+        /// <typeparam name="TNode">The type of syntax node to remove trivia from.</typeparam>
+        /// <param name="node">The syntax node to remove trivia from.</param>
+        /// <returns>A copy of the input syntax node with leading and trailing trivia removed.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="node"/> is <see langword="null"/>.</exception>
+        public static TNode WithoutTrivia<TNode>(this TNode node)
+            where TNode : SyntaxNode
+        {
+            if (node == null)
+                throw new ArgumentNullException(nameof(node));
+
+            return node.WithoutLeadingTrivia().WithoutTrailingTrivia();
+        }
+
+        /// <summary>
+        /// Removes the leading and trailing trivia associated with a syntax token.
+        /// </summary>
+        /// <param name="token">The syntax token to remove trivia from.</param>
+        /// <returns>A copy of the input syntax token with leading and trailing trivia removed.</returns>
+        public static SyntaxToken WithoutTrivia(this SyntaxToken token)
+        {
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SpacingExtensions.cs
```diff
@@ -43,6 +44,71 @@ public static SyntaxTriviaList WithoutWhitespace(this SyntaxTriviaList syntaxTri
...
+        /// Removes the leading and trailing trivia associated with a syntax node.
+        /// </summary>
+        /// <typeparam name="TNode">The type of syntax node to remove trivia from.</typeparam>
+        /// <param name="node">The syntax node to remove trivia from.</param>
+        /// <returns>A copy of the input syntax node with leading and trailing trivia removed.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="node"/> is <see langword="null"/>.</exception>
+        public static TNode WithoutTrivia<TNode>(this TNode node)
+            where TNode : SyntaxNode
+        {
+            return node.WithoutLeadingTrivia().WithoutTrailingTrivia();
```
pdelvo: It might be a good idea to check node for null here to get a consistend stack frame with the other extension methods.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SpacingExtensions.cs
```diff
@@ -43,6 +44,71 @@ public static SyntaxTriviaList WithoutWhitespace(this SyntaxTriviaList syntaxTri
...
+        /// <typeparam name="TNode">The type of syntax node to apply the trivia to.</typeparam>
+        /// <param name="targetNode">The syntax node to update.</param>
+        /// <param name="node">The syntax node with the desired leading and trailing trivia.</param>
+        /// <returns>A copy of <paramref name="targetNode"/> with the leading and trailing trivia replaced with the
+        /// trivia from <paramref name="node"/>.</returns>
+        /// <exception cref="ArgumentNullException">
+        /// <para>If <paramref name="targetNode"/> is <see langword="null"/>.</para>
+        /// <para>-or-</para>
+        /// <para>If <paramref name="node"/> is <see langword="null"/>.</para>
+        /// </exception>
```
pdelvo: I love how much work you put into your documentation!

pdelvo:
Looks good. But I really think we should put these helpers/extensions in a different namespace. You said you did not like the name "Helpers" for the namespace. Do you have a better name suggestion for it?

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SpacingExtensions.cs
sharwell: :book:

sharwell:
> But I really think we should put these helpers/extensions in a different namespace.

I plan to evaluate all the common and shared code we created during the initial development and refactor it after Preview 1.

pdelvo:
Sounds good

sharwell:
I went ahead and merged this so @macpak can use it in #400.

PR #409, opened 2015-01-21
Implement SA1642

### LINKED ISSUE #162: SA1642: ConstructorSummaryDocumentationMustBeginWithStandardText
http://www.stylecop.com/docs/SA1642.html

pdelvo:
Fixes #162 (ConstructorSummaryDocumentationMustBeginWithStandardText)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
@@ -1,0 +1,246 @@
+﻿namespace StyleCop.Analyzers.Test.DocumentationRules
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using StyleCop.Analyzers.DocumentationRules;
+    using TestHelper;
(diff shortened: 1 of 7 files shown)


sharwell:
No. Way. :open_mouth:

sharwell:
And it even comes with a code fix! :open_mouth:

pdelvo:
Of course it does! I don't want to write these texts myself ;) The code fix was actually very painfull to write. Maybe Im missing an easier way to do that

pdelvo:
Why is this diagnostic so important in particular btw?

sharwell:
Because [I do not joke around with my documentation](https://github.com/rackerlabs/dotnet-threading/blob/master/Rackspace.Threading/CoreTaskExtensions.cs). :+1: 

This serves two purposes:

1. It verifies correctness of something that is not wrong very often but requires *way too much* reading to find those cases.
2. It provides a reference example for this type of behavior (checking the text of documentation associated with certain elements). I plan to, for example, provide an analyzer for implementations of `IHttpApiCall<T>` [like this one](https://github.com/openstacknetsdk/openstack.net/blob/master/src/OpenStack.Net/OpenStack/Services/Identity/V2/AuthenticateApiCall.cs) which use a standard form of text. There will be hundreds of these classes in that SDK when it's complete.

sharwell:
We're going to need to tweak this a bit from the rules established by StyleCop. For now I'm taking the following approach:

1. Update the code to allow additional text at the end of the first sentence. For example, you could have this:

    > Initializes a new instance of the &lt;see cref="MyType"/&gt; class using A and B.

2. Merge this pull request.

3. File issues for discussion and/or work on modifications to the behavior of this analysis rule.

PR #411, opened 2015-01-22
In tests only check distinct diagnostics

### LINKED ISSUE #366: Bug in tests: duplicate reporting of diagnostics
The tests for #350 require a workaround because the test framework is reporting some diagnostics multiple times. The multiple reports *do not* appear when the analyzer is run on equivalent code within Visual Studio.

pdelvo:
Fixes #366.
I actually found a bug in Roslyn while fixing this. https://github.com/dotnet/roslyn/issues/57
So I implemented a minor work around until this is fixed.


sharwell:
It still doesn't explain *why* we were seeing duplicates. But it does make tests for this project easier.

PR #412, opened 2015-01-23
Dogfooding

sharwell:
So, should we start using this?

:sunglasses:


pdelvo:
We should! Looks good

PR #417, opened 2015-01-24
Allow the default text for non private constructors to be used on private constructors

### LINKED ISSUE #413: SA1642 (summary text) misleading for private constructors
Currently the following constructor results in SA1642:

```csharp
/// <summary>
/// Initializes a new instance of the <see cref="ApiStatus"/> class with the specified name.
/// </summary>
private ApiStatus(string name)
    : base(name)
{
}
```

The expected text is the following:

> Prevents a default instance of the &lt;see cref="ApiStatus"/&gt; class from being created.

The currently expected text is simply not correct in this case.

**Proposal**

1. SA1642 should allow *either* the "non-private" or "private" forms for private constructors.
2. When generating text for a private constructor, the current "non-private" form should be used as the template for this text. Users who are writing new APIs and truly wish to prevent an instance of the type from being created can do so by making `static class`. We can safely assume that users writing private instance constructors are actually just controlling the manner in which instances are created.

pdelvo:
Fixes #413

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
@@ -150,9 +150,22 @@
         }
 
         [TestMethod]
+        public async Task TestPrivateConstructorCorrectDocumentationSimpleVariation()
+        {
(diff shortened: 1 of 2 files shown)


pdelvo:
Sometimes I think im blind. There is already a PR for this in #416.

sharwell:
Sorry about that, I didn't mean to waste your time. :frowning: 

I ended up working on it because I couldn't sleep. :sleepy:

pdelvo:
It's not your fault if im blind ;)

PR #425, opened 2015-01-28
Implement SA1608

### LINKED ISSUE #128: SA1608: ElementDocumentationMustNotHaveDefaultSummary

pdelvo:
Fixes #128.
I implemented it the way StyleCop does. It only reports the issue if the summary starts with the default text, not if it appears in the middle.


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1608ElementDocumentationMustNotHaveDefaultSummary.cs
```diff
@@ -50,7 +55,35 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+            {
+                var summaryElement = XmlCommentHelper.GetTopLevelElement(documentationTrivia, XmlCommentHelper.SummaryXmlTag) as XmlElementSyntax;
+
+                if (summaryElement != null)
+                {
+                    var textElement = summaryElement.Content.FirstOrDefault() as XmlTextSyntax;
+
+                    if (textElement != null)
+                    {
+                        string text = XmlCommentHelper.GetText(textElement);
```
sharwell: You should call `GetText(textElement, true)` here. Setting `normalizeWhitespace` to true covers cases where there are multiple spaces and/or newline characters words.

pdelvo:
I changed that

PR #441, opened 2015-02-02
Implement a FixAllProvider for SA1407 and SA1408

pdelvo:
This adds back the fix all functionality and fixes the bug we had with it by implementing a `FixAllProvider`.


sharwell:
For future reference, this was motivated by dotnet/roslyn#199.

Issue #432, opened 2015-01-31
Code fix: SA1303 (ConstFieldNamesMustBeginWithUpperCaseLetter)


sharwell:
Implement a code fix for #79.

macpak:
grabbed

sharwell:
Someone's going to town with the `Renamer` :+1:

macpak:
For SAs "*WithUpperCaseLetter" I can use codefix from #429 :)

pdelvo:
If @macpak is okay with it im grabbing this

PR #443, opened 2015-02-02
Refactoring for rename code fixes

sharwell:
* Share code
* Improve efficiency through lazy calculation of code fix result
* Include the RenameAnnotation to invoke the rename UI

@macpak I'm assigning this pull request to you for code review, since you originally implemented both of these code fixes. For reference, your unit tests did a very good job at catching some mistakes I made while implementing this. :+1:


macpak:
Hi Sam, great work. I will definitely look at it :)

macpak:
Seems ok. I will also use this RenameHelper for other fix, like  #432

PR #446, opened 2015-02-03
Rewrite SA1100 analysis using the speculative semantic model

sharwell:
Following feedback in dotnet/roslyn#200, this pull request updates SA1100 to use the speculative semantic model instead of creating new `Compilation` instances.


sharwell:
I noticed a few additional problems with the code fix for this diagnostic, so I fixed them in the latest update to this pull request.

pdelvo:
Looks good :+1: GetSpeculativeSymbolInfo seems very usefull

sharwell:
It does seem useful, although I'm not sure we'll need to use it for any other rules. In particular, I didn't need it for #447.

PR #447, opened 2015-02-03
Implement diagnostic and code fix for SA1101 PrefixLocalCallsWithThis

### LINKED ISSUE #35: SA1101: PrefixLocalCallsWithThis
http://www.stylecop.com/docs/SA1101.html

sharwell:
Fixes #35

The number of test cases is large to say the least. Basically I need to choose one item from each of the categories below and build a test for that set of conditions.

* Element type
  * Type
  * Method
  * Property
  * Field
  * Event
  * Parameter
  * Local
* Inheritance
  * New element
  * Override
  * Interface implementation
  * *Explicit* interface implementation
  * Hides static base element
  * Hides instance base element
* Is Static
  * True
  * False
* Expression type
  * Assignment to element
  * Read from (simple, e.g. `var x = this.Property;`)
    * For methods this is a reference to the method group
  * Read from (qualified, e.g. `var x = this.Property.GetHashCode();`)
  * `nameof` expression
* Expression location
  * Attribute value
  * In static member
  * In instance member
  * Field initializer for static field
  * Field initializer for instance field
  * String interpolation
  * Documentation comment
* Receiver
  * Current instance (or the containing type if Is Static is true)
  * Another instance

This gives a maximum of 4704 tests for complete coverage, although some combinations do not make since (e.g. you cannot assign to an element within a documentation comment or `nameof` expression).

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1101UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1101UnitTests.cs
@@ -1,0 +1,287 @@
+﻿namespace StyleCop.Analyzers.Test.ReadabilityRules
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using StyleCop.Analyzers.ReadabilityRules;
+    using TestHelper;
+
+    [TestClass]
+    public class SA1101UnitTests : CodeFixVerifier
+    {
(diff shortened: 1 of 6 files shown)


sharwell:
I think the only reasonable way to do the testing for this is create a helper method that takes an enum value representing each of the above category and builds a test for that case.

sharwell:
`nameof` poses a remarkably bizarre set of challenges for this feature.

* The `nameof` expression can be qualified with `this` when the target of the expression is an instance member, as long as you are in a context where `this` is available.
* If the target of the expression is a method group, you can use `this` even if none of the methods in the group are instance methods.
* The expression can always be qualified with the name of the containing class, regardless of whether or not the target is an instance member.

sharwell:
Since I really want to get this feature in a release, I'm implemented initial testing on a variety of edge cases in the above list.

sharwell:
I'm merging this one now that tests are in place so it gets included in the Alpha 2 release.

PR #450, opened 2015-02-04
Fix that the SA1119 code fix can produce invalid syntax.

### LINKED ISSUE #449: SA1119 code fix needs to add whitespaces
At least if the code would become invalid otherwise. Example:
```csharp
return(Tmp);
```
The code fix removes the parenthesis which results in
```csharp
returnTmp;
```

pdelvo:
If the opening parenthesis does not have leading trivia in some cases the code fix would merge the leading node (token) and the syntax node (token) in the parenthesis creating invalid syntax.

Fixes #449


pdelvo:
Weird that the test failed. It works on my machine.

sharwell:
Did you catch that the failing test isn't the one you added?

sharwell:
We need a way to check the fix all provider here too. Any ideas?

sharwell:
Superseded by #453

PR #453, opened 2015-02-06
Fix potential spacing problems when removing parentheses

sharwell:
This is an alternative solution to #449. It reuses the tests originally provided in #450.

The new implementation detects a spacing error by writing the new syntax root to a string, and then parsing it to verify that the parsed version is the same as expected. If these trees are not considered equivalent (for any reason), the algorithm assumes a spacing error resulted in the combination of one or more nodes and a space is added where the opening parentheses once was.


pdelvo:
I have a look at this next week. I don't have time right now

sharwell:
Thanks :+1:

sharwell:
It turns out this situation affects other diagnostics as well. I'm changing the milestone to post-alpha 2 to give us more time to find a general solution. For example, the code fix for SA1101 takes this input:

```
return@instanceField;
```

And turns it into this:

```
returnthis.@instanceField;
```

sharwell:
Superseded by #646.

PR #457, opened 2015-02-08
Tests for SA1001

camilojd:
Tests for CommasMustBeSpacedCorrectly. 

I got coverage for all cases in SA1001CommasMustBeSpacedCorrectly, but no luck covering lines 61 and 69 in the CodeFixProvider: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/7072b2ad0d41aa2a86794c012834c308e5fa63f5/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1001CodeFixProvider.cs

It is my understanding that in `CodeFixVerifier:VerifyFixAsync`, fixes are applied only if diagnostics were reported from analyzed sources. So, as things like `Func<,>` get ignored in `SA1001CommasMustBeSpacedCorrectly`, should them exist in `SA1001CodeFixProvider`?

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
@@ -1,0 +1,132 @@
+﻿namespace StyleCop.Analyzers.Test.SpacingRules
+{
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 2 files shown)


sharwell:
I love that you are making me think (about why I implemented it the way I did) :+1: 

We'll definitely want to include a test for the case described in #257.

I'm still reviewing this :smile:

sharwell:
Remember this code fix can be invoked as a result an error before the comma, after the comma, or both. The code fix needs to analyze everything again to determine what the specific cause is. You can hit line 69 with either of the following:

```
Func< ,>
Func< ,,>
```

You can hit line 61 with the following:

```
Foo(a ,
b);
```

sharwell:
These are fine tests, so I went ahead and merged this. I'm submitting a separate pull request to cover the additional edge cases I identified.

camilojd:
Cool, thanks Sam! :+1: 
Some edge cases are tricky to get right :smile:

PR #458, opened 2015-02-09
SA1001 tests

sharwell:
* Additional tests to cover cases identified during review of #457 
* Fix handling of commas at the beginning of a line (#257)
* Enable SA1001 by default now that tests are in place

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace StyleCop.Analyzers.Test.SpacingRules
 {
+    using System;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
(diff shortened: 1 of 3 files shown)


sharwell:
@camilojd Can you review this?

sharwell:
I'm merging this one so it gets included in the Alpha 2 release where it will be more extensively tested. I still value your code review opinion @camilojd :+1:

camilojd:
@sharwell  Yes! Will review in the next few hours :+1:

sharwell:
@camilojd Sorry if that sounded like I was pressuring you to hurry. I've already merged the code; the code review could come any time you get around to it and you can let me know at that time if you found any bugs. :sunglasses:

camilojd:
@sharwell No problem at all! :grin:

camilojd:
@sharwell I missed the single, correct case of empty diagnostics with `f(a, b);` (which is also being checked by `TestSpaceBeforeComma`, but I prefer to be explicit :sunglasses:).
I have a pull request -- #475 -- for the former case which includes a change for `TestLastCommaInLine`, now using `{Environment.NewLine}` and looking better.

Besides that, tests look complete!

According to SA1001 things like `System.Func<, >` aren't errors, so diagnostics behave correctly in that case too.

The wording of StyleCop rules can be confusing, as for example: *"comma should always be followed by a single space, unless it is the last character on the line"* can be understood as "when a comma is the last character on the line, it should not be followed by a single space", but really means "when a comma is the last character of the line, **rule doesn't apply**".

So, I'm taking into account that bit of possible ambiguity to continue implementing more rules :-)

sharwell:
@camilojd Can you post that last comment directly on #475?

camilojd:
@sharwell Yup, makes sense.

PR #467, opened 2015-02-09
Fixes that SA1101 is reported for MemberAccessExpressions

pdelvo:
Fixes #464

PR #468, opened 2015-02-09
Fix InvalidCastException in PrefixLocalCallsWithThis

### LINKED ISSUE #464: InvalidCastException in code fix for SA1101 PrefixLocalCallsWithThis
In the following code (which appears in **SA1304NonPrivateReadonlyFieldsMustBeginWithUpperCaseLetter.cs**), a SA1101 warning is reported for the identifier `Variables`.

```csharp
var variables = syntax.Declaration?.Variables;
```

When the code fix is run, an `InvalidCastException` occurs in the code fix.

sharwell:
Fixes #464, with regression test


sharwell:
This fix is effectively the same as the one in #467 (meaning @pdelvo and I both already agreed that this solves the problem). :+1: 

I'm merging this one since the original issue was assigned. :metal:

Issue #452, opened 2015-02-06
Figure out how to test custom fix all providers


sharwell:
This issue is currently affecting #450 and work being done separately by @giggio.

giggio:
It is working now. Checkout the testing infrastructure on CodeCracker:
https://github.com/code-cracker/code-cracker/blob/f28ec4c16982cd2db3485cb65f56f27d770d2730/test/CSharp/CodeCracker.Test/Verifiers/CodeFixVerifier.cs#L132-L222
And related files.

I got Document and Project fix all changes to work, I did not go into Solution as it is mostly foreach on the projects, as far as I noticed.

sharwell:
Fixed in #1126.

PR #470, opened 2015-02-09
Fix SA1101 batch fix is not robust 

### LINKED ISSUE #466: Invalid fix all result for SA1101 (PrefixLocalCallsWithThis)
When the batch fixer applies the code fix for SA1101 to the following block, the result is not correct.

#### Original code:

```csharp
string fixedTest = @"namespace Namespace
{
    class Type
    {
        void Foo()
        {
            int x = 1 " + Sign + @" " + Sign + @"3;
        }
    }
}
";
```

#### Expected result

```csharp
string fixedTest = @"namespace Namespace
{
    class Type
    {
        void Foo()
        {
            int x = 1 " + this.Sign + @" " + this.Sign + @"3;
        }
    }
}
";
```

#### Actual result:

```csharp
string fixedTest = @"namespace Namespace
{
    class Type
    {
        void Foo()
        {
            int x = 1 " + this.Sign + @" " + Sign + @" " + this.Sign + @"3;
        }
    }
}
";
```

pdelvo:
I generalized the robust fix all provider I wrote for SA1407 and SA1408 to make it work with SA1101 as well.

Fixes #466.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/RobustFixAllProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/RobustFixAllProvider.cs
@@ -1,5 +1,6 @@
-﻿namespace StyleCop.Analyzers.MaintainabilityRules
+﻿namespace StyleCop.Analyzers.Helpers
 {
+    using System;
     using System.Collections.Generic;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
@@ -9,9 +10,24 @@
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using SpacingRules;
 
-    internal sealed class SA1407SA1408FixAllProvider : FixAllProvider
+    internal sealed class RobustFixAllProvider : FixAllProvider
     {
-        private static readonly SyntaxAnnotation NeedsParenthesisAnnotation = new SyntaxAnnotation("StyleCop.NeedsParenthesis");
+        private static readonly SyntaxAnnotation Annotation = new SyntaxAnnotation("StyleCop.Annotations.FixAllAnnotation");
+
+        private string codeActionText;
+
+        private Func<SyntaxNode, SyntaxNode, SyntaxNode> replaceNode;
+
+        public RobustFixAllProvider(string codeActionText, Func<SyntaxNode, SyntaxNode, SyntaxNode> replaceNode)
+        {
+            if (replaceNode == null)
+            {
+                throw new ArgumentNullException(nameof(replaceNode));
+            }
+
+            this.codeActionText = codeActionText;
+            this.replaceNode = replaceNode;
+        }
 
         public override async Task<CodeAction> GetFixAsync(FixAllContext fixAllContext)
         {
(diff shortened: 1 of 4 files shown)


sharwell:
My first concern of course is #452.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/RobustFixAllProvider.cs
```diff
@@ -19,11 +35,11 @@ internal sealed class SA1407SA1408FixAllProvider : FixAllProvider
...
                 var newRoot = await FixAllInDocumentAsync(fixAllContext, fixAllContext.Document);
-                return CodeAction.Create("Add parentheses", fixAllContext.Document.WithSyntaxRoot(newRoot));
+                return CodeAction.Create(codeActionText, fixAllContext.Document.WithSyntaxRoot(newRoot));
 
             case FixAllScope.Project:
                 Solution solution = await GetProjectFixesAsync(fixAllContext, fixAllContext.Project);
-                return CodeAction.Create("Add parentheses", solution);
+                return CodeAction.Create(codeActionText, solution);
 
             case FixAllScope.Solution:
```
sharwell: I'm not confident our handling of solutions supports cases where a single file is included in multiple projects, and the result of applying the fix to a document in one project differs from the result of applying the fix in another project.

For any two projects P1 and P2 which contain the same document D, the result of applying the code fix to D in P1 should be identical to the result of applying the code fix to D in P2, with the following exception. Without loss of generality, assume P1 contains a span of text T which is part of the syntax, but the same span T in P2 is `DisabledTextTrivia`. For the span T:

1. The code fix to D in P2 is expected to not change the span (i.e. no code fix should make any changes to a `DisabledTextTrivia` trivia).
2. For the span T, the code fix to D in P1 is used to produce the final change to the document D.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/RobustFixAllProvider.cs
```diff
@@ -9,9 +10,24 @@
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using SpacingRules;
 
-    internal sealed class SA1407SA1408FixAllProvider : FixAllProvider
+    internal sealed class RobustFixAllProvider : FixAllProvider
```
sharwell: I lean towards calling this `DocumentBasedFixAllProvider` instead, indicating it is a `FixAllProvider` that applies changes to one `Document` at a time.

sharwell:
@pdelvo I'm working on this at a high level for a bit. I suggest putting further work on the Fix All providers on hold until I can get you more details.

pdelvo:
Okay. Im currently continueing to work on the status page

sharwell:
OK I'm back from a meeting. I'm currently working on porting the `BatchFixAllProvider` and `BatchSimplificationFixAllProvider` from Roslyn to this project and generalizing them to operate in a "document at a time" manner. I'll send a pull request once I have the initial work done and then we can discuss the specific additional needs we'll have.

I do plan on including a custom attribute which contains a link to the original source file on GitHub and commit hash of the specific commit I ported, allowing us to be notified if/when the reference code is updated and thus our port may need to be updated.

sharwell:
This was fixed on the other side by dotnet/roslyn#2286.

Issue #21, opened 2014-11-18
SA1015: ClosingGenericBracketsMustBeSpacedCorrectly


sharwell:
http://www.stylecop.com/docs/SA1015.html

PR #475, opened 2015-02-11
Add simple test for no diagnostics on correct space after identifiers. Better implementation of TestLastCommaInLine

camilojd:
@sharwell I missed the single, correct case of empty diagnostics with `f(a, b);` (which is also being checked by `TestSpaceBeforeComma`, but I prefer to be explicit :sunglasses:).
This pull request includes a change for `TestLastCommaInLine`, now using `{Environment.NewLine}` and looking better.

Besides that, tests look complete!

According to SA1001 things like `System.Func<, >` aren't errors, so diagnostics behave correctly in that case too.

The wording of StyleCop rules can be confusing, as for example: *"comma should always be followed by a single space, unless it is the last character on the line"* can be understood as "when a comma is the last character on the line, it should not be followed by a single space", but really means "when a comma is the last character of the line, **rule doesn't apply**".

So, I'm taking into account that bit of possible ambiguity to continue implementing more rules :-)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
@@ -30,6 +30,13 @@
         }
 
         [TestMethod]
+        public async Task TestSpaceAfterComma()
+        {
+            string statement = "f(a, b);";
+            await this.TestCommaInStatementOrDecl(statement, EmptyDiagnosticResults, statement);
+        }
+
+        [TestMethod]
         public async Task TestNoSpaceAfterComma()
         {
             string statementWithoutSpace = @"f(a,b);";
@@ -117,8 +124,7 @@
         [TestMethod]
         public async Task TestLastCommaInLine()
         {
-            string statement = @"f(a,
-              b);";
+            string statement = $"f(a,{Environment.NewLine}b);";
             await this.TestCommaInStatementOrDecl(statement, EmptyDiagnosticResults, statement);
         }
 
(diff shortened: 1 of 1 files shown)


camilojd:
@sharwell I missed the single, correct case of empty diagnostics with `f(a, b);` (which is also being checked by `TestSpaceBeforeComma`, but I prefer to be explicit :sunglasses:).
I have a pull request -- #475 -- for the former case which includes a change for `TestLastCommaInLine`, now using `{Environment.NewLine}` and looking better.

Besides that, tests look complete!

According to SA1001 things like `System.Func<, >` aren't errors, so diagnostics behave correctly in that case too.

The wording of StyleCop rules can be confusing, as for example: *"comma should always be followed by a single space, unless it is the last character on the line"* can be understood as "when a comma is the last character on the line, it should not be followed by a single space", but really means "when a comma is the last character of the line, **rule doesn't apply**".

So, I'm taking into account that bit of possible ambiguity to continue implementing more rules :-)

sharwell:
The handling for `System.Func<,>` is an intentional deviation from the literal wording of the StyleCop rule.

:bulb: Can you add a test that `Func<, >` *does* report a warning, and the code fix changes it to `Func<,>`. In addition, can you verify that `Func<, ,>` is corrected to `Func<,,>`?

PS: See the bottom of [this document](https://github.com/openstacknetsdk/openstack.net/blob/master/CONTRIBUTING.md) for an explanation of common emoji I use in comments.

camilojd:
@sharwell In the case of `Func<, >` the current implementation of the `DiagnosticAnalyzer` doesn't convey a warning for SA1001 and as such, is consistent with StyleCop (tested with build 4.7.49.0). However, the same build of StyleCop generates a SA1015 warning for that input, so we'll have to take that into account when implementing tests for #21. As such, do we still need a test for a result of `EmptyDiagnosticResults` with that input?

However, `Func<, ,>` *does* issue a warning and gets fixed to `Func<,,>`. I added a test for that.

sharwell:
Thanks for taking a closer look. Whenever two diagnostics can be triggered by the same issue, we pick one to report and suppress the other. In this case, since SA1015 is more specific than SA1001, I agree with you on only reporting SA1015.

sharwell:
One more thing - we should add a test for this case to make sure SA1001 is not reported. Make sure to include a comment stating the input triggers as SA1015 warning instead.

camilojd:
@sharwell You're welcome! It's done now.

Issue #469, opened 2015-02-09
Unresolved conflict between rules as applied to StyleCopAnalyzers


sharwell:
Currently StyleCop.Analyzers has several fields named `_supportedDiagnostics`. Warning SA1309 (FieldNamesMustNotBeginWithUnderscore) is reported for these fields. If the code fix for SA1309 is applied, each of these fields is renamed to `supportedDiagnostics`, which does not cause any errors. However, since each of these fields is marked `static readonly`, the new name results in warning SA1311 (StaticReadonlyFieldsMustBeginWithUpperCaseLetter) getting reported. If the code fix for this rule is applied, the new name of the field is `SupportedDiagnostics`. This name conflicts with the existing property name `SupportedDiagnostics`.

This issue is currently open for discussions regarding potential solutions for addressing this situation now as well as a policy for handling similar naming conflicts in the future.

sharwell:
@pdelvo @macpak Please feel free to add descriptions of possible solutions. We'll discuss them all and decide which one to go with. :+1:

sharwell:
## Possible solution: Suppress warnings

In situations like this, SA1311 could be suppressed through the use of a `#pragma` directive or the `SuppressMessage` attribute.

Advantages:

* This allows the field name to clearly represent its relation to the property with the same name (with the exception of the case of the first letter)

Disadvantages:

* It means we are not following StyleCop rules within a project that *implements StyleCop rules*
* It makes the code ugly
* It means there is more work involved with developing the code, and maintenance costs may rise since otherwise simple operations like renaming a variable involve examining the warning suppressions associated with that variable

sharwell:
### Possible solution: Getter-only auto-property with inline initializer

#### Example:

Before:

```csharp
private static readonly ImmutableArray<DiagnosticDescriptor> supportedDiagnostics =
    ImmutableArray.Create(Descriptor);

public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
{
    get
    {
        return supportedDiagnostics;
    }
}
```

After:

```csharp
public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
{
    get;
} = ImmutableArray.Create(Descriptor);
```

#### Advantages:

* Eliminates the problem and provides even tighter association of the underlying value with the property that exposes it
* Does not result in any additional runtime overhead, unlike some other options like eliminating the cached variable or expression-bodied members

#### Disadvantages:

[...]

Issue #437, opened 2015-01-31
Update SA1311 (StaticReadonlyFieldsMustBeginWithUpperCaseLetter) to handle conflicting names


sharwell:
The implementation of SA1311's code fix provided in #429 suffers from the same issue described in #436. This issue affects all code fixes which use `RenameHelper.RenameSymbolAsync`.

Issue #436, opened 2015-01-31
Update SA1302 (InterfaceNamesMustBeginWithI) to handle conflicting names


sharwell:
The following code results in an SA1302 warning. If the code fix is applied, it introduces a new compiler error.

```csharp
public interface Foo { }
public interface IFoo { }
public class Bar : Foo { }
```

We first need to decide how we want to handle this case (and similar), followed by implementation and tests.

macpak:
First of all we need to check if a type with the new name exists in the same namespace. If it's in a different namespace - it's safe to apply the fix. If the type aleady exists in the namespace, maybe we should append a number to the name, to make it unique  (of course, if we have Foo interface, and IFoo and IFoo2  already exist, we should rename it to IFoo3).

sharwell:
We *could* do that. First I'm curious if there is a way to check with `Renamer` to see if the rename operation introduced new errors, and cancel the code fix if it did.

macpak:
Isn't it done in #443 ?

sharwell:
No, that pull request makes improvements based on the feedback in dotnet/roslyn#197, but this particular issue did not behave in the manner expected by that team. Hopefully the next release of the code fix framework allows us to notify the user of a conflict prior to applying it.

PR #481, opened 2015-02-13
Improve SA1311 code fix

pdelvo:
Implements the changes discussed in #469.

No tests are implemented because we have not yet fully agreed on the solution.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
@@ -69,23 +69,5 @@
 
             return false;
         }
-
-        internal static string GetTypeName(this TypeDeclarationSyntax typeDeclaration)
-        {
-            if (typeDeclaration is ClassDeclarationSyntax)
-            {
-                return ((ClassDeclarationSyntax)typeDeclaration).Identifier.ToString();
-            }
-            if (typeDeclaration is StructDeclarationSyntax)
-            {
-                return ((StructDeclarationSyntax)typeDeclaration).Identifier.ToString();
-            }
-            if (typeDeclaration is InterfaceDeclarationSyntax)
-            {
-                return ((InterfaceDeclarationSyntax)typeDeclaration).Identifier.ToString();
-            }
-
-            return null;
-        }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
@@ -69,5 +69,23 @@
 
             return false;
         }
+
+        internal static string GetTypeName(this TypeDeclarationSyntax typeDeclaration)
+        {
+            if (typeDeclaration is ClassDeclarationSyntax)
+            {
+                return ((ClassDeclarationSyntax)typeDeclaration).Identifier.ToString();
+            }
+            if (typeDeclaration is StructDeclarationSyntax)
+            {
+                return ((StructDeclarationSyntax)typeDeclaration).Identifier.ToString();
+            }
+            if (typeDeclaration is InterfaceDeclarationSyntax)
+            {
+                return ((InterfaceDeclarationSyntax)typeDeclaration).Identifier.ToString();
+            }
+
+            return null;
+        }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1311UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1311UnitTests.cs
@@ -163,6 +163,96 @@
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
         }
 
+        [TestMethod]
+        public async Task TestClassNameConflict()
+        {
+            var testCode = @"public class Bar
+{
+    public static readonly string bar;
+
+    static Foo()
+    {
+        bar = ""aa"";
+    }
+}";
+
+            var expected = new[]
+            {
+                new DiagnosticResult
+                {
+                    Id = DiagnosticId,
+                    Message = "Static readonly fields must begin with upper-case letter",
+                    Severity = DiagnosticSeverity.Warning,
+                    Locations =
+                        new[]
+                        {
+                            new DiagnosticResultLocation("Test0.cs", 3, 35)
+                        }
+                }
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None);
+
+            var fixedCode = @"public class Bar
+{
+    public static readonly string BarValue;
+
+    static Foo()
+    {
+        BarValue = ""aa"";
+    }
+}";
+
+            await this.VerifyCSharpFixAsync(testCode, fixedCode);
+        }
+
+        [TestMethod]
+        public async Task TestMemberNameConflict()
+        {
+            var testCode = @"public class Foo
+{
+    public static readonly string bar;
+
+    static Foo()
+    {
+        bar = ""aa"";
+    }
+
+    public static readonly string Bar;
+}";
+
+            var expected = new[]
+            {
+                new DiagnosticResult
+                {
+                    Id = DiagnosticId,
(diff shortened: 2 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
```diff
@@ -43,8 +45,69 @@ public override ImmutableArray<string> GetFixableDiagnosticIds()
...
+                    }
+
+                    context.RegisterFix(CodeAction.Create($"Rename field to '{newName}'", cancellationToken => RenameHelper.RenameSymbolAsync(document, root, token, newName, cancellationToken)), diagnostic);
+                }
+            }
+        }
+
+        private bool IsValidNewMemberName(INamedTypeSymbol typeSymbol, string name)
+        {
+            foreach (var member in typeSymbol.GetMembers())
```
sharwell: `typeSymbol.GetMembers(name)`

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
```diff
@@ -43,8 +45,69 @@ public override ImmutableArray<string> GetFixableDiagnosticIds()
...
+            {
+                if (member.Kind == SymbolKind.Method)
+                {
+                    // Field names can't colide with method names
```
sharwell: :exclamation: A class is not allowed to have a field and a method with the same name.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
```diff
@@ -43,8 +45,69 @@ public override ImmutableArray<string> GetFixableDiagnosticIds()
...
+                    }
+                }
             }
+            if (typeSymbol.SpecialType == SpecialType.System_Object)
+            {
+                return true;
+            }
+            else
+            {
+                return this.IsValidNewMemberName(typeSymbol.BaseType, name);
```
sharwell: There is no conflict if a field in a derived class has the same name as a member of the base class.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
```diff
@@ -43,8 +45,69 @@ public override ImmutableArray<string> GetFixableDiagnosticIds()
...
+                {
+                    SemanticModel semanticModel = await document.GetSemanticModelAsync();
+
+                    var typeDeclarationSymbol = semanticModel.GetDeclaredSymbol(typeSyntax) as INamedTypeSymbol;
+                    if (typeDeclarationSymbol == null)
+                    {
+                        return;
+                    }
+
+                    while (!this.IsValidNewMemberName(typeDeclarationSymbol, newName) || newName == typeSyntax.GetTypeName())
```
sharwell: I don't think we need a `while` loop. Further conflicts would be addressed by #436, #437, etc.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
```diff
@@ -43,8 +45,69 @@ public override ImmutableArray<string> GetFixableDiagnosticIds()
...
+            else
+            {
+                return this.IsValidNewMemberName(typeSymbol.BaseType, name);
+            }
+        }
+
+        private TypeDeclarationSyntax GetParentTypeDeclaration(SyntaxToken token)
+        {
+            SyntaxNode parent = token.Parent;
+            while (!(parent is TypeDeclarationSyntax))
```
sharwell: :question: Can you replace the `is`/`cast` here with `as`?

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
pdelvo: If there is a member declared in a base class and the field is getting the same name you will access that field instead of the base class member. Thats a breaking change

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
sharwell: :bulb: I do not think the test against `typeSyntax.GetTypeName()` is necessary. However, if it is then I would move it into `IsValidNewMemberName`.

The reason I don't think it's needed is we are making a very special case here for when a field is specifically used and named in order for it to be a backing field for a property (or property-like method). In any other case, the `someName` &rarr; `SomeNameValue` may or may not make sense, and I would prefer to leave this determination for the user.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
```diff
@@ -69,5 +69,23 @@ internal static bool IsContainedInNativeMethodsClass(SyntaxNode syntax)
 
             return false;
         }
+
+        internal static string GetTypeName(this TypeDeclarationSyntax typeDeclaration)
```
sharwell: :bulb: Perhaps leave this method for another time?

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
sharwell: > the renamer is smart...

Yes, yes it really, really is! :sunglasses:

pdelvo:
I applied the changes you suggested

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
```diff
@@ -43,8 +47,58 @@ public override ImmutableArray<string> GetFixableDiagnosticIds()
...
+                {
+                    SemanticModel semanticModel = await document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+
+                    var typeDeclarationSymbol = semanticModel.GetDeclaredSymbol(typeSyntax) as INamedTypeSymbol;
+                    if (typeDeclarationSymbol == null)
+                    {
+                        return;
+                    }
+
+                    while (!this.IsValidNewMemberName(typeDeclarationSymbol, newName) || newName == typeDeclarationSymbol.Name)
```
sharwell: Can you change this line to the following?

```csharp
if (!this.IsValidNewMemberName(typeDeclarationSymbol, newName))
```

sharwell:
With the one remaining change I suggested in place, I think we are good to go on the implementation :+1:

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1311CodeFixProvider.cs
pdelvo: I agree with putting the name check into the IsValidNewMemberName method. But without the while the code fix may suggest a code fix that will not compile if there are still name conflicts after the fix.

sharwell: It's OK for it to propose a name that conflicts. It just means the user is responsible for choosing an appropriate name.

The user experience surrounding conflicting names is unpleasant, but that's another issue that will likely need to wait for a change in Roslyn.

sharwell: @pdelvo As a further clarification, I am not opposed to expanding the algorithm for this code fix in the future, but I think each move in that direction should be narrowly scoped and designed to address a particularly common scenario. For this pull request, the only situation I'm hoping to cover is the simplest case where a backing field and property differ only by the casing of the first letter. :+1:

sharwell:
:bulb: You might consider enabling the batch fix all provider for this. I saved a great deal of time creating #491 by enabling the fix all provider beforehand.

pdelvo:
I implemented your additional feedback, rebased everything on master and implemented tests. I have no idea why this one "contains merge conflicts" because merging this into master would be a fast forward.

Issue #57, opened 2014-11-18
SA1124: DoNotUseRegions


sharwell:
http://www.stylecop.com/docs/SA1124.html

pdelvo:
grabbed

PR #486, opened 2015-02-16
Finish SA1123

### LINKED ISSUE #246: Diagnostic unit tests: DoNotPlaceRegionsWithinElements
Implement unit tests for #56.

### LINKED ISSUE #247: Code fix: DoNotPlaceRegionsWithinElements
Implement code fix for #56.

pdelvo:
I changed the diagnostic to not report some uses of regions where SA1124 should be reported instead. I implemented a code fix that removes regions (for SA1123 and SA1124). and I implemented tests for SA1123.

The diagnostic now has a handy method which can be used by SA1124 to exclude cases where SA1123 should be reported instead of SA1124.
With that implementing SA1124 should be very easy.

Fixes #246. Fixes #247.


pdelvo:
I updated this PR (switched to xunit, addressed changed api).

#57 is blocked on this and trivial when this is merged.

PR #488, opened 2015-02-17
Upgrade code analysis to 1.0.0rc1

pdelvo:
Do not merge.

Currently all SA1400 tests that require a syntax node action of type EventFieldDeclaration or FieldDeclaration being called are failing.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1400UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1400UnitTests.cs
@@ -2,11 +2,12 @@
 {
     using System.Threading;
     using System.Threading.Tasks;
+    using Analyzers.MaintainabilityRules;
+    using Helpers;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.Diagnostics;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
-    using StyleCop.Analyzers.MaintainabilityRules;
     using TestHelper;
 
     [TestClass]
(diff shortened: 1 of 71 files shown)


pdelvo:
Im pretty sure that it is a bug in roslyn. I opened an issue for that https://github.com/dotnet/roslyn/issues/565.

sharwell:
So yes it's a bug in Roslyn. How should we handle it in the meantime?

pdelvo:
I think we should merge this. The bug does not crash anything and it does not produce unnecessary diagnostics. Currently this project does not work in ctp 6 at all so having most of the diagnostics work is an improvment. The bug is fixed so the unit tests and the diagnostics should work as expected in 1.0.0 rc2. 

Should I add an ignore  attribute (and a WorkItem attribute) to the test methods?

sharwell:
No need to create a new work item attribute at this time. I imagine we will eventually switch to xUnit.net  and create one at that time.

Yes, we'll need to ignore these:

```csharp
[Ignore("https://github.com/dotnet/roslyn/issues/565")]
```

pdelvo:
I disabled the tests and created a new issue so that i could link to it. The visual studio test framework does have an attribute that we could use (WorkItemAttribute), but it only takes an integer as parameter so I would not be able to put a link in there.

pdelvo:
I rebased the branch on master because the merged built failed because there was work added on master that used CSharpKind() instead of Kind()

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1400UnitTests.cs
```diff
@@ -487,37 +488,37 @@ public class SA1400UnitTests : CodeFixVerifier
 
         #region EventFieldDeclarationSyntax
 
-        [TestMethod]
+        [TestMethod, Ignore, OpenIssue("https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/496")]
```
sharwell: :bulb: I typically prefer these one per line.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1212PropertyAccessorsMustFollowOrder.cs
```diff
@@ -91,8 +91,8 @@ private static void AnalyzeProperty(SyntaxNodeAnalysisContext context, BasePrope
                 return;
             }
 
-            if (accessors[0].CSharpKind() == SyntaxKind.SetAccessorDeclaration &&
-                accessors[1].CSharpKind() == SyntaxKind.GetAccessorDeclaration)
+            if (accessors[0].Kind() == SyntaxKind.SetAccessorDeclaration &&
```
sharwell: :bulb: This is a good place for `IsKind`.

sharwell:
It looks good :+1: 

The last items I can easily fix during a merge later today.

PR #500, opened 2015-02-26
SA1503 implementation (incl. tests and codefix)

### LINKED ISSUE #104: SA1503: CurlyBracketsMustNotBeOmitted
http://www.stylecop.com/docs/SA1503.html

vweijsters:
Fixes #104


StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1503UnitTests.cs
```diff
@@ -0,0 +1,340 @@
...
+            await this.TestCodeFixForStatement(ForTestStatement);
+        }
+
+        [TestMethod, TestCategory("MaintainabilityRules/SA1503")]
+        public async Task TestCodeFixProviderForForEachStatement()
+        {
+            await this.TestCodeFixForStatement(ForEachTestStatement);
+        }
+
+        // VW: Not sure if this is a valid test at all.
```
sharwell: :bulb: This is an awesome test. You can remove this comment if you want.

sharwell:
It's really cool that you made sure to avoid changing code formatting, even if the code isn't indented in the expected manner. However, I'm concerned that the new implementation might not provide the expected behavior for the following code:

```csharp
if (i == 0) if (i == 0) Debug.Assert(true);
```

I still have some work to do on the review since I haven't thought about the impact of things like comments and preprocessor directives on this implementation.

vweijsters:
There are issues with the code formatting, working on a fix.

vweijsters:
Cleaned up the code and made the code fix much simpler. Indentation is left up to the Roslyn code, which will do a fine job when the file being fixed is using the indentation rules for the solution that it is in.

Keeping correct indentation for source files that deviate from the solution indentation rules is extremely difficult (and error prone) and in my opinion therefore not desirable.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1503UnitTests.cs
```diff
@@ -0,0 +1,466 @@
...
+        protected override CodeFixProvider GetCSharpCodeFixProvider()
+        {
+            return new SA1503CodeFixProvider();
+        }
+
+        protected override Solution CreateSolution(ProjectId projectId, string language)
+        {
+            var newSolution = base.CreateSolution(projectId, language);
+
+            newSolution.Workspace.Options = newSolution.Workspace.Options.WithChangedOption(FormattingOptions.IndentationSize, language, this.indentSize);
```
sharwell: :question: Why did this need to be done?

vweijsters: I added the code allow me to test with different indentation sizes. But maybe I'm just retesting Roslyn functionality, which is not really useful. The indentation test could prove useful when an alternate implementation is made for the code fix, that doesn't rely on Roslyn that much.

pdelvo: It would not be the first time our unit tests catch a roslyn bug :smile:

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1503CodeFixProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+            return this.ReplaceStatementWithWrappedStatement(root, forStatement, forStatement.CloseParenToken, forStatement.Statement);
+        }
+
+        private SyntaxNode FixForEachStatement(SyntaxNode root, SyntaxNode parent)
+        {
+            var foreachStatement = (ForEachStatementSyntax)parent;
+            return this.ReplaceStatementWithWrappedStatement(root, foreachStatement, foreachStatement.CloseParenToken, foreachStatement.Statement);
+        }
+
+        private SyntaxNode ReplaceStatementWithWrappedStatement(SyntaxNode root, SyntaxNode parent, SyntaxToken parentEndToken, StatementSyntax statement)
```
pdelvo: The parent and parentEndToken parameters can be removed.

pdelvo:
My favorite topic: preprocessor directives.
Your code fix will currently transform this code
```csharp
            if (true)
#if DEBUG
                System.Console.WriteLine("Foobar");
#endif
            System.Console.WriteLine("Foobar 2");
```
into
```csharp
            if (true)
            {
#if DEBUG
                System.Console.WriteLine("Foobar");
            }
#endif
            System.Console.WriteLine("Foobar 2");
```

which will break if topic is not defined. For comments and other preprocessor directives this should work fine though.
There is an easy and a difficult solution for this:

1. Search the expression of the if statement for ConditionalDirectiveTriviaSyntax and if there is one thats partially in, and partially out of the expression dont offer a fix. 
2. Search the expression of the if statement for ConditionalDirectiveTriviaSyntax and if there is one thats partially in, and partially out of the expression move the syntax to still create valid code. This would be definitly more work and Im not sure that this is going to be really usefull in the end. If someone has code like that they are probably not going to want StyleCop in the first place ;)

vweijsters:
I took the easy way out and code fixes are no longer offered if preprocessor directives are in play.

vweijsters:
Appveyor seems to have failed in resolving xunit nuget packages. 
Is there a way to retrigger the AppVeyor build to see if the error is temporary one?

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1503UnitTests.cs
sharwell: :exclamation: I would prefer to consider the changes to the test framework separately from the implementation of SA1503.

vweijsters: See pull request #522.
I'm leaving the code in here for now so that this can still be built.

sharwell:
I merged everything except the changes in #522. There is one test which relied on the ability to set these properties, so I marked it as skipped with a reference to #660.

PR #502, opened 2015-02-26
Implemented SA1518 (incl. tests and codefix)

### LINKED ISSUE #119: SA1518: CodeMustNotContainBlankLinesAtEndOfFile
http://www.stylecop.com/docs/SA1518.html

vweijsters:
Fixes #119


pdelvo:
Very cool! This looks good, but I have a few remarks im going to add

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+    {
+        Debug.Assert(true);
+    }
+}";
+        
+        /// <summary>
+        /// Verifies that the analyzer will properly handle an empty source.
+        /// </summary>
+        /// <returns></returns>
+        [TestMethod, TestCategory("MaintainabilityRules/SA1518")]
```
pdelvo: We have not used TestCategory anywhere else. Maybe we should either revert this back or add this at the other tests as well. Does this make these unit tasts stand out e.g. in the test explorer?

vweijsters: I added the TestCategory to make my tests easier to find in the test explorer. 
And reading your comment made me realize I used MaintainabilityRules instead of LayoutRules as well :(

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1518CodeMustNotContainBlankLinesAtEndOfFile.cs
```diff
@@ -43,7 +44,17 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
         {
-            // TODO: Implement analysis
+            context.RegisterSyntaxTreeAction(this.HandleSyntaxTreeAction);
+        }
+
+        private void HandleSyntaxTreeAction(SyntaxTreeAnalysisContext context)
+        {
+            var lastToken = context.Tree.GetRoot().GetLastToken(includeZeroWidth: true);
+
+            if (lastToken.HasLeadingTrivia)
```
pdelvo: Not all trivias are whitespaces. A comment at the end of the file is reported too or the endif here:
```csharp
#if true
namespace Foo
{
}
#endif
```

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
pdelvo: No problem!
I think adding these to all other tests would make testing easier. You can group by class in the test explorer which would group the same tests together as well, but you don't see the category there so it is not as clear. I whish you could apply the TestCategory attribute to classes but unfortunatly thats not possible.

pdelvo:
Because github doesn't let me to add comments in code if they are not changed:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/502/files#diff-49a90c5ccb5d1642fed439776c5101f0L29
The AnalyzerConstants.DisabledNoTests can be replaced by true to enable the test by default for Release builds. Otherwise it only works when the project is built in Debug mode.

vweijsters:
I'll be working on fixes then :) Thanks for pointing out the issues.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
sharwell: Since the tests are all placed in classes according to the name of the tested diagnostic, I don't think it makes sense to add this test category at this time. You can use Test Explorer to group tests according to class.

image

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        public async Task TestWithSingleCarriageReturnLineFeedAtEndOfFile()
+        {
+            var testCode = BaseCode + "\r\n";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Verifies that a no carriage return / linefeed at the end of the file will not produce a warning.
```
sharwell: :bulb: You might add a comment that SA1513 is reported for this case.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+    public void Bar(int i)
+    {
+        Debug.Assert(true);
+    }
+}";
+        
+        /// <summary>
+        /// Verifies that the analyzer will properly handle an empty source.
+        /// </summary>
+        /// <returns></returns>
```
sharwell: :bulb: Can you remove the empty `<returns>` elements?

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using TestHelper;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.CodeFixes;
+
+namespace StyleCop.Analyzers.Test.LayoutRules
+{
+    /// <summary>
+    /// Unit tests for the SA1518: CodeMustNotContainBlankLinesAtEndOfFile
```
sharwell: :bulb: `<see cref="SA1518CodeMustNotContainBlankLinesAtEndOfFile"/>`

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// </summary>
+        /// <returns></returns>
+        [TestMethod, TestCategory("MaintainabilityRules/SA1518")]
+        public async Task TestWithoutCarriageReturnLineFeedAtEndOfFile()
+        {
+            await this.VerifyCSharpDiagnosticAsync(BaseCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Verifies that a single carriage return / linefeed at the end of the file will not produce a warning.
```
sharwell: :exclamation: This summary does not match the test behavior

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        /// <summary>
+        /// Verifies that the code fix provider will strip trailing blank lines that include whitespace.
+        /// </summary>
+        /// <remarks>The CRLF after the last curly bracket will not be stripped!</remarks>
+        /// <returns></returns>
+        [TestMethod, TestCategory("MaintainabilityRules/SA1518")]
+        public async Task TestCodeFixProviderStripsTrailingBlankLinesIncludingWhitespace()
+        {
+            var testCode = BaseCode + "\r\n   \r\n   \r\n";
```
sharwell: :bulb: I'd like to see some additional tests using other trivia, such as comments or preprocessor directives, just to make sure we don't accidentally remove something the user wanted to keep.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1518UnitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        /// <summary>
+        /// Verifies that blank lines at the end of the file will produce a warning.
+        /// </summary>
+        /// <returns></returns>
+        [TestMethod, TestCategory("MaintainabilityRules/SA1518")]
+        public async Task TestWithBlankLinesAtEndOfFile()
+        {
+            var testCode = BaseCode + "\r\n\r\n";
```
sharwell: :bulb: You might consider including some additional tests that use `\n` instead of `\r\n` for the line endings.

sharwell:
:bulb: You might consider including some additional tests that use `\n` instead of `\r\n` for the line endings.

sharwell:
I agree with @pdelvo. This is a great contribution. :dancers:

There are a few small things to fix but I have no doubt this will soon be part of the codebase :sunglasses:

vweijsters:
Updated the pull request with the comments above processed.

I'm not convinced that SA1513 should be reported for the mentioned case, so I'll leave that until SA1513 gets implemented.

PR #510, opened 2015-03-02
Implemented diagnostic SA1509: OpeningCurlyBracketsMustNotBePrecededByBlankLine

### LINKED ISSUE #110: SA1509: OpeningCurlyBracketsMustNotBePrecededByBlankLine
http://www.stylecop.com/docs/SA1509.html

macpak:
Fixes #110.


vweijsters:
What's the policy on comments for test methods? The unit tests will trigger a lot of SA1600 messages when it gets enabled.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1509UnitTests.cs
```diff
@@ -0,0 +1,702 @@
...
+                            new[]
+                            {
+                                new DiagnosticResultLocation("Test0.cs", 7, 1)
+                            }
+                    }
+                };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None);
+        }
+
```
vweijsters: I don't think this is a valid test. I had a look and the original StyleCop doesn't trigger on this.

vweijsters:
The test code could be made a fair bit smaller by using a method for generating expected Diagnostic results.

vweijsters:
I did a review and the code looks good. There is a nice extensive set of tests as well :thumbsup:

pdelvo:
We disabled SA1600 for the unit test project. It is enabled for the main project. This would just be too much work.

You are right we really could clean up our tests by writing a method that creates diagnostic results by using some default values we have in most, if not all, of our tests.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1509UnitTests.cs
pdelvo: We don't do everything the same way StyleCop does. We have some differences in places where we think that StyleCop doesn't do the right thing. But I'm not sure whats the best thing to do here.

sharwell: When we do notice behavior deviations, we try to get them at least added to #359. That said, I don't think we should report this case as a diagnostic right now for the following reasons:

1. Users are likely to see this as *not* being preceded by a blank line
2. The code fix for this might not have the desired behavior

It's probably something we can revisit in the future if you find a related case where you do want to see a warning reported.

macpak: I see your point. I'll change it. Thanks !.

macpak: Do you think this is correct ?:

    class Foo
    
    /*this is a comment
    that spans 2 lines */
    //another comment
    {
    
    }

sharwell: At least for now, if the line before the `{` isn't empty then I wouldn't report SA1509.

macpak:
I changed the logic as we discussed above and added code fix.

vweijsters:
The test code needs some work to make it work with the Xunit framework that was introduced in the master.

macpak:
Do we want to use XUnit here? Generally, is Xunit our default Framework for unit tests?

sharwell:
@macpak We recently switched to xunit, which turned out to be surprisingly straightforward (most of the tests didn't require any changes at all). I sided in favor of switching primarily because xunit isn't tied to a particular platform which is good since Roslyn also runs on multiple platforms.

macpak:
Allright. I will look at the code and try to adjust it to XUnit.

sharwell:
We can also do it during the merge if you want to just transition to it for the next implementation you do. :)

macpak:
I can change the tests, no problem:)

macpak:
Changes applied :)

macpak:
Just noticed that helpers for unit tests have been introduced (great work :+1: ). It'd be good to use them for here and #513. I'll add the changes.

Issue #286, opened 2014-12-02
Diagnostic unit tests: UsingDirectivesMustBePlacedWithinNamespace


sharwell:
Implement unit tests for #60.

PR #513, opened 2015-03-07
Implemented diagnostic for SA1214: StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements

### LINKED ISSUE #74: SA1214: StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements
http://www.stylecop.com/docs/SA1214.html

macpak:
Fixes #74.


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
```diff
@@ -43,7 +46,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+
+        private static void AnalyzeType(SyntaxNodeAnalysisContext context, TypeDeclarationSyntax typeDeclaration)
+        {
+            var staticFields = typeDeclaration.Members
+                .OfType<BaseFieldDeclarationSyntax>()
+                .Where(f => f.Modifiers.Any(m => m.IsKind(SyntaxKind.StaticKeyword)))
+                .ToList();
+
+            var endLocationsOfStaticNonReadonlyFields =
+                staticFields.Where(f => f.Modifiers.Any(m => m.IsKind(SyntaxKind.ReadOnlyKeyword)) == false)
```
sharwell: :bulb: `!` is simpler than `== false`

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
```diff
@@ -43,7 +46,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+            var classDeclaration = (ClassDeclarationSyntax) context.Node;
+
+            AnalyzeType(context, classDeclaration);
+        }
+
+        private static void AnalyzeType(SyntaxNodeAnalysisContext context, TypeDeclarationSyntax typeDeclaration)
+        {
+            var staticFields = typeDeclaration.Members
+                .OfType<BaseFieldDeclarationSyntax>()
+                .Where(f => f.Modifiers.Any(m => m.IsKind(SyntaxKind.StaticKeyword)))
```
sharwell: :bulb: You can also just use `f.Modifiers.Any(SyntaxKind.StaticKeyword)`

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
```diff
@@ -43,7 +46,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
         /// <inheritdoc/>
         public override void Initialize(AnalysisContext context)
         {
-            // TODO: Implement analysis
+            context.RegisterSyntaxNodeAction(this.AnalyzeClass, SyntaxKind.ClassDeclaration);
+            context.RegisterSyntaxNodeAction(this.AnalyzeStruct, SyntaxKind.StructDeclaration);
+        }
+
+        private void AnalyzeStruct(SyntaxNodeAnalysisContext context)
```
sharwell: :exclamation: Rather than separate methods `AnalyzeStruct`, `AnalyzeClass`, and `AnalyzeType`, can you create one method `HandleTypeDeclaration` which starts with this:

```csharp
var typeDeclaration  = (TypeDeclarationSyntax)context.Node;
```

You can then register this method as the handler for both `ClassDeclaration` and `StructDeclaration`.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
```diff
@@ -43,7 +46,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+        {
+            var classDeclaration = (ClassDeclarationSyntax) context.Node;
+
+            AnalyzeType(context, classDeclaration);
+        }
+
+        private static void AnalyzeType(SyntaxNodeAnalysisContext context, TypeDeclarationSyntax typeDeclaration)
+        {
+            var staticFields = typeDeclaration.Members
+                .OfType<BaseFieldDeclarationSyntax>()
```
sharwell: :exclamation: `BaseFieldDeclarationSyntax` will include `EventFieldDeclarationSyntax`, which is not what we want (for performance reasons). You should just use `.OfType<FieldDeclarationSyntax>()` here.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
macpak: To be honest I did it deliberately, because for me, when you look at !f.Modifiers.Any(m => m.IsKind(SyntaxKind.ReadOnlyKeyword)) you may omit ! at the beginning.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
```diff
@@ -43,7 +46,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                                                      .Where(v => !v.IsMissing && !v.Identifier.IsMissing)
+                                                      .Select(v => v.Identifier.GetLocation())
+                                                      .ToList(),
+                        FieldEndLinePosition = f.GetLocation().GetLineSpan().EndLinePosition
+                    })
+                    .ToList();
+
+            foreach (var location in readonlyFieldLocations)
+            {
+                if (endLocationsOfStaticNonReadonlyFields.Any(e => e < location.FieldEndLinePosition))
```
sharwell: :exclamation: You don't need all `static` non-`readonly` fields. You only need the first one. Fixing this will change the search algorithm from O(n²) to O(n).

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
macpak: Hmm, I find it more readable to have 2 separate methods, but of course we can have one :+1:

sharwell:
I made several notes in the diff. Before you make the changes, you should think about the following as well (which you may or may not use, but either way it's good to think about).

An alternative implementation could use a single forward pass through the `FieldDeclarationSyntax` members of the type, and have a single state variable `foundStaticNonReadonlyField`. When this state variable is true and a `static readonly` field is encountered, report a warning. When you reach the end of the members, you are done. This works because the members are enumerated in lexigraphical order.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
```diff
@@ -43,7 +46,61 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                {
+                    foreach (var fieldLocation in location.FieldLocations)
```
sharwell: :bulb: Reporting a warning for each variable will likely make a future code fix more difficult, because it will want to move the entire `FieldDeclarationSyntax` at once. I recommend reporting one warning for the entire `FieldDeclarationSyntax` unless you have a compelling reason not to.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
sharwell: :exclamation: I overlooked a *much* bigger issue here. If you use `BaseFieldDeclarationSyntax` then a static event will be treated as a static non-readonly field for the purpose of this diagnostic, which is definitely not what we want.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
sharwell: In some cases I would agree, but considering the simplicity of the file as a whole I don't think this reduction will cause a readability problem.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -0,0 +1,204 @@
+﻿using System.Threading;
```
sharwell: :bulb: As soon as #286 is fixed, this will become a warning.

macpak:
Thanks @sharwell . I will fix it soon.

sharwell:
No problem. Also, I was tougher on this than I normally am. I try to do that every once in a while to keep people thinking outside the box. :smile: You should know that all of the work you've done for this project is very much appreciated. :tada:

macpak:
No problem. Good feedback is a very valuable feedback, If I make some errors/bad desgins, it's always good if someone points them out to me. That's the best way to avoid them in future.

sharwell:
In addition to the items marked in the diff, I made the following changes while merging this:

* Use `AnalyzerConstants.EnabledByDefault`
* Use `.ConfigureAwait(false)` when awaiting a `Task`
* Rename asynchronous methods to include the `Async` suffix

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.cs
```diff
@@ -43,7 +46,55 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
         /// <inheritdoc/>
         public override void Initialize(AnalysisContext context)
         {
-            // TODO: Implement analysis
+            context.RegisterSyntaxNodeAction(this.AnalyzeTypeDeclaration, SyntaxKind.ClassDeclaration, SyntaxKind.StructDeclaration);
```
sharwell: :bulb: This should use `RegisterSyntaxNodeActionHonorExclusions`.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -0,0 +1,156 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.OrderingRules;
+    using TestHelper;
+    using Xunit;
+
+    public class SA1214UnitTests : CodeFixVerifier
+    {
+        public string DiagnosticId { get; } = SA1214StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements.DiagnosticId;
```
sharwell: :bulb: This property is unused and can be removed.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -0,0 +1,156 @@
...
+                this.CSharpDiagnostic().WithLocation(11, 5),
+                this.CSharpDiagnostic().WithLocation(18, 9),
+                this.CSharpDiagnostic().WithLocation(21, 5),
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None);
+        }
+
+
+        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
```
sharwell: :bulb: This needs to be updated for the new test API.

PR #518, opened 2015-03-08
Test cleanup

pdelvo:
I implemented some changes to make our tests less verbose. This is not done because I want to get some feedback on this. The diagnostic result factory lets us remove a lot of the redundant code to create expected diagnostic results.
One problem? with that is that I get the MessageFormat directly from the DiagnosticAnalyzer. That way you dont have to provide it everytime in the tests. But that also means that the MessageFormat itself is no longer tested. We could maybe fix that by having a test that just makes sure that the MessageFormat is correct. I changed the tests for SA1600 to use the new methods to demonstrate the LOC saved.

What do think?

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
@@ -37,23 +37,7 @@ TypeName
 {{
 }}";
 
-            DiagnosticResult[] expected;
-
-            expected =
-                new[]
-                {
-                    new DiagnosticResult
-                    {
-                        Id = this.DiagnosticId,
-                        Message = "Elements must be documented",
-                        Severity = DiagnosticSeverity.Warning,
-                        Locations =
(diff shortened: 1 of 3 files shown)


sharwell:
> Make `CancellationToken` an optional parameter

MSTest has a big limitation where you can't create a test with the following signature:

```
[Test]
public async Task TestName(CancellationToken cancellationToken)
```

If I'm not mistaken, xUnit.net does not have this limitation.

pdelvo:
Does someone have some more feedback on that? I really think the diagnostic result builder is really usefull and I would like to use it for unit tests Im currently working on

sharwell:
Here's my feedback:

1. Since I currently do not want to make the `CancellationToken` change, it would be a good idea to move that to a separate pull request where we can discuss it further, rather than have it delay this one unnecessarily.

2. The VB change is fine. If you move it to a separate pull request it will expedite it.

3. I agree that simplifying the use of `DiagnosticResult` will be very valuable. I'm working on an alternative implementation of this functionality which I'd like to compare with the one here.

sharwell:
Superseded by #525 and #526.

PR #519, opened 2015-03-08
Run the code formatter on all files

pdelvo:
I ran the vs code formatter on all files to remove some unnecessary whitespaces.


sharwell:
I've been leaving many of these on purpose. They'll go away as more of the rules get into an actual release and we use the code fixes.

sharwell:
**wontfix** is a bit misleading. It will be fixed, but not quite yet. :+1:

PR #522, opened 2015-03-11
Added extension point to DiagnosticVerifier

vweijsters:
Added the CreateSolution extension point to the DiagnosticsVerifier, in order to get control over solution properties.


sharwell:
@vweijsters I would prefer to wait on this for now. Do you mind if I close it as `wontfix` until such time as we hit a case that is substantially impacted by it?

vweijsters:
SA1502 (#547) uses this to test the indentation code in the code fixer. 
If we are okay with only testing for 4 spaces indentation in the unit tests, then I'm fine with closing this.

sharwell:
Thanks, I'll keep this open at least until I review that. :smile:

vweijsters:
Merged in upstream/master

pdelvo:
I could use this to add a test for #929.

PR #523, opened 2015-03-11
SA1501 implementation (incl. tests + codefix)

### LINKED ISSUE #102: SA1501: StatementMustNotBeOnSingleLine
http://www.stylecop.com/docs/SA1501.html

vweijsters:
Fixes #102


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
```diff
@@ -95,5 +96,81 @@ internal static int IndexOfTrailingWhitespace(SyntaxTriviaList triviaList)
...
             return (whiteSpaceStartIndex < triviaList.Count) ? whiteSpaceStartIndex : -1;
         }
+
+        /// <summary>
+        /// Strips all trialing whitespace trivia from the trivia list until a non-whitespace trivia is encountered.
+        /// end-of-line trivia are ignored, but kept.
+        /// </summary>
+        /// <param name="triviaList">The trivia list to strip of its trailing whitespace.</param>
+        /// <returns>The modified triviaList.</returns>
+        internal static SyntaxTriviaList WithoutTrailingWhitespace(this SyntaxTriviaList triviaList)
```
sharwell: I have some concerns about potentially confusing this with `SpacingExtensions.WithoutTrailingWhitespace`, but I'm not yet sure what course of action I would recommend.

vweijsters: We should probably merge the two, in order to eliminate the confusion. Not quite sure on how to do that yet.

Just had a look at `SpacingExtensions.WithoutTrailingWhitespace` and I think there is an issue with that code. It strips all whitespace (and optionally end of line tokens) from the trailing trivia, instead of just the ones at the end.

vweijsters:
Updated to the latest state of the project.

PR #524, opened 2015-03-13
Improve assertion messages

pdelvo:
One thing that was annoying me when writing unit tests is that assertion messages where not really usefull. When your tests expects 20 diagnostics and your diagnostic only reports 19 it would only tell you that the number of diagnostics didnt match but it would not tell you which diagnostic was missing. So I was debugging the unit tests and diffing the expected and the actual results in the auto window myself which is really annoying. This change changes the way assertion messages are generated. With this change the VerifyDiagnosticResults will diff the expected results and the actual results of the diagnostic and it will include the changes in the assertion message. This makes writing tests a lot easier.

One drawback with this is that if the severity, the message or any other property of the diagnostic is wrong it will not print the specific assertion message defined in AssertDiagnosticsMatch but it will instead report that one diagnostic is missing, and that one was reported that was not expected. But these mistakes are easy to figure out anyway and they are not really the ones you run into very often.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
@@ -1,12 +1,12 @@
-﻿using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeAnalysis.Diagnostics;
-using Xunit;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Xunit;
+using Xunit.Sdk;
 
 namespace TestHelper
 {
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -135,71 +135,142 @@ private static void VerifyDiagnosticResults(IEnumerable<Diagnostic> actualResult
...
 
-                Assert.True(false,
-                    string.Format("Mismatch between number of diagnostics returned, expected \"{0}\" actual \"{1}\"\r\n\r\nDiagnostics:\r\n{2}\r\n", expectedCount, actualCount, diagnosticsOutput));
+            AppendMissmatchErrorText(actualResults, analyzer, expectedResults, errorStringBuilder);
+
+            string errorString = errorStringBuilder.ToString();
+            if (!string.IsNullOrEmpty(errorString))
+            {
+                Assert.True(false, errorString);
             }
```
vweijsters: 147-151 can be replaced by a single statement:
`Assert.True(errorStringBuilder.Length == 0, errorStringBuilder.ToString());`
or alternatively replace 148-151 with
`Assert.True(string.IsNullOrEmpty(errorString), errorString);`

vweijsters:
I see a lot of
```
if (...)
{
   Assert.True(false, ... );
}
```
constructions in the code. I'm not particularly fond of those and would personally rather see them wrapped up in one statement.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -135,71 +135,142 @@ private static void VerifyDiagnosticResults(IEnumerable<Diagnostic> actualResult
...
+                        expected.Severity, actual.Severity, FormatDiagnostics(analyzer, actual)));
+            }
+
+            if (actual.GetMessage() != expected.Message)
+            {
+                Assert.True(false,
+                    string.Format("Expected diagnostic message to be \"{0}\" was \"{1}\"\r\n\r\nDiagnostic:\r\n    {2}\r\n",
+                        expected.Message, actual.GetMessage(), FormatDiagnostics(analyzer, actual)));
+            }
         }
```
vweijsters: Minor point, but why not verify the Id, Severity and Message before trying to figure out if the locations match.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -135,71 +135,142 @@ private static void VerifyDiagnosticResults(IEnumerable<Diagnostic> actualResult
...
             {
                 var actual = actualResults.ElementAt(i);
                 var expected = expectedResults[i];
 
-                if (expected.Line == -1 && expected.Column == -1)
+                AssertDiagnosticsMatch(analyzer, actual, expected);
+            }
+        }
+
+        private static void AppendMissmatchErrorText(IEnumerable<Diagnostic> actualResults, DiagnosticAnalyzer analyzer, DiagnosticResult[] expectedResults, StringBuilder errorStringBuilder)
```
vweijsters: Typo: AppendMissmatchErrorText should be AppendMismatchErrorText

vweijsters:
The last for statement in VerifyDiagnosticsResult() seems to imply that the actual diagnostics and the expected diagnostics should be matched in the same order. AppendMissmatchErrorText() does a full cross check to allow matchings in random order. 

If the diagnostics should be matched in the same order, a more efficient implementation can be made.

The code outline below shows what I mean

```
// The code below assumes that both expectedDiagnostics and actualDiagnostics are arrays!
var indexActual = 0;
var indexExpected = 0;
var matches = new Dictionary<Diagnostic,Diagnostic>();

while ((indexActual < actualDiagnostics.Length) && (indexExpected < expectedDiagnostics.Length))
{
    if (DiagnosticMatches(expectedDiagnostics[indexExpected], actualDiagnostics[indexActual]))
    {
        matches.Add(expectedDiagnostics[indexExpected], actualDiagnostics[indexActual]);
        indexActual++;
        indexExpected++;
    }
    else
    {
        // detect if there are additional actual diagnostics
        // if there are no matches in the remaining actual diagnostics, 
        // the expected diagnostic is skipped as unmatched.
        for (var i = indexActual + 1; i < actualDiagnostics.Length; i++)
        {
            if (DiagnosticMatches(expectedDiagnostics[indexExpected], actualDiagnostics[i]))
            {
                matches.Add(expectedDiagnostics[indexExpected], actualDiagnostics[i]);
                indexActual = i + 1;
                break;
            }
        }
     
        indexExpected++;
    }
    
    var unmatchedExpectedDiagnostics = expectDiagnostics.Except(matches.Keys);
    var unmatchedActualDiagnostics = actualDiagnostics.Except(matches.Values);

    // TODO: Report the mismatches.
}
```

pdelvo:
Thanks for your feedback!

I adressed all of your feedback except the last one. You are right, that the cross search is not efficient (O(n^2)), but I think that the number of diagnostics we deal with in our tests is so small, that it does not really matter if we use an O(n) or O(n^2) algorithm and I think (maybe its just because I wrote it) that my version is easier to read.

The actual diagnostics array is ordered by the caller and the expected diagnostics array (should) be ordered by the test that uses it.

vweijsters:
You are quick in processing feedback :+1: 

The readability of your code is good, so I'm perfectly fine with leaving stuff as it is.

sharwell:
@pdelvo I'm reviewing this now, but for a quick consideration, how certain are you that this will not cause problems when we move to testing [SA1027](http://www.stylecop.com/docs/SA1027.html)?

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -135,71 +135,120 @@ private static void VerifyDiagnosticResults(IEnumerable<Diagnostic> actualResult
...
+        {
+            foreach (var actual in actualResults)
+            {
+                bool success = false;
+                foreach (var expected in expectedResults)
                 {
-                    if (actual.Location != Location.None)
+                    try
+                    {
+                        AssertDiagnosticsMatch(analyzer, actual, expected);
```
sharwell: :exclamation: Avoid using exception handling for flow control. In this, it will become impossible to distinguish between `XunitException` which was thrown from this code and `XunitException` which is thrown for a legitimate test failure.

sharwell:
:exclamation: In addition to my comment in the diff, the current algorithm allows an item in `actualResults` to match multiple items in `expectedResults` (and vice versa). How certain are we that this behavior will not cause a problem?

pdelvo:
The old and the new version will fail at the same condition. the difference is the way it formats the message for it. If you have a bigger test with a lot of diagnostics that should be created it can be really annoying to figure out which specific diagnostic that should be reported is reported, or which diagnostic is reported that shouldnt be. At least for me these two things are the main reason that a unit test I write fails. So this solution is optimized create an assertion message that helps you figuring out what is going wrong. The old one was better if maybe you expect one diagnostic, it is present, and you are expecting a different servity. But thats not really what happens a lot of times

pdelvo:
The algorithm only allows an item in actualResults to match multiple in expectedResults and vice versa if both list have the same number of items in them. This last test is basicly the old test as a fall back if the new one did not detect a problem. This makes sure that the number of the same diagnostics in the lists match.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
pdelvo: Fixed

pdelvo:
I don't think there is any interest in this PR any more. and the underlying code changed too much. Im closing this.

PR #525, opened 2015-03-14
Remove VB Test code

pdelvo:
Because we don't have any VB analyzers this is dead code


sharwell:
:+1: I'll merge this after the build completes.

PR #527, opened 2015-03-15
Fix SA1212 can throw NRE if a expression property is used

pdelvo:


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1212PropertyAccessorsMustFollowOrder.cs
```diff
@@ -84,17 +84,20 @@ private void HandlePropertyDeclaration(SyntaxNodeAnalysisContext context)
 
         private static void AnalyzeProperty(SyntaxNodeAnalysisContext context, BasePropertyDeclarationSyntax propertyDeclaration)
         {
-            var accessors = propertyDeclaration.AccessorList.Accessors;
-            if (propertyDeclaration.AccessorList.IsMissing ||
-                accessors.Count != 2)
+            if (propertyDeclaration?.AccessorList != null)
```
sharwell: :bulb: If you change this to the following, the diff for this method is simplified because it will not affect the indentation of the existing code.

```csharp
if (propertyDeclaration?.AccessorList == null)
    return;
```

pdelvo:
I simplified the diff

sharwell:
Thank you :+1:

PR #528, opened 2015-03-15
Fix that SA1114 can throw a NRE if a object creation has no argument list

pdelvo:


macpak:
Yeah, must have missed that. Good catch @pdelvo . :+1:

PR #532, opened 2015-03-15
Disable diagnostic IDE0003 (Remove 'this' qualification)

### LINKED ISSUE #530: Discussion: Disable SA1101 or IDE0003 for the project.
There is a clash between SA1101 and IDE0003, where either one is suggesting to 'fix the code', thereby triggering the other one. I think we should disable one of them, in order to no longer have the highlighting and offered code fixes all the time.

My personal preference would be to disable SA1101, as I think it offers no benefit for anything but fields. 

I created this issue to hear your opinions on it.

sharwell:
This rule conflicts with SA1101, where the StyleCop rule is preferred for this project.

Fixes #530


vweijsters:
This works for me.

PR #543, opened 2015-03-16
Fix that SA1410 and SA1411 CF removes trivia

### LINKED ISSUE #541: Make sure SA1410 and SA1411 preserve comments inside parentheses
The code fixes for SA1410 and SA1411 remove nodes of the form `()` from a document. The unit tests for these code fixes should be updated to ensure that when a node of the form `(/*comment*/)` is removed by this code fix, the `/*comment*/` is not removed as well.

pdelvo:
Fixes #541.


vweijsters:
:+1: Did a quick review and it looks good to me.

sharwell:
Thanks for helping with the review @vweijsters :+1:

PR #550, opened 2015-03-20
Async cleanup

sharwell:
Considering the number of lines changed, I encourage you to look at the individual commits.

* Use `ConfigureAwait(false)` with all `await` operations (automated detection and refactoring per Code Cracker CC0070).
* Pass a `CancellationToken` to asynchronous methods when available and supported.
* Rename asynchronous methods with the `Async` suffix (automated detection and refactoring per Code Cracker CC0061).

PR #551, opened 2015-03-21
Implemented SA1507 (incl. codefix + tests)

### LINKED ISSUE #108: SA1507: CodeMustNotContainMultipleBlankLinesInARow
http://www.stylecop.com/docs/SA1507.html

vweijsters:
Fixes #108

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
@@ -123,9 +123,9 @@ public class Foo
                 this.CSharpDiagnostic().WithLocation(6, 1),
                 this.CSharpDiagnostic().WithLocation(12, 1),
                 this.CSharpDiagnostic().WithLocation(15, 1),
-                /* line 19 should not report a diagnostic, as its part of the directive is inactive */
+                /* line 19 should not report a diagnostic, as it's part of the directive is inactive */
                 this.CSharpDiagnostic().WithLocation(23, 1)
-                /* line 26 should not report a diagnostic, as its part of a comment */
+                /* line 26 should not report a diagnostic, as it's part of a comment */
             };
 
             await this.VerifyCSharpDiagnosticAsync(TestCode, expectedDiagnostics, CancellationToken.None);
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
@@ -13,7 +13,37 @@
     /// </summary>
     public class SA1507UnitTests : CodeFixVerifier
     {
-        private const string DiagnosticId = SA1507CodeMustNotContainMultipleBlankLinesInARow.DiagnosticId;
+        private const string TestCode = @"public class Foo
+{
+
+
+    using System;
+
+
+    using System.Collections;
+
+    void Bar()
+    {
+
+
+#if !IGNORE
+
+
+        return;
+#else
+
+
+        return;
+#endif
+
+
+/*
+
+
+*/
+    }
+}
+";
 
         /// <summary>
         /// Verifies that the analyzer will properly handle an empty source.
@@ -61,42 +91,9 @@ public class Foo
         /// <summary>
         /// Validate that all invalid multiple blank lines are reported.
         /// </summary>
-        /// <returns></returns>
         [Fact]
         public async Task TestInvalidMultipleBlankLines()
         {
-            var testCode = @"public class Foo
-{
-
-
-    using System;
-
-
-    using System.Collections;
-
-    void Bar()
-    {
-
-
-#if !IGNORE
-
-
-        return;
-#else
-
-
-        return;
-#endif
-
-
-/*
-
-
-*/
-    }
-}
-";
-
             var expectedDiagnostics = new[]
             {
                 this.CSharpDiagnostic().WithLocation(3, 1),
@@ -108,13 +105,55 @@ public class Foo
                 /* line 26 should not report a diagnostic, as its part of a comment */
             };
 
-            await this.VerifyCSharpDiagnosticAsync(testCode, expectedDiagnostics, CancellationToken.None);
+            await this.VerifyCSharpDiagnosticAsync(TestCode, expectedDiagnostics, CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Validate that the code fixes for all invalid multiple blank lines works properly.
+        /// </summary>
+        [Fact]
+        public async Task TestInvalidMultipleBlankLinesCodeFix()
+        {
+            var fixedTestCode = @"public class Foo
+{
+
+    using System;
+
+    using System.Collections;
+
+    void Bar()
+    {
+
+#if !IGNORE
+
+        return;
+#else
+
+
+        return;
+#endif
+
+/*
+
+
+*/
+    }
+}
+";
+
+            await this.VerifyCSharpFixAsync(TestCode, fixedTestCode);
         }
 
         /// <inheritdoc/>
         protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
         {
             return new SA1507CodeMustNotContainMultipleBlankLinesInARow();
+        }
+
+        /// <inheritdoc/>
+        protected override CodeFixProvider GetCSharpCodeFixProvider()
+        {
+            return new SA1507CodeFixProvider();
         }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
@@ -1,0 +1,120 @@
+﻿namespace StyleCop.Analyzers.Test.LayoutRules
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.LayoutRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1507CodeMustNotContainMultipleBlankLinesInARow"/>.
+    /// </summary>
+    public class SA1507UnitTests : CodeFixVerifier
+    {
+        private const string DiagnosticId = SA1507CodeMustNotContainMultipleBlankLinesInARow.DiagnosticId;
+
+        /// <summary>
+        /// Verifies that the analyzer will properly handle an empty source.
+        /// </summary>
+        [Fact]
+        public async Task TestEmptySource()
+        {
+            var testCode = string.Empty;
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Verifies that empty lines at the start of the file do not trigger any diagnostics.
+        /// (This will be handled by SA1517)
+        /// </summary>
+        [Fact]
+        public async Task TestEmptyLinesAtStartOfFile()
+        {
+            var testCode = @"
+
+public class Foo
+{
+}";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Verifies that empty lines at the end of the file do not trigger any diagnostics.
+        /// (This will be handled by SA1518)
+        /// </summary>
+        [Fact]
+        public async Task TestEmptyLinesAtEndOfFile()
+        {
+            var testCode = @"public class Foo
+{
+}
+
+
+";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Validate that all invalid multiple blank lines are reported.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeMustNotContainMultipleBlankLinesInARow.cs
```diff
@@ -54,16 +57,54 @@ public class SA1507CodeMustNotContainMultipleBlankLinesInARow : DiagnosticAnalyz
         /// <inheritdoc/>
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
         {
-            get
-            {
-                return SupportedDiagnosticsValue;
-            }
+            get { return SupportedDiagnosticsValue; }
```
pdelvo: Would this trigger [SA1502](http://www.stylecop.com/docs/SA1502.html)?

vweijsters: I don't think so and it should not either

See excerpt from SA1502 documentation below:
> As an exception to this rule, accessors within properties, events, or indexers are allowed to be written all on a single line, as long as the accessor is short.

vweijsters:
Working on a fix where SA1507 triggers within verbatim string literals

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeFixProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            foreach (Diagnostic diagnostic in context.Diagnostics.Where(d => FixableDiagnostics.Contains(d.Id)))
+            {
+                context.RegisterCodeFix(CodeAction.Create("Remove multiple blank lines", token => GetTransformedDocument(context.Document, diagnostic, token)), diagnostic);
+            }
+
+            return Task.FromResult(true);
+        }
+
+        private static async Task<Document> GetTransformedDocument(Document document, Diagnostic diagnostic, CancellationToken token)
```
pdelvo: This should be renamed to GetTransformedDocumentAsync. All the other ones dont have the suffix right now, but #550 will change that.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeFixProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+                    endIndex = i - 1;
+                    break;
+                }
+            }
+
+            if (endIndex >= (startIndex + 1))
+            {
+                var replaceSpan = TextSpan.FromBounds(sourceText.Lines[startIndex + 1].SpanIncludingLineBreak.Start, sourceText.Lines[endIndex].SpanIncludingLineBreak.End);
+                var newSourceText = sourceText.Replace(replaceSpan, string.Empty);
+                return document.WithText(newSourceText);
```
pdelvo: Im afraid that this will make roslyn reparse the whole file.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeFixProvider.cs
vweijsters: Lol, I've been fixing that in my other pull requests and now I forget it in a new one :cry:

vweijsters:
Fixed the issue with verbatim string literals.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeMustNotContainMultipleBlankLinesInARow.cs
pdelvo: Ah okay. Didnt see that

pdelvo:
Otherwise this looks good.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeFixProvider.cs
vweijsters: Fixed

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeFixProvider.cs
vweijsters: Is there anyway to determine when a reparse will (is) happening?

pdelvo: I looked into the source and it looks like is is only reparsing the changed parts. So this is okay

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1507CodeMustNotContainMultipleBlankLinesInARow.cs
sharwell: @vweijsters @pdelvo I generally disagree with the exception, but not enough to require everything appear in the long format. However, I would prefer the pull request not change the layout of an existing block already in the preferred form.

vweijsters: Fair point, I've reverted the change.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            var expectedDiagnostics = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(3, 1),
+                this.CSharpDiagnostic().WithLocation(6, 1),
+                this.CSharpDiagnostic().WithLocation(12, 1),
+                this.CSharpDiagnostic().WithLocation(15, 1),
+                /* line 19 should not report a diagnostic, as its part of the directive is inactive */
+                this.CSharpDiagnostic().WithLocation(23, 1)
+                /* line 26 should not report a diagnostic, as its part of a comment */
```
sharwell: :bulb: it's

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1507UnitTests.cs
```diff
@@ -0,0 +1,182 @@
...
+        [Fact]
+        public async Task TestInvalidMultipleBlankLines()
+        {
+            var expectedDiagnostics = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(3, 1),
+                this.CSharpDiagnostic().WithLocation(6, 1),
+                this.CSharpDiagnostic().WithLocation(12, 1),
+                this.CSharpDiagnostic().WithLocation(15, 1),
+                /* line 19 should not report a diagnostic, as it's part of the directive is inactive */
```
sharwell: :bulb: This one was correct before (`its`). :laughing: 

PS: The difference in usage was the reason why I couldn't resist commenting

sharwell:
:question: @vweijsters I want you to consider the following alternative implementation, decide which approach we should favor, and then explain your position. :smile: Please note that the tests do not pass out-of-the-box with this due to dotnet/roslyn#1655.

```csharp
{
    context.RegisterSyntaxTreeAction(this.HandleSyntaxTreeAnalysis);
}

private void HandleSyntaxTreeAnalysis(SyntaxTreeAnalysisContext context)
{
    SyntaxNode root = context.Tree.GetRoot(context.CancellationToken);
    foreach (var token in root.DescendantTokens(descendIntoTrivia: false))
    {
        if (token.IsKind(SyntaxKind.EndOfFileToken))
        {
            // If the file ends with blanks lines, ignore them, they will be handled by SA1518.
            continue;
        }

        int blankLineIndex = 0;
        int blankLineEndIndex = -1;
        int blankLineCount = 0;
        SyntaxTriviaList leadingTrivia = token.LeadingTrivia;
        for (int i = 0; i < leadingTrivia.Count; i++)
        {
            switch (leadingTrivia[i].Kind())
            {
            case SyntaxKind.WhitespaceTrivia:
                continue;

            case SyntaxKind.EndOfLineTrivia:
                blankLineEndIndex = i;
                blankLineCount++;
                if (blankLineIndex < 0)
                {
                    blankLineIndex = i + 1;
                    blankLineCount = 0;
                }

                continue;

            default:
                ReportDiagnosticIfNecessary(context, leadingTrivia, blankLineIndex, blankLineEndIndex, blankLineCount);
                blankLineIndex = -1;
                continue;
            }
        }

        ReportDiagnosticIfNecessary(context, leadingTrivia, blankLineIndex, blankLineEndIndex, blankLineCount);
    }
}

private static void ReportDiagnosticIfNecessary(SyntaxTreeAnalysisContext context, SyntaxTriviaList leadingTrivia, int blankLineIndex, int blankLineEndIndex, int blankLineCount)
{
    if (blankLineIndex < 0 || blankLineEndIndex <= blankLineIndex)
    {
        // nothing to report
        return;
    }

    if (blankLineCount < 2)
    {
        // only care about multiple blank lines in a row
        return;
    }

    if (leadingTrivia[blankLineIndex].SpanStart == 0)
    {
        // blank lines at the beginning are reported by SA1517
        return;
    }

    TextSpan span = TextSpan.FromBounds(leadingTrivia[blankLineIndex].SpanStart, leadingTrivia[blankLineEndIndex].Span.End);
    Location location = Location.Create(context.Tree, span);
    context.ReportDiagnostic(Diagnostic.Create(Descriptor, location));
}
```

vweijsters:
The issue reported in the Roslyn database is not a bug. Have another close look at the test code (I needed several) and notice that the test code is trying to use using statements inside a class declaration, which is of course invalid C#. I'll fix the test code straight away.

Still trying decide which implementation I like more...

vweijsters:
I'm still undecided. 

I very much like the way the alternative implementation elegantly handles the verbatim string literals, the multi-line comments and the disabled text trivia. 

I don't like the fact that it can't handle empty lines in the trailing trivia (or worse a mix of leading and trailing trivia that cause empty lines). Not sure if the Roslyn parser will ever generate them, if not then this is a moot point.

@sharwell What made you propose an alternative implementation?

vweijsters:
This is a fixed alternative implementation that passes the revised tests
The original alternative implementation had issues with directive trivia, which imply a new line at the end of that trivia.

```C#
        {
            context.RegisterSyntaxTreeAction(this.HandleSyntaxTreeAnalysis);
        }

        private void HandleSyntaxTreeAnalysis(SyntaxTreeAnalysisContext context)
        {
            SyntaxNode root = context.Tree.GetRoot(context.CancellationToken);
            foreach (var token in root.DescendantTokens(descendIntoTrivia: false))
            {
                if (token.IsKind(SyntaxKind.EndOfFileToken))
                {
                    // If the file ends with blanks lines, ignore them, they will be handled by SA1518.
                    continue;
                }

                int blankLineIndex = 0;
                int blankLineEndIndex = -1;
                int blankLineCount = 0;
                SyntaxTriviaList leadingTrivia = token.LeadingTrivia;
                for (int i = 0; i < leadingTrivia.Count; i++)
                {
                    switch (leadingTrivia[i].Kind())
                    {
                        case SyntaxKind.WhitespaceTrivia:
                            break;

                        case SyntaxKind.EndOfLineTrivia:
                            blankLineEndIndex = i;
                            blankLineCount++;
                            break;

                        default:
                            ReportDiagnosticIfNecessary(context, leadingTrivia, blankLineIndex, blankLineEndIndex, blankLineCount);
                            blankLineIndex = i + 1;
                            blankLineCount = 0;
                            break;
                    }
                }

                ReportDiagnosticIfNecessary(context, leadingTrivia, blankLineIndex, blankLineEndIndex, blankLineCount);
            }
        }

        private static void ReportDiagnosticIfNecessary(SyntaxTreeAnalysisContext context, SyntaxTriviaList leadingTrivia, int blankLineIndex, int blankLineEndIndex, int blankLineCount)
        {
            if (blankLineIndex < 0 || blankLineEndIndex <= blankLineIndex)
            {
                // nothing to report
                return;
            }

            if (blankLineCount < 2)
            {
                // only care about multiple blank lines in a row
                return;
            }

            if (leadingTrivia[blankLineIndex].SpanStart == 0)
            {
                // blank lines at the beginning are reported by SA1517
                return;
            }

            TextSpan span = TextSpan.FromBounds(leadingTrivia[blankLineIndex].SpanStart, leadingTrivia[blankLineEndIndex].Span.End);
            Location location = Location.Create(context.Tree, span);
            context.ReportDiagnostic(Diagnostic.Create(Descriptor, location));
        }
```

sharwell:
> What made you propose an alternative implementation?

Among other things, the call to `GetText` in the original implementation concerns me. Whenever I have concerns about performance or simplicity, sometimes it helps to see two implementations side-by-side and discuss them. If I put in that work and it gets thrown out in the end, I'll know it's because the one we picked is *at least* as good. :smile: 

> I don't like the fact that it can't handle empty lines in the trailing trivia (or worse a mix of leading and trailing trivia that cause empty lines).

From everything I've seen so far, trailing trivia for a token includes but never extends beyond the `EndOfLineTrivia` containing the token. This means a `SyntaxTrivia` cannot both exist on a blank line and be containing in the trailing trivia for a token.

vweijsters:
Your effort to make an alternative implementation is very much appreciated and in this case it won't be thrown out :smile: I was just curious what triggered you to make it.

I've made up my mind and given the current behavior of Roslyn with respect to trailing trivia (and the low probability that that behavior will change before release) I think we should go with the alternative implementation, as it is much more elegant at handling the special cases.

@sharwell If you are ok with it, I'll change the pull request to use the fixed alternative implementation.

sharwell:
@vweijsters that's fine with me.

vweijsters:
Switch to alternative implementation has been committed.

Issue #548, opened 2015-03-20
SA1516 (ElementsMustBeSeparatedByBlankLine) reported for adjacent fields; code fix does not work


sharwell:
When a field follows a multi-line field, SA1516 gets reported. In this case, the code fix is proposed but the set of changes is empty. The following code demonstrates this case (SA1516 is reported for `experiment2`).

```csharp
private string experiment =
    string.Empty;
private string experiment2;
```

* If SA1516 should be reported in this case, the code fix should be updated to handle it.
* Otherwise, SA1516 should not be reported here.

PR #553, opened 2015-03-21
Improve layout

vweijsters:
Fixed layout issues in our code base, while testing SA1507 implementation.
(fixed all instances of SA1507, SA1516 and SA1518)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
@@ -10,7 +10,6 @@
     using TestHelper;
     using Xunit;
 
-
     /// <summary>
     /// This class contains unit tests for <see cref="SA1121UseBuiltInTypeAlias"/>
     /// </summary>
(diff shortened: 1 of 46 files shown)


macpak:
Changes look ok.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
```diff
@@ -21,6 +20,7 @@ public class SA1121UnitTests : CodeFixVerifier
             new Tuple<string, string>("object", nameof(Object)),
             new Tuple<string, string>("string", nameof(String))
         };
+
```
sharwell: :question: Is this supposed to be reported, or is it just a result of #548?

vweijsters: That is a result of SA1516 indeed (#548)

sharwell: I received confirmation in #561 that these needed to be fixed.

PR #557, opened 2015-03-24
Sa1307 is reported on read only fields

pdelvo:
We usually try to reduce the number of diagnostic reported for the same issue. Read only fields are reported by SA1304 so we should not report SA1307 here. We have a test that makes sure that read only fields are not reported, but the test is currently broken because of the roslyn bug that made us disable a bunch of tests. So we have a regression test when we have rc2.
I also changed that the diagnostic was reported on the whole declarator and not on the identifier directly.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1307UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1307UnitTests.cs
@@ -95,8 +95,6 @@ string Bar, Car, Dar;
             await this.TestThatDiagnosticIsReported_SingleFieldImpl("public");
             await this.TestThatDiagnosticIsReported_SingleFieldImpl("internal");
             await this.TestThatDiagnosticIsReported_SingleFieldImpl("protected internal");
-            await this.TestThatDiagnosticIsReported_SingleFieldImpl("public readonly");
-            await this.TestThatDiagnosticIsReported_SingleFieldImpl("internal readonly");
         }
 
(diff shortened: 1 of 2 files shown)


sharwell:
It seems users might care more about StyleCop violations which are visible to other assemblies (e.g. a naming violation of a public member might be "worse" than a naming violation for an internal or private member). If that's the case, SA1307 might be the preferable warning when SA1304 and SA1307 both apply. What do you think?

pdelvo:
Im not sure. SA1304 sounds more specific to me, but I see your point. StyleCop reports both.

vweijsters:
I think that SA1307 has the better message and always reporting the same message will give a better understanding by users most of the time.
It would effectively render SA1304 useless for anything but protected fields. Not sure about protected internal fields though, my gut feeling says that protected internal should be treated the same as internal.

sharwell:
If SA1307 takes precedence over SA1304, then SA1304 would only apply to `internal` fields. `protected` and `protected internal` are both accessible to other assemblies, so SA1307 would apply.

I'll place my "primary" vote for reporting SA1307 instead of SA1304 when both apply, and a fallback vote for reporting both (since that's what the reference implementation does).

vweijsters:
The StyleCop documentation for SA1307 only mentions public and internal. 
I agree with your point that it should also report it for protected and protected internal, but its a change from what StyleCop does.

I'm still a bit hesitant about internal, as InternalsVisibleTo makes internal fields also accessible to other assemblies and that construction gets used quite often for unit tests belonging to a certain assembly.

sharwell:
Closing as wontfix; SA1304 should be suppressed in this case instead. :+1:

PR #559, opened 2015-03-24
Implement SA1201

### LINKED ISSUE #61: SA1201: ElementsMustAppearInTheCorrectOrder
http://www.stylecop.com/docs/SA1201.html

pdelvo:
Fixes #61. i have not yet implemented a code fix because I think we should discuss how it should fix a violation.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1201ElementsMustAppearInTheCorrectOrder.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1201ElementsMustAppearInTheCorrectOrder.cs
@@ -6,6 +6,8 @@
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
     using Helpers;
+    using System.Collections.Generic;
+
 
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


sharwell:
I created issue #573 for the code fix.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1201ElementsMustAppearInTheCorrectOrder.cs
```diff
@@ -103,17 +107,40 @@ public class SA1201ElementsMustAppearInTheCorrectOrder : DiagnosticAnalyzer
...
 
         private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.DisabledNoTests, Description, HelpLink);
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, true, Description, HelpLink);
 
         private static readonly ImmutableArray<DiagnosticDescriptor> SupportedDiagnosticsValue =
             ImmutableArray.Create(Descriptor);
 
+        // extern alias and usings are missing here because the compiler itself is enforcing the right order.
+        private readonly static ImmutableArray<SyntaxKind> OuterOrder = ImmutableArray.Create(
```
sharwell: :bulb: [SA1206](http://www.stylecop.com/docs/SA1206.html) says `private static readonly`.

sharwell:
I don't like how a nested class at the beginning of a class can result in *every other element* getting a warning. What would happen if we are more selective in our reporting?

:memo: Only one of the following items would be used.

1. Report a warning on the current element if it should come after the element that follows it.
2. Report a warning on the current element if it should appear before the element that precedes it.

pdelvo:
I implemented the change you suggested. The diagnostics that get reported are now: A X should not follow a Y.

pdelvo:
I need to work on thisd a bit more to remove the compiler errors in the unit tests. I also just discovered a crash if the document is in an invalid state and has an IncompleteMemberSyntax in it

sharwell:
:question: Is this ready now?

pdelvo:
I think so.

PR #561, opened 2015-03-24
Fix that SA1516 CF doesnt work sometimes

### LINKED ISSUE #548: SA1516 (ElementsMustBeSeparatedByBlankLine) reported for adjacent fields; code fix does not work

pdelvo:
Fixes #548.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1516UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1516UnitTests.cs
@@ -372,6 +372,38 @@ public class Foo
             await this.VerifyCSharpFixAsync(testCode, fixedCode);
         }
 
+        [Fact]
+        public async Task TestThatCodeFixWorksOnFieldsAdjacentToMultiLineFields()
+        {
+            string testCode = @"using System;
+
+public class Foo
+{
+    private string experiment =
+        string.Empty;
+    private string experiment2;
+}";
+            var expected = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(7, 20)
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None);
+
+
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1516UnitTests.cs
```diff
@@ -372,6 +372,38 @@ public class Foo
...
         }
 
+        [Fact]
+        public async Task TestThatCodeFixWorksOnFieldsAdjacentToMultiLineFields()
+        {
+            string testCode = @"using System;
+
+public class Foo
+{
+    private string experiment =
```
sharwell: :question: I'm curious if the reference implementation reports SA1516 for this case. Can you check?

pdelvo: I just checked. It does report for that code. Whats weird is the location the diagnostic is at. It reports for char 1 at that line, so thats a difference. but I think the location we report is better.

vweijsters: I can understand StyleCop reporting char 1 as position, as that is the point where the empty line is missing. I personally prefer that over reporting it at the element name, as it is much more clear for me where I should do something.

sharwell: @vweijsters Please file a separate issue for that if you'd like us to consider a change from the current behavior. It would take more discussion.

Issue #562, opened 2015-03-25
Reporting location for SA1516


vweijsters:
Opened this to continue the discussion started in #561.

We are deviating from the reference implementation with respect to reporting location for SA1516.
My opinion is that the position reported by StyleCop is the more intuitive one, as it reports to location where something is missing. Reporting it at the element location leads me to believe that something is wrong with the element itself, instead of just the layout around it.

sharwell:
I vote for placing it on the first non-whitespace (or newline) token following the location where the new line needs to be inserted.

pdelvo:
![location](https://cloud.githubusercontent.com/assets/1710209/6869417/1c12a866-d494-11e4-9d14-f5c4bc0e33e8.PNG)
I didnt like that the squiggles are so small, but I guess Im outvoted. Im going to create a PR to change that

sharwell:
@pdelvo My vote for that image is placing it on the word "internal".

Even if you go with the whitespace, what about the *entire* leading whitespace instead of just the first character?

pdelvo:
What about this:

If the node has leading trivia, put the diagnostic at the first leading trivia. If not put the diagnostic at the first token.
This would look like this:

![location2](https://cloud.githubusercontent.com/assets/1710209/6869885/87fdc5b6-d498-11e4-9a5c-ba845119fb03.PNG)

sharwell:
:+1: :sparkles:

vweijsters:
That looks very good :+1:

PR #565, opened 2015-03-27
SA1516 should be reported on identifiers

### LINKED ISSUE #558: SA1516 is reported for the declarator, not the identifer
If SA1516 should be reported for this code:
```csharp
private static readonly bool foo = false;
```
Then "foo = false" is the location for the diagnostic. I think reporting on just "foo" would be better.

pdelvo:
Fixes #558.


sharwell:
Is this superseded by #562?

PR #568, opened 2015-03-27
Change the location of SA1516 diagnostics

### LINKED ISSUE #562: Reporting location for SA1516

pdelvo:
Fixes #562.


sharwell:
What's the relation between this and #565?

pdelvo:
#565 was a mistake. I closed it.

PR #569, opened 2015-03-27
Fix an off by one in our test code

### LINKED ISSUE #567: Tests dont verify location if the actual location is at line start
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/c26d390a4f0963f279d06d5079d7448cfe8d92ee/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs#L210
Line start is at character = 0, so we got an off by one error here. The same thing aplies for line = 0

pdelvo:
Fixes #567.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
@@ -195,8 +195,8 @@ namespace TestHelper
 
             var actualLinePosition = actualSpan.StartLinePosition;
 
-            // Only check line position if there is an actual line in the real diagnostic
-            if (actualLinePosition.Line > 0)
+            // Only check line position if it matters
+            if (expected.Line > 0)
             {
                 if (actualLinePosition.Line + 1 != expected.Line)
                 {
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -196,7 +196,7 @@ private static void VerifyDiagnosticLocation(DiagnosticAnalyzer analyzer, Diagno
             var actualLinePosition = actualSpan.StartLinePosition;
 
             // Only check line position if there is an actual line in the real diagnostic
-            if (actualLinePosition.Line > 0)
+            if (actualLinePosition.Line >= 0)
```
sharwell: The bug is actually that `actualLinePosition` is used here. The original intent was for the user to be able to specify in the expected results that the line doesn't matter. The condition should be this:

```csharp
if (expected.Line > 0)
```

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -207,7 +207,7 @@ private static void VerifyDiagnosticLocation(DiagnosticAnalyzer analyzer, Diagno
             }
 
             // Only check column position if there is an actual column position in the real diagnostic
-            if (actualLinePosition.Character > 0)
+            if (actualLinePosition.Character >= 0)
```
sharwell: :exclamation: See my note above.

pdelvo:
I implemented the CR feedback and im creating a roslyn issue now.

PR #572, opened 2015-03-28
Implemented diagnostic for SA1213: EventAccessorsMustFollowOrder

### LINKED ISSUE #73: SA1213: EventAccessorsMustFollowOrder
http://www.stylecop.com/docs/SA1213.html

macpak:
Fixes #73.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1213EventAccessorsMustFollowOrder.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1213EventAccessorsMustFollowOrder.cs
@@ -2,6 +2,8 @@
 {
     using System.Collections.Immutable;
     using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1213EventAccessorsMustFollowOrder.cs
```diff
@@ -51,18 +53,33 @@ public class SA1213EventAccessorsMustFollowOrder : DiagnosticAnalyzer
...
+            if (eventDeclaration.AccessorList.IsMissing ||
+                accessors.Count != 2)
+            {
+                return;
+            }
+
+            if (accessors[0].Kind() == SyntaxKind.RemoveAccessorDeclaration &&
+                accessors[1].Kind() == SyntaxKind.AddAccessorDeclaration)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, accessors[0].GetLocation()));
```
pdelvo: Maybe it would be better to report on the keyword instead of the full accessor.

sharwell: :exclamation: Good call, I agree. :smile:

macpak: Yeah, you're right, that'd be better approach. Thanks !

PR #579, opened 2015-03-29
Simplify SA1605 tests using [Theory]

sharwell:


vweijsters:
Code looks good :+1:

Have you considered using the MemberData attribute, instead of the multiple InlineData ones?

sharwell:
I had not. I'll take a look and see.

sharwell:
@vweijsters I find that it makes it harder to simply look at the test and understand its purpose. It seems the main advantage would be the following situations:

1. You have a very large number of test cases
2. Your test methods have inputs that can't be expressed as a constant
3. You need to generate your test cases at runtime for some reason

vweijsters:
Okay, I'm fine with leaving it the way it is.

PR #580, opened 2015-03-29
Improve reporting granularity of SA1307 tests using [Theory]

sharwell:
This also fixes a bug where `protected const` was tested twice and `protected internal const` was not tested at all.


vweijsters:
Code looks good :+1:

PR #581, opened 2015-03-29
Simplify SA1607 tests using [Theory]

sharwell:


vweijsters:
Code looks good :+1:

I would prefer MemberData instead of InlineData here to reduce duplication.

PR #586, opened 2015-03-29
Improve reporting granularity of SA1306 tests using [Theory]

sharwell:


StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1306UnitTests.cs
```diff
@@ -19,7 +19,18 @@ public class SA1306UnitTests : CodeFixVerifier
...
+        [InlineData("private const")]
+        [InlineData("internal const")]
+        [InlineData("protected const")]
+        [InlineData("protected internal const")]
+        [InlineData("internal readonly")]
+        [InlineData("protected readonly")]
+        [InlineData("protected internal readonly")]
+        [InlineData("public")]
+        [InlineData("internal")]
+        public async Task TestThatDiagnosticIsNotReported(string modifiers)
```
vweijsters: Maybe not related to this pull request:
I expected to see `public const`, `public readonly`, and `protected internal` here as well, or am I missing something?

vweijsters:
Not related to the pull request:
I'm missing test cases for static fields, which should never trigger this diagnostic according to StyleCop documentation.

sharwell:
@vweijsters Can you open a new issue for those test cases?

StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1306UnitTests.cs
sharwell: I would add this to the other issue I suggested creating. Mention that the tests do not cover all combinations of modifiers but should.

PR #604, opened 2015-04-03
Implement "Do not use placeholder elements" as new diagnostic SA1651

### LINKED ISSUE #536: Avoid placeholders in documentation
I would like to propose the implementation and inclusion of a new StyleCop diagnostic.

**Category:** Documentation
**Rule:** Do not use placeholder elements
**Description:** See openstacknetsdk/OpenStackNetAnalyzers#10 (which even includes screenshots)

**Rationale:** Since the implementation of #534 involves inserting a "best guess" regarding the `<value>` text for a property, I would like to wrap the generated text in a `<placeholder></placeholder>` element for the user to review. As shown in the issue linked above, a new diagnostic which reports a warning for these elements encourages users to actively review the text which was inserted for the properties. Without this diagnostic, a code fix which automatically inserts documentation text has a high likelihood of producing incorrect results which are not easily recognized or corrected.

sharwell:
Fixes #536

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651CodeFixProvider.cs
@@ -73,7 +73,7 @@
                 return document;
 
             var leadingTrivia = elementSyntax.StartTag.GetLeadingTrivia();
-            leadingTrivia = leadingTrivia.AddRange(elementSyntax.EndTag.GetTrailingTrivia());
+            leadingTrivia = leadingTrivia.AddRange(elementSyntax.StartTag.GetTrailingTrivia());
             leadingTrivia = leadingTrivia.AddRange(content[0].GetLeadingTrivia());
             content = content.Replace(content[0], content[0].WithLeadingTrivia(leadingTrivia));
 
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651CodeFixProvider.cs
@@ -18,7 +18,7 @@
     /// <para>To fix a violation of this rule, review the placeholder documentation for accuracy and remove the
     /// &lt;placeholder&gt; tags.</para>
     /// </remarks>
-    [ExportCodeFixProvider(nameof(SA1642SA1643CodeFixProvider), LanguageNames.CSharp)]
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(SA1642SA1643CodeFixProvider))]
     [Shared]
     public class SA1651CodeFixProvider : CodeFixProvider
     {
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651DoNotUsePlaceholderElements.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// The ID for diagnostics produced by the <see cref="SA1651DoNotUsePlaceholderElements"/>
+        /// analyzer.
+        /// </summary>
+        public const string DiagnosticId = "SA1651";
+        private const string Title = "Do not use placeholder elements";
+        private const string MessageFormat = "Do not use placeholder elements";
+        private const string Category = "StyleCop.CSharp.DocumentationRules";
+        private const string Description = "The element documentation contains a <placeholder> element.";
+        private const string HelpLink = null;
```
pdelvo: Maybe we should think about putting up some documentation for the help link

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651CodeFixProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        private async Task<Document> GetTransformedDocumentAsync(Document document, XmlElementSyntax elementSyntax, CancellationToken cancellationToken)
+        {
+            SyntaxList<XmlNodeSyntax> content = elementSyntax.Content;
+            if (content.Count == 0)
+                return document;
+
+            var leadingTrivia = elementSyntax.StartTag.GetLeadingTrivia();
+            leadingTrivia = leadingTrivia.AddRange(elementSyntax.EndTag.GetTrailingTrivia());
```
pdelvo: Why EndTag?

pdelvo:
Looks good!

Maybe we can use the github wiki for documentation on new rules for now. Its infortunatly a bit limited and doesnt support PRs. If needed I could host something on my server.

Before you merge this I think we should wait for more opinions on this.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651DoNotUsePlaceholderElements.cs
sharwell: We'll get AppVeyor configured to build a SHFB site and migrate our links to point there.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1651CodeFixProvider.cs
sharwell: Should have been `StartTag`. However, I can't find any case where the trailing trivia of the start or end tag XML documentation node is not empty. If that holds true always this never would have revealed itself.

pdelvo:
The changes since last time I had a look at this look good and I think this one can be merged if there are no non trivial merge conflicts holding it back.

PR #606, opened 2015-04-03
Implement a code fix for SA1609 and SA1610

### LINKED ISSUE #534: Code fix: SA1610 (PropertyDocumentationMustHaveValueText)
Implement a code fix for #130.

Normally it is not possible to generate the text for a documentation element. However, in many cases where the `<summary>` text for a property follows a particular form (e.g. the form prescribed in #143 and #144), the `<value>` text can be filled in using a slightly reworded variant of the summary text. A concrete example of this, although targeted to the documentation requirements of a particular SDK, can be seen in openstacknetsdk/OpenStackNetAnalyzers#6.

sharwell:
Fixes #534

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
@@ -11,6 +11,7 @@
     using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Microsoft.CodeAnalysis.Formatting;
     using StyleCop.Analyzers.Helpers;
 
     /// <summary>
@@ -87,9 +88,11 @@
                 content = XmlSyntaxFactory.List(XmlSyntaxFactory.PlaceholderElement(content));
             }
 
-            XmlElementSyntax valueElement = XmlSyntaxFactory.MultiLineElement(XmlCommentHelper.ValueXmlTag, content);
+            string newLineText = document.Project.Solution.Workspace.Options.GetOption(FormattingOptions.NewLine, LanguageNames.CSharp);
 
-            XmlNodeSyntax leadingNewLine = XmlSyntaxFactory.NewLine();
+            XmlElementSyntax valueElement = XmlSyntaxFactory.MultiLineElement(XmlCommentHelper.ValueXmlTag, newLineText, content);
+
+            XmlNodeSyntax leadingNewLine = XmlSyntaxFactory.NewLine(newLineText);
 
             // HACK: The formatter isn't working when contents are added to an existing documentation comment, so we
             // manually apply the indentation from the last line of the existing comment to each new line of the
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
@@ -1,0 +1,199 @@
+﻿namespace StyleCop.Analyzers.DocumentationRules
+{
+    using System;
+    using System.Collections.Immutable;
+    using System.Composition;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using StyleCop.Analyzers.Helpers;
+
+    /// <summary>
+    /// Implements a code fix for <see cref="SA1609PropertyDocumentationMustHaveValue"/> and
+    /// <see cref="SA1610PropertyDocumentationMustHaveValueText"/>.
+    /// </summary>
+    /// <remarks>
+    /// <para>To fix a violation of this rule, fill-in a description of the value held by the property within the
+    /// &lt;value&gt; tag.</para>
+    /// </remarks>
+    [ExportCodeFixProvider(nameof(SA1609SA1610CodeFixProvider), LanguageNames.CSharp)]
+    [Shared]
+    public class SA1609SA1610CodeFixProvider : CodeFixProvider
+    {
+        private static readonly ImmutableArray<string> FixableDiagnostics =
+            ImmutableArray.Create(SA1609PropertyDocumentationMustHaveValue.DiagnosticId, SA1610PropertyDocumentationMustHaveValueText.DiagnosticId);
+
+        /// <inheritdoc/>
+        public override ImmutableArray<string> FixableDiagnosticIds => FixableDiagnostics;
+
+        /// <inheritdoc/>
+        public override FixAllProvider GetFixAllProvider()
+        {
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+
+        /// <inheritdoc/>
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                if (!FixableDiagnostics.Contains(diagnostic.Id, StringComparer.Ordinal))
+                    continue;
+
+                string description = "Document value from summary";
+                context.RegisterCodeFix(CodeAction.Create(description, cancellationToken => this.GetTransformedDocumentAsync(context.Document, diagnostic, cancellationToken)), diagnostic);
+            }
+
+            return Task.FromResult(true);
+        }
+
+        private async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        {
+            var documentRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
```diff
@@ -0,0 +1,199 @@
...
+                        if (index >= 0)
+                            valueText = valueText.Remove(index, prefix.Length);
+
+                        SyntaxToken replaced = SyntaxFactory.Token(getsToken.LeadingTrivia, getsToken.Kind(), text, valueText, getsToken.TrailingTrivia);
+                        summaryContent = summaryContent.Replace(firstText, firstText.ReplaceToken(getsToken, replaced));
+                        return true;
+                    }
+                }
+            }
+
```
vweijsters: :point_up: The nested if statements above are not inline with the rest of the code, where an return as quickly as possible strategy is used.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
```diff
@@ -0,0 +1,199 @@
...
+                return document;
+
+            SyntaxList<XmlNodeSyntax> summaryContent = summaryElement.Content;
+            if (!this.TryRemoveSummaryPrefix(ref summaryContent, "Gets or sets "))
+            {
+                if (!this.TryRemoveSummaryPrefix(ref summaryContent, "Gets "))
+                {
+                    this.TryRemoveSummaryPrefix(ref summaryContent, "Sets ");
+                }
+            }
```
vweijsters: I don't particularly like the ref parameters in `TryRemoveSummaryPrefix` calls. It is not similar to the Roslyn API, where each transforming method returns the modified node.

I'd prefer to see something like:
`summaryContent = this.TryRemoveSummaryPrefix(summaryContent, "Gets or sets ", "Gets ", "Sets ")`

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
```diff
@@ -0,0 +1,199 @@
...
+                    {
+                        string text = getsToken.Text;
+                        string valueText = getsToken.ValueText;
+                        int index = text.IndexOf(prefix);
+                        if (index >= 0)
+                        {
+                            bool additionalCharacters = index + prefix.Length < text.Length;
+                            text = text.Substring(0, index)
+                                + (additionalCharacters ? char.ToUpperInvariant(text[index + prefix.Length]).ToString() : string.Empty)
+                                + text.Substring(index + (additionalCharacters ? (prefix.Length + 1) : prefix.Length));
```
vweijsters: This will throw an `ArgumentOutOfRangeException` if there is only a single character after the prefix. For example: `Gets x`

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DocumentationCommentExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+                            firstTextToken.ValueText.TrimStart(),
+                            firstTextToken.TrailingTrivia);
+
+                        summaryContent = summaryContent.Replace(firstTextSyntax, firstTextSyntax.ReplaceToken(firstTextToken, newFirstToken));
+                    }
+                }
+            }
+
+            return summaryContent;
+        }
```
vweijsters: :point_up: The method name does not cover the functionality of the method. It potentially removes more than just the first and last empty line.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlSyntaxFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+            return TextNewLine(true);
+        }
+
+        public static SyntaxToken TextNewLine(bool continueComment)
+        {
+            SyntaxToken token = SyntaxFactory.XmlTextNewLine(
+                SyntaxFactory.TriviaList(),
+                Environment.NewLine,
+                Environment.NewLine,
+                SyntaxFactory.TriviaList());
```
vweijsters: The code above should use `FormattingOptions.NewLine` instead of `Environment.NewLine`, to keep in line with the new line style used in a project / solution.

vweijsters:
Besides the remarks above, the code looks good :+1: 

@sharwell When there is no properly formatted `<summary>` tag, I think it would be better to generate a `<value>` tag with a placeholder that contains the name of the property. What are your thoughts on that?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DocumentationCommentExtensions.cs
sharwell: Hm. Perhaps it should trim all leading and trailing whitespace?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
sharwell: I don't see it?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
sharwell: How would you know whether or not to try the next prefix?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlSyntaxFactory.cs
sharwell: It should actually use the line ending of the current document, even if that differs from both of the above. Unfortunately I don't know how to get that.

sharwell: For better or worse, according to dotnet/roslyn#202 the behavior implemented here matches Roslyn's current formatting behavior.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
vweijsters: You can't from the return value for sure. That's why I added all the possible prefixes as parameters to the call and they should then match in left to right order I guess.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
vweijsters: My bad, I mistakenly assumed that string.Substring() will throw an exception when the start index is equal to the length, which isn't the case. It only throws that exception when the start index is larger than the length.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlSyntaxFactory.cs
vweijsters: That's not possible afaik and might even present a problem if someone is mixing different types of line endings in one document.

vweijsters: The issue is has a milestone assigned to it, so I guess they will be fixing it before release.
If they do fix it, I think the `FormattingOptions.NewLine` is the way to go.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
sharwell: I prefer it the way its written, but that just means we should get more eyes on it. @pdelvo and @macpak what do you think?

sharwell:
> When there is no properly formatted `<summary>` tag, I think it would be better to generate a `<value>` tag with a placeholder that contains the name of the property. What are your thoughts on that?

Can you give an example of what it currently does, and what you're thinking could be preferable?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DocumentationCommentExtensions.cs
sharwell: @vweijsters This method was tough to write considering the way trivia works within documentation comments. When I used this in practice, it did a remarkably good job of inserting exactly the text I wanted to see. If you aren't opposed to it, I think I'd like to file a new issue to revisit the behavior in the future.

That said, I have no objection to renaming it, although I'm not sure exactly what to call it.

vweijsters: @sharwell I can't really think of a better name as well. 
As far as I'm concerned there is no need to revisit the behavior. I would prefer documentation for the method that explains the additional behavior, so that its clear for other developers reading the code that the method does more than the name implies.

vweijsters:
Never mind about my suggestion above. I was typing an example to show what I meant and when I had it done, it looked worse than the current solution.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
macpak: I'd keep @sharwell 's approach, mostly because for me the code is more readable.

PR #609, opened 2015-04-04
Expand XmlSyntaxFactory functionality

sharwell:
Blocked on (and includes the work of) #606.

* Expand the `XmlSyntaxFactory` class introduced by #606.
* Simplify certain uses of `XmlCommentHelper` using `XmlSyntaxFactory` methods.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DocumentationCommentExtensions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DocumentationCommentExtensions.cs
@@ -6,6 +6,7 @@
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using StyleCop.Analyzers.SpacingRules;
 
     internal static class DocumentationSyntaxExtensions
     {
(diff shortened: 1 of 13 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DocumentationCommentExtensions.cs
```diff
@@ -212,5 +213,35 @@ public static bool IsXmlWhitespace(this SyntaxToken node)
...
+                where token.IsKind(SyntaxKind.XmlTextLiteralNewLineToken)
+                where token.HasTrailingTrivia
+                let next = token.GetNextToken(includeZeroWidth: true, includeSkipped: true, includeDirectives: true, includeDocumentationComments: true)
+                where !next.IsMissingOrDefault()
+                select new KeyValuePair<SyntaxToken, SyntaxToken>(token, next);
+
+            Dictionary<SyntaxToken, SyntaxToken> replacements = new Dictionary<SyntaxToken, SyntaxToken>();
+            foreach (var pair in tokensForAdjustment)
+            {
+                replacements[pair.Key] = pair.Key.WithTrailingTrivia();
```
vweijsters: It seems to me that this should be `WithoutTrailingTrivia()`, otherwise the trailing trivia will be duplicated.

sharwell: There is no `WithoutTrailingTrivia()` method for `SyntaxToken`. This is using the `WithTrailingTrivia(params SyntaxTrivia[])` overload of `WithTrailingTrivia`, which effectively removes the trailing trivia.

vweijsters: I definitely got fooled by that :unamused: 
In that case its working as intended.

vweijsters:
:+1: Looks good

PR #623, opened 2015-04-04
Only report SA1307 on a variable name

### LINKED ISSUE #618: SA1307 (AccessibleFieldsMustBeginWithUpperCaseLetter) reported for entire initializer
SA1307 should be reported only on the identifier for a variable. It is currently reported on the entire initializer expression.

sharwell:
Fixes #618


pdelvo:
Looks good. I wonder if it would be valuable to test that. Our testing framework doesnt support this right now.

sharwell:
Added issue #625 to track the testing issue.

PR #624, opened 2015-04-04
Fix 620

### LINKED ISSUE #620: SA1121 (UseBuiltInTypeAlias) code fix does not work inside documentation comments
While SA1121 is reported for the following case, the code fix does not provide an automatic correction:

```csharp
/// <see cref="System.String"/>
```

sharwell:
* Fix reporting location of SA1121 within `nameof` expressions

  * Was: `nameof(System.[String].ToString)`
  * Now: `nameof([System.String].ToString)`

* Fix reporting location of SA1121 within `cref` attributes of documentation comments

  * Was: `<see cref="System.[String]"/>`
  * Now: `<see cref="[System.String]"/>`

* Fix #620


pdelvo:
Looks good to me!

PR #644, opened 2015-04-06
Add a mailmap entry for Vincent

sharwell:
This change causes the history to show Vincent Weijsters instead of vweijsters. :+1:


vweijsters:
Go ahead and merge it :+1:

PR #646, opened 2015-04-07
Improved fix 449

### LINKED ISSUE #449: SA1119 code fix needs to add whitespaces

sharwell:
Fixes #449 by using an elastic marker *only* when the trivia list starts out empty.

Supersedes #453.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119CodeFixProvider.cs
@@ -7,9 +7,9 @@
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CodeActions;
     using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Formatting;
-    using StyleCop.Analyzers.SpacingRules;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119CodeFixProvider.cs
```diff
@@ -62,13 +62,12 @@ public override FixAllProvider GetFixAllProvider()
 
         private static Task<Document> GetTransformedDocument(Document document, SyntaxNode root, ParenthesizedExpressionSyntax syntax)
         {
-            var leadingTrivia = syntax.OpenParenToken.GetAllTrivia().Concat(syntax.Expression.GetLeadingTrivia());
-            var trailingTrivia = syntax.Expression.GetTrailingTrivia().Concat(syntax.CloseParenToken.GetAllTrivia());
+            var leadingTrivia = SyntaxFactory.TriviaList(syntax.OpenParenToken.GetAllTrivia().Concat(syntax.Expression.GetLeadingTrivia()));
```
pdelvo: Concat can be replaced with AddRange here as well

pdelvo:
Looks good to me!

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1119CodeFixProvider.cs
sharwell: No, `GetAllTrivia()` returns an `IEnumerable<SyntaxTrivia>`.

Issue #576, opened 2015-03-28
Question: Localization


pdelvo:
Roslyn supports having localized messages in analyzers. Should we localize our messages?

sharwell:
Yes. But before we translate them we need to revisit their structure for consistency and overall usability.

PR #665, opened 2015-04-15
Implemented SA1003 unit tests

### LINKED ISSUE #273: Code fix unit tests: SymbolsMustBeSpacedCorrectly
Implement unit tests for #236.

### LINKED ISSUE #281: Diagnostic unit tests: SymbolsMustBeSpacedCorrectly
Implement unit tests for #9.

vweijsters:
Fixes #281 
Fixes #273

:exclamation: There are major issues with either the tests or the verifier and the code fix provider. Lets have a discussion on what should be fixed and who will do that.


StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
```diff
@@ -0,0 +1,469 @@
...
+        var v11 = v6 ? 1 :0;
+        var v12 = v6 ? 1:0;
+
+        var v13 = v6? 1: 0;
+        var v14 = v6 ?1 :0;
+        var v15 = v6?1:0;
+
+        // invalid lambda expressions
+        Action a1 = ()=> { };
+        Action a2 = () =>{ };
```
sharwell: {F704, layout=inline, width=20} Should we report SA1003 or SA1012 here?

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
```diff
@@ -0,0 +1,469 @@
...
+        var v10 = v6 ? 1: 0;
+        var v11 = v6 ? 1 :0;
+        var v12 = v6 ? 1:0;
+
+        var v13 = v6? 1: 0;
+        var v14 = v6 ?1 :0;
+        var v15 = v6?1:0;
+
+        // invalid lambda expressions
+        Action a1 = ()=> { };
```
sharwell: :question: Should we report SA1003 or SA1009 here?

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
```diff
@@ -0,0 +1,469 @@
...
+        Action a2 = () =>{ };
+        Action a3 = ()=>{ };
```
sharwell: :question: Should we report SA1003 or both SA1009 and SA1012 here?

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
```diff
@@ -0,0 +1,469 @@
...
+        public async Task TestInvalidBinaryExpressions()
+        {
+            var testCode = @"using System;
+
+public class Foo
+{
+    public void Bar()
+    {
+        // invalid assignment
+        var v1= 1;
```
sharwell: I missed this case in the analysis because it's not in an expression.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
```diff
@@ -0,0 +1,469 @@
...
+        var v2 =2;
+        var v3=3;
+
+        // invalid operators
+        var v4 = v1+ v2;
+        var v5 = v1 |v2;
+        var v6 = v1>v2;
+
+        // invalid conditionals
+        var v7 = v6? 1 : 0;
```
sharwell: :bulb: These are technically ternary operators. It would make sense to place them in a separate test.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
vweijsters: Good question, StyleCop reports both, but that doesn't seem right somehow.

I have a very minor preference for SA1003.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
vweijsters: StyleCop reports both here as well, but that still doesn't seem right.

I have a very minor preference for SA1003 again.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
vweijsters: StyleCop reports 4 diagnostics on this line, definitely not right.

A single SA1003 would be preferable for readability.

vweijsters:
In general if a single operator could generate two diagnostics (one on each side), we should report a single one.

vweijsters:
Another point to add to the discussion:
I would prefer if we use the (more general) Stylecop message, instead of the current messages. I think it would also work better with the only one diagnostic per operator rule, if we decide to go for that.

>The spacing around the symbol '=>' is invalid.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1003UnitTests.cs
vweijsters: Good suggestion, I'll split the test cases.

I assume you mean:
```C#
var v7 = v6 ? (1): 0;
```
as invalid?

sharwell:
New messages will be *required* for #576, as described in #597.

vweijsters:
Working on a fixed implementation. Need #935 to make the tests work properly.

vweijsters:
This should be all done now :smile:

Issue #627, opened 2015-04-05
Code Fixes For Naming Rules SA1308 and SA1309 Do Not Always Fix The Name Entirely


johnkoerner:
If a violation contains multiple prefixes, then only the first prefix is fixed, which would result in the user having to run the fix multiple times to fix the violation.  

Example:

    private static string __x = string.Empty;

This code violates SA1309 and the current fix results in a single underscore being removed from the variable name:

    private static string _x = string.Empty;

This new code still violates SA1309 and applying another fix will remove the final underscore and the final result will be no violations.

The same type of issue occurs with SA1308 if the user prefixes a variable with `m_m_` or `s_s_` (or any permutations.)

I think for these cases we should attempt to get the variable name into a known good state such that there are no violations for that particular rule after the code fix is applied.

johnkoerner:
Note, when trying to test the running the fix multiple times, be aware that depending on what version of roslyn you are running, you may encounter an error if you do not close the rename dialog.

This issue is tracked on the Roslyn repo: https://github.com/dotnet/roslyn/issues/554

sharwell:
Did you want to work on this?

johnkoerner:
Yes, I will take this.

PR #666, opened 2015-04-15
Implemented SA1027 unit tests + code fix

### LINKED ISSUE #235: Diagnostic unit tests: TabsMustNotBeUsed
Implement unit tests for #33.

### LINKED ISSUE #243: Code fix: TabsMustNotBeUsed
Implement code fix for #33.

vweijsters:
Fixes #235
Fixed #243

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UnitTests.cs
@@ -40,6 +40,24 @@
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
         }
 
+        /// <summary>
+        /// Verifies that tabs used inside disabled code is not producing diagnostics.
+        /// </summary>
+        [Fact]
+        public async Task TestDisabledCode()
+        {
+            var testCode =
+                "public class Foo\r\n" +
+                "{\r\n" +
+                "#if false\r\n" +
+                "\tpublic const string ValidTestString = \"Text\";\r\n" +
+                "\tpublic const char ValidTestChar = 'c';\r\n" +
+                "#endif\r\n" +
+                "}\r\n";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None);
+        }
+
         [Fact]
         public async Task TestInvalidTabs()
         {
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UnitTests.cs
@@ -64,7 +64,7 @@
             var testCode =
                 "using\tSystem.Diagnostics;\r\n" +
                 "\r\n" +
-                "public class Foo\r\n" +
+                "public\tclass\tFoo\r\n" +
                 "{\r\n" +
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027CodeFixProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        {
+            var indentationOptions = IndentationOptions.FromDocument(document);
+            var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+
+            var violatingTrivia = syntaxRoot.FindTrivia(diagnostic.Location.SourceSpan.Start);
+
+            var tabReplacement = new string(' ', indentationOptions.TabSize);
```
sharwell: :exclamation: This only works if the hard tab appears at the beginning of the line, but I don't think that's the only thing this diagnostic looks for.

vweijsters: Good spot, I never considered that. I'll work on a fix.

sharwell:
:bulb: Add a test for a block of text inside `#if false` (it should be ignored).

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UnitTests.cs
```diff
@@ -0,0 +1,99 @@
...
+                "public class Foo\r\n" +
+                "{\r\n" +
+                "\tpublic void Bar()\r\n" +
+                "\t{\r\n" +
+                "\t  \t// Comment\r\n" +
+                "\t \tDebug.Indent();\r\n" +
+                "   \t}\r\n" +
+                "}\r\n";
+
+            // There is only a single space between the using and the System.Diagnostics,
```
sharwell: :exclamation: This is not correct. The code fix should result in 3 spaces between `using` and `System`.

vweijsters: I don't agree. Inserting three spaces there will indeed keep the original layout, but it will immediately trigger SA1025, something that is not desirable in my opinion.

vweijsters:
Added the `#if false` test case.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UnitTests.cs
sharwell: In general we don't make cross-rule special cases for the code fixes. The only cases where we fix multiple violations at once is renaming a variable that contains the same violation more than once (e.g. #627).

sharwell: For example, what if this code is contained in the following?

```csharp
#pragma warning disable SA1025
[something with a hard tab]
#pragma warning restore SA1025
```

vweijsters: Not making cross-rule specials will definitely keep the code more simple in the long run, which is good.
I was looking at that a bit too much from how I would use it, which may of course not be how other people see it. I'll change the code.

vweijsters:
The code fix provider will now always add the amount of spaces required to keep the formatting the same.

PR #667, opened 2015-04-18
SA1617 Code Fix

### LINKED ISSUE #598: Code fix: SA1617 (VoidReturnValueMustNotBeDocumented)
Implement a code fix for #137.

johnkoerner:
Fixes #598 .  Implemented a codefix and associated unit tests for SA1617.


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1617CodeFixProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+            var document = context.Document;
+            var root = await document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                if (!diagnostic.Id.Equals(SA1617VoidReturnValueMustNotBeDocumented.DiagnosticId))
+                {
+                    continue;
+                }
+
```
vweijsters: I would prefer
```C#
foreach (Diagnostic diagnostic in context.Diagnostics.Where(d => FixableDiagnostics.Contains(d.Id)))
```
instead of the lines above. Its more generic and less clutter IMHO.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1617CodeFixProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+                var node = root.FindNode(diagnostic.Location.SourceSpan);
+                if (node.IsMissing)
+                {
+                    continue;
+                }
+
```
vweijsters: I don't think its possible to have a missing node for method / delegate declarations.

vweijsters:
The `RegisterCodeFixesAsync` is not following the async pattern that we seem to be using the newer code fixes

```C#
        public override Task RegisterCodeFixesAsync(CodeFixContext context)
        {
            foreach (Diagnostic diagnostic in context.Diagnostics.Where(d => FixableDiagnostics.Contains(d.Id)))
            {
                context.RegisterCodeFix(new PreservingDocumentChangeAction("Align curly brackets", token => GetTransformedDocumentAsync(context.Document, diagnostic, token)), diagnostic);
            }

            return Task.FromResult(true);
        }

        private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
        {
```

I think this results in unnecessary work being done when the code fix is not triggered by the developer. (@pdelvo or @sharwell can you confirm?)

vweijsters:
The construction where a changed document is being modified via the solution in `RemoveHelper` seems overkill to me.

As far as I can determine something like the code below will also work quite well and is a bit easier for the eyes.

```C#
private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
{
    /* lots of code removed */

    var newSyntaxRoot = root.RemoveNodes(nodesToFix, SyntaxRemoveOptions.KeepLeadingTrivia);
    return document.WithSyntaxRoot(newSyntaxRoot);
}
```

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1617CodeFixProvider.cs
sharwell: I prefer the way it's written now :smiley:

johnkoerner:
I was basing this on some of the SA13xx code fix providers and they seem to follow the same pattern where all of the code fix code is done in the `RegisterCodeFixesAsync` method.  But based on the feedback from @vweijsters , I now see that the SA15xx fix providers follow the `GetTransformedDocumentAsync` pattern.  I can do it either way. I think consistency would be nice in this area.  Maybe we could even write an analyzer to enforce this :smile:.

vweijsters:
:+1: Changes look good.

PR #670, opened 2015-04-21
Skip running StyleCop rules on generated files and elements

### LINKED ISSUE #555: Analyzers will report in auto generated files
As far as I know the only way to exclude them is to use something like [this](https://github.com/code-cracker/code-cracker/blob/08c1a01337964924eeed12be8b14c8ce8ec6b626/src/Common/CodeCracker.Common/Extensions/GeneratedCodeAnalysisExtensions.cs).

AArnott:
Fixes #555


AArnott:
Obviously this needs to be applied to *all* rules and not just SA1600. I didn't want to go too far before getting feedback as this is my first contribution. Please let me know what you think, including:

1. Should I replace *all* uses of `RegisterSyntaxNodeAction` with `RegisterSyntaxNodeActionHonorExclusions`? 
1. How do you like the name of the `RegisterSyntaxNodeActionHonorExclusions` method?
1. Do I need to add tests for this to every single analyzer or is it enough that it's tested once in the context of one of them?

sharwell:
I'm primarily concerned about the performance implications of this. In the proposed implementation, the check would be run once for each item analyzed, as opposed to once per file. When you consider all the other analyzers we have here, we could be talking about multiple calls to check for exclusions for *every token* in the input, including trivia tokens.

AArnott:
Good point. But how else can we skip an entire file? At the time the analyzer registers itself, we don't have a file context to check.

sharwell:
No idea. Best guess right now is it might be possible to use `RegisterSyntaxTreeAction` to get a callback that runs before the `RegisterSyntaxNodeAction` callbacks (I don't know when, if ever, this order is guaranteed to be correct for this use case), and "attach" a flag to the tree (e.g. using `ConditionalWeakTable`) that later `SyntaxNode` handlers look for.

If it worked, this would allow the result to be shared across all analyzers in the solution. If it worked.

sharwell:
Oh hey, it would certainly work if we treat the `ConditionalWeakTable` as an opportunistic cache, i.e. simply recompute the value any time it's absent and we need it. The check and conditional update could be inside the action you register with `RegisterSyntaxNodeActionHonorExclusions`.

AArnott:
Cool. I'll try that out and see if I can test its effectiveness as well.

AArnott:
What does the 'do not merge' label mean? Is that a bad thing?

sharwell:
No, it just means issues were identified during the last code review (it saves me time when looking through the list of outstanding PRs).

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1000KeywordsMustBeSpacedCorrectly.cs
```diff
@@ -60,7 +60,7 @@ public override void Initialize(AnalysisContext context)
             context.RegisterSyntaxTreeAction(this.HandleSyntaxTree);
 
             // handle nameof (which appears as an invocation expression??)
-            context.RegisterSyntaxNodeAction(this.HandleInvocationExpressionSyntax, SyntaxKind.InvocationExpression);
+            context.RegisterSyntaxNodeActionHonorExclusions(this.HandleInvocationExpressionSyntax, SyntaxKind.InvocationExpression);
```
sharwell: :exclamation: We're also going to need to handle `RegisterSyntaxTreeAction` above.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+            return false;
+        }
+
+        internal static bool IsGenerated(this SyntaxNode node) => node.HasAttributeOnAncestorOrSelf("DebuggerNonUserCode", "GeneratedCode");
+
+        internal static bool IsGenerated(this SyntaxTree tree, CancellationToken cancellationToken) => (tree.FilePath?.IsOnGeneratedFile() ?? false) || tree.HasAutoGeneratedComment(cancellationToken);
+
+        internal static bool HasAutoGeneratedComment(this SyntaxTree tree, CancellationToken cancellationToken)
+        {
+            lock (GeneratedHeaderPresentCheck)
```
sharwell: This entire block can be shortened to the following (no `lock` required, and note that `box` is used again in a comment below).

```csharp
StrongBox<object> box = GeneratedHeaderPresentCheck.GetOrCreateValue(tree);
if (box.Value != null)
{
    return box.Value == GeneratedHeaderPresent;
}
```

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (commentLines.Count != 2) return false;
+            bool autoGenerated = commentLines[1].ToString() == "// <auto-generated>";
+
+            lock (GeneratedHeaderPresentCheck)
+            {
+                // We don't know if there was a race, and we want to avoid throwing,
+                // so now that we have the lock, remove the key if it is already there.
+                GeneratedHeaderPresentCheck.Remove(tree);
+                GeneratedHeaderPresentCheck.Add(tree, autoGenerated ? GeneratedHeaderPresent : GeneratedHeaderAbsent);
+            }
```
sharwell: After replacing the above `lock` statement, this one can be replaced by the following (again, no `lock` necessary):

```csharp
box.Value = autoGenerated ? GeneratedHeaderPresent : GeneratedHeaderAbsent;
```

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
AArnott: ah yes! I was trying to remember how to make a value type appear as a reference type.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1000KeywordsMustBeSpacedCorrectly.cs
AArnott: Great. I had used a search and replace for the first method, so I missed this other one.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
AArnott: Yup. That makes sense since ConditionalWeakTable is thread-safe and its entry is now guaranteed. Nice.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            var commentLines = trivia.Where(t => t.IsKind(SyntaxKind.SingleLineCommentTrivia)).Take(2).ToList();
+            if (commentLines.Count != 2) return false;
+            return commentLines[1].ToString() == "// <auto-generated>";
+        }
+
+        private static bool IsOnGeneratedFile(this string filePath) =>
+            Regex.IsMatch(filePath, @"(^service|^TemporaryGeneratedFile_.*|^assemblyinfo|^assemblyattributes|\.(g\.i|g|designer|generated|assemblyattributes))\.(cs|vb)$",
+                RegexOptions.IgnoreCase);
```
sharwell: :bulb: Add `RegexOptions.ExplicitCapture`.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// A cache of the result of computing whether a document has an auto-generated header.
+        /// </summary>
+        /// <remarks>
+        /// This allows many analyzers that run on every token in the file to avoid checking
+        /// the same state in the document repeatedly.
+        /// </remarks>
+        private static readonly ConditionalWeakTable<SyntaxTree, StrongBox<bool?>> GeneratedHeaderPresentCheck = new ConditionalWeakTable<SyntaxTree, StrongBox<bool?>>();
+
+        internal static bool IsGenerated(this SyntaxNodeAnalysisContext context, CancellationToken cancellationToken) => (context.Node.SyntaxTree?.IsGenerated(cancellationToken) ?? false) || (context.Node?.IsGenerated() ?? false);
```
sharwell: :exclamation: I don't like how there are several `IsGenerated` methods, but they all have completely different behavior. In particular, the `SyntaxNode` method only looks at attributes, but the `SyntaxTree` method only looks at the file path and leading comment(s).

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                if (syntaxReference.SyntaxTree.IsGenerated(cancellationToken)) return true;
+                var root = syntaxReference.SyntaxTree.GetRoot(cancellationToken);
+                var node = root?.FindNode(syntaxReference.Span);
+                if (node.IsGenerated()) return true;
+            }
+            return false;
+        }
+
+        internal static bool IsGenerated(this SyntaxNode node) => node.HasAttributeOnAncestorOrSelf("DebuggerNonUserCode", "GeneratedCode");
```
sharwell: :exclamation: I disagree on `DebuggerNonUserCode`.

:question: Do we really need the `GeneratedCode` check? We are paying a heavy performance cost for this when most users would rely completely on file names and/or a comment in the header.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        internal static bool IsGenerated(this SyntaxTree tree, CancellationToken cancellationToken) => (tree.FilePath?.IsOnGeneratedFile() ?? false) || tree.HasAutoGeneratedComment(cancellationToken);
```
sharwell: :question: Is `tree.FilePath` the full path, or just the file name? The `IsOnGeneratedFile` method appears to assume it is just the file name.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+            else
+            {
+                if (!firstToken.HasLeadingTrivia) return false;
+                trivia = firstToken.LeadingTrivia;
+            }
+
+            var commentLines = trivia.Where(t => t.IsKind(SyntaxKind.SingleLineCommentTrivia)).Take(2).ToList();
+            if (commentLines.Count != 2) return false;
+            return commentLines[1].ToString() == "// <auto-generated>";
```
sharwell: This may be "too strict", in the sense that it fails to exclude files that people assume would be excluded (e.g. due to spacing differences). Of particular concern is the example of generated code given in the [documentation for SA1633](http://stylecop.com/docs/SA1633.html) would not be considered auto-generated according to this implementation (both for being on the first line, and for being a self-closing XML element instead of the one shown here).

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
AArnott: I agree. I have found that confusing as well. This came from the other analyzer repo linked to in the original issue.
I'll fix.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
AArnott: I agree on removing DebuggerNonUserCode (again, inherited code).
As for the GeneratedCode check, it makes sense to have it to me, and I suggest we leave it until perf measurements verify that it is hurting us. But I'll remove it if you want.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
AArnott: I agree, they treat this as just the filename. But I don't know what Roslyn's behavior is. In the tests, it's just a filename but I think the behavior can vary by host. So I've added `Path.GetFileName` to resolve.

StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
AArnott: I thought it was awfully strict myself. When I had written my own code for this it was more lenient. I'll fix

sharwell:
@AArnott pro-tip: GitHub does not send me a notification if you push more commits. You have to actually comment afterwards for it to notify me. :+1: (or break the build and have AppVeyor email me :wink:)

AArnott:
@sharwell thanks for that. I wasn't aware. :)

pdelvo:
I love this! I think it would be nice if the changes made to GeneratedCodeAnalysisExtensions would be contributed back to code cracker as well.

PR #671, opened 2015-04-21
Add *.log to .gitignore

AArnott:


sharwell:
:question: So we have a reference - can you clarify why these files would be created?

AArnott:
Ya... I ran MSBuild at the command line in the git repo with the `/fl` switch.

sharwell:
:+1:

PR #675, opened 2015-04-22
Add developmentDependency=true nuspec setting

AArnott:
Consistent with http://docs.nuget.org/Create/Nuspec-Reference, setting this prevents any project that adds the StyleCop.Analyzers package from expressing it as a dependency to its own respective users.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/StyleCop.Analyzers.nuspec
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/StyleCop.Analyzers.nuspec
@@ -13,6 +13,7 @@
     <releaseNotes>https://github.com/DotNetAnalyzers/StyleCopAnalyzers/releases/$version$</releaseNotes>
     <copyright>Copyright Sam Harwell 2015</copyright>
     <tags>StyleCop DotNetAnalyzers Roslyn Diagnostic Analyzer</tags>
+    <developmentDependency>true</developmentDependency>
   </metadata>
   <files>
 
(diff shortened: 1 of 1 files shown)


sharwell:
:exclamation: Can you also set the `minClientVersion`, as seen in tunnelvisionlabs/antlr4cs#62?

AArnott:
Done. But that change you link to only moves it to 2.7 for some reason, but 2.8 is what introduced the setting.
I wonder though if this is actually necessary. Even if 2.8 is the first client to support the property, if prior versions ignored it, it's probably better to not require the latest client.

sharwell:
The release notes for 2.7 mention this flag. Are they in error?
http://docs.nuget.org/release-notes/nuget-2.7

AArnott:
It seems NuGet docs have a contradiction here. 

http://docs.nuget.org/release-notes/nuget-2.7 claims it was added then
http://docs.nuget.org/Create/Nuspec-Reference claims it was added in 2.8.

I'll ask the NuGet folks.

sharwell:
I installed the NuGet.CommandLine package version 2.7.0 and opened up NuGet.exe in ILSpy. This is what I found:

image

I say we set the minimum version to 2.7 until we hear otherwise.

sharwell:
Also @AArnott I sent you a message on Gitter. :smile:

AArnott:
I think I found the explanation of the confusion between 2.7 and 2.8. 2.7 adds support for the attribute in the packages.config file, while 2.8 adds support for it being suggested by the nuspec file. So while 2.8 improves the experience and makes the attribute I add here worthwhile, I don't think we strictly need to require 2.8 just because we use its feature.

[More info](http://stackoverflow.com/questions/15012963/dont-include-dependencies-from-packages-config-file-when-creating-nuget-package/24216882#24216882)

PR #677, opened 2015-04-22
Implemented SA1205 (incl. codefix + tests)

### LINKED ISSUE #65: SA1205: PartialElementsMustDeclareAccess
http://www.stylecop.com/docs/SA1205.html

vweijsters:
Fixes #65

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1205CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1205CodeFixProvider.cs
@@ -1,0 +1,81 @@
+﻿namespace StyleCop.Analyzers.OrderingRules
+{
+    using System.Collections.Immutable;
+    using System.Composition;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Extensions;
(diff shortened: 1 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1205CodeFixProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+
+        /// <inheritdoc/>
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            foreach (Diagnostic diagnostic in context.Diagnostics.Where(d => FixableDiagnostics.Contains(d.Id)))
+            {
+                context.RegisterCodeFix(CodeAction.Create("Add public access modifier", token => GetTransformedDocumentAsync(context.Document, diagnostic, SyntaxKind.PublicKeyword, token)), diagnostic);
```
sharwell: :exclamation: The symbol will already have a correct access modifier (if no other fragment has a modifier, the default is `internal`). You'll need to get the declared symbol for the class and determine the effective access modifier from that symbol.

vweijsters: Should we go the extra mile and scan all `DeclaringSyntaxReferences` of the symbol? That way we can offer the option to declare the partial element public if none of the other declarations has an access modifier specified, which is more friendly than defaulting to internal.

There might be performance issues associated with it though. Any thoughts?

sharwell: Probably not. We don't do that with other rules like SA1400 (AccessModifierMustBeDeclared).

vweijsters:
Implemented new code fix strategy.

PR #698, opened 2015-04-23
Enhance SA1504 + enable by default

### LINKED ISSUE #690: SA1504 (AllAccessorsMustBeSingleLineOrMultiLine) should always be reported on the single-line accessor
Currently SA1504 is always reported on the *second* accessor, which results in the bizarre behavior shown in dotnet/roslyn#2199. This analyzer should be updated to report the diagnostic on whichever accessor is the single-line accessor, even if it appears first. This change in behavior will simplify the implementation of the quick fix, because the quick fix behavior will need to change the single-line accessor to be multi-line.

### LINKED ISSUE #691: SA1504 (AllAccessorsMustBeSingleLineOrMultiLine) should only be reported on the keyword
Currently SA1504 is reported on an entire accessor. It should be updated to only report the diagnostic on the keyword (e.g. `get` or `add`).

### LINKED ISSUE #697: SA1504 (AllAccessorsMustBeSingleLineOrMultiLine) should be enabled by default
It is currently implemented with tests, but has not been enabled.

@macpak gets the honors. :+1:

macpak:
Fixes #690
Fixes #691
Fixes #697

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1504AllAccessorsMustBeSingleLineOrMultiLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1504AllAccessorsMustBeSingleLineOrMultiLine.cs
@@ -70,7 +70,7 @@
         private const string HelpLink = "http://www.stylecop.com/docs/SA1504.html";
 
         private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.DisabledNoTests, Description, HelpLink);
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, true, Description, HelpLink);
 
         private static readonly ImmutableArray<DiagnosticDescriptor> SupportedDiagnosticsValue =
             ImmutableArray.Create(Descriptor);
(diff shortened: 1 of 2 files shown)


sharwell:
GitHub pro-tip: the way you wrote the summary, it only knows to close #697 when this gets merged. You need to have "Fixes" in front of *each* issue you want to close (that's why I use one per line).

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1504AllAccessorsMustBeSingleLineOrMultiLine.cs
```diff
@@ -96,35 +96,46 @@ private void HandlePropertyDeclaration(SyntaxNodeAnalysisContext context)
...
             }
 
-            if (!this.AreAllAccessorsInMultiLines(property.AccessorList.Accessors)
-               && !this.AreAllAccessorsInOneLine(property.AccessorList.Accessors))
+            var accessorsAnalyzeResult = this.AnalyzeAccessorsLineSpans(property.AccessorList.Accessors);
+            if (accessorsAnalyzeResult.SingleLineAccessors.Count == 1
+                && accessorsAnalyzeResult.MultiLineAccessors.Count == 1)
             {
-                context.ReportDiagnostic(Diagnostic.Create(Descriptor, property.AccessorList.Accessors[1].GetLocation()));
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, accessorsAnalyzeResult.SingleLineAccessors.Single().Keyword.GetLocation()));
```
sharwell: :exclamation: Use `First()` instead of `Single()` to guarantee we won't get an exception if there's ever a case where 3 or more accessors could exist (with 2 or more single-line, and 1 or more multi-line).

macpak:
Corrected.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1504AllAccessorsMustBeSingleLineOrMultiLine.cs
```diff
@@ -96,35 +96,46 @@ private void HandlePropertyDeclaration(SyntaxNodeAnalysisContext context)
...
 
-        private bool AreAllAccessorsInMultiLines(IEnumerable<AccessorDeclarationSyntax> accessors)
+        private class AccessorsAnalysisResult
         {
-            return accessors.All(a =>
-            {
-                var fileLinePositionSpan = a.GetLocation().GetLineSpan();
-                return fileLinePositionSpan.StartLinePosition.Line < fileLinePositionSpan.EndLinePosition.Line;
-            });
+            public List<AccessorDeclarationSyntax> MultiLineAccessors { get; set; } = new List<AccessorDeclarationSyntax>();
```
sharwell: :bulb: The setter here is unnecessary.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1504AllAccessorsMustBeSingleLineOrMultiLine.cs
```diff
@@ -96,35 +96,46 @@ private void HandlePropertyDeclaration(SyntaxNodeAnalysisContext context)
...
+
+            public List<AccessorDeclarationSyntax> SingleLineAccessors { get; set; } = new List<AccessorDeclarationSyntax>();
```
sharwell: :bulb: The setter here is unnecessary.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1504AllAccessorsMustBeSingleLineOrMultiLine.cs
macpak: Is it possible ? I check if we get one single and one multi by calling Count.

sharwell: Might as well make those `>=` comparisons. :)

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1504AllAccessorsMustBeSingleLineOrMultiLine.cs
macpak: Yeah, I will remove them (and the one below). Unfortunatelly, need to go now, but I will fix it in a few hours :) Thanks !

macpak:
Unused setters removed.

sharwell:
:+1:

PR #700, opened 2015-04-23
Fix that SA1516 is reported for single line accessors

### LINKED ISSUE #693: SA1516 (ElementsMustBeSeparatedByBlankLine) should allow property accessors to omit blank line when they are all one line
The following should be an allowed property definition by SA1516. The original StyleCop allowed it. 

```csharp
public string DisplayName
{
	get { return this.displayName; }
	set { this.displayName = value; }
}
```

However, as soon as *one* of the accessors became multi-line, they both had to be multi-line and a blank line is required between them.

pdelvo:
Fixes #693.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1516UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1516UnitTests.cs
@@ -411,8 +411,8 @@ public class Foo
 {
     public string FooProperty
     {
-       get { return ""bar""; }
-       set { }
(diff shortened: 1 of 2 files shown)


vweijsters:
Could you please add a test case to verify that SA1516 will be reported when a property has a single line accessor and a multi-line accessor (as mentioned in #693)

The code below should report SA1516.
```C#
public string DisplayName
{
    get { return this.displayName; }
    set
    {
        this.displayName = value;
    }
}
```

vweijsters:
:+1: Looks good otherwise.

pdelvo:
I added more tests

PR #701, opened 2015-04-23
Treat `t_` as a special prefix for rule SA1308

### LINKED ISSUE #688: Consider updating SA1308 (VariableNamesMustNotBePrefixed) to handle 't_'
Currently the prefix `t_` (used by some projects for thread-local variables) is not treated as a "special prefix" for rule SA1308. It is instead reported as SA1310 (FieldNamesMustNotContainUnderscore). I believe we should update rule SA1308 to handle this special case in addition to its current treatment of `m_` and `s_`.

sharwell:
Fixes #688


vweijsters:
:+1: Changes look good to me

PR #702, opened 2015-04-23
Fix SA1509 getting reported for consecutive blocks

### LINKED ISSUE #689: SA1509 (OpeningCurlyBracketsMustNotBePrecededByBlankLine) reported for consecutive blocks
Currently SA1509 is reported for the following code. If the space is removed, SA1513 (ClosingCurlyBracketMustBeFollowedByBlankLine) is reported.

```csharp
{
    var target = new int[s_manyElements.Length];
    s_manyElements.CopyTo(target);
    Assert.Equal(target, s_manyElements);
}

{
    var target = new int[0];
    Assert.Throws<NullReferenceException>(() => s_emptyDefault.CopyTo(target));
}
```

:question: What is the expected behavior in this situation?

sharwell:
Fixes #689

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1509UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1509UnitTests.cs
@@ -674,6 +674,122 @@ class Foo
         }
 
         [Fact]
+        public async Task TestBlockStatementsWithBlockCommentAsync()
+        {
+            var testCode = @"
+class Foo
+{
+    void Bar()
+    {
+        /* Comment */
+
+        {
+        }
+
+        {
+        }
+    }
+}";
+
+            var fixedCode = @"
+class Foo
+{
+    void Bar()
+    {
+        /* Comment */
+        {
+        }
+
+        {
+        }
+    }
+}";
+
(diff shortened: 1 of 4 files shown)


vweijsters:
What should happen with the code below? I think it will report SA1509, but I think it should not report any diagnostic.

```C#
class Foo
{
    void Bar()
    {
        /* test comment */

        {
        }

        {
        }
    }
}";
```

Przemyslaw-W:
I still claim it should raise SA1509 for first nested block in both cases:

```
class Foo
{
    void Bar()
    {
        /* test comment */

        {
        }

        {
        }
    }
}";
```
and
```
class Foo
{
    void Bar()
    {
        // test comment

        {
        }

        {
        }
    }
}";
```

Przemyslaw-W:
And also for:

```
class Foo
{
    void Bar()
    {
        #region Test region

        {
        }

        {
        }
        #endregion
    }
}";
```

sharwell:
@vweijsters The current code reports SA1509 in that case, and this pull request does not change that. Since that line is not between blocks, it's also outside the scope of this particular issue; you can file a new issue if you'd like to see the behavior in that case changed.

@Przemyslaw-W I added unit tests to verify that the behavior of this pull request adheres to your expectations.

pdelvo:
The docs should reflect the behavior.

oatkins:
:+1:

PR #704, opened 2015-04-24
Multiple fixes for SA1642 (ConstructorSummaryDocumentationMustBeginWithStandardText)

### LINKED ISSUE #418: SA1642 (constructor summary text) special case for structures
Currently SA1642 always enforces the use of the word "class" in the documentation for constructors. Should the constructors for a `struct` type "prefer" the use of another word here, in the same sense as #413 handled the wording for private constructors?

### LINKED ISSUE #676: SA1642 misfires on nested structs, requiring text describing the outer type
When a struct is nested in a class, its constructor is inappropriately required to carry the documentation of its parent class.

For example, the following correct constructor doc:

    class Foo
    {
        struct Bar
        {
            /// <summary>
            /// Initializes a new instance of the <see cref="Bar"/> struct.
            /// </summary>
            public Bar(int v) { }
        }
    }

Causes the SA1642 rule to misfire. The only way to appease it is to change it to refer to `Foo` (and class instead of struct, due to #418)

sharwell:
* Require `struct` instead of `class` for struct types (Fixes #418)
* Fix support for nested types (Fixes #676)
* Use semantic checking of the type name reference instead of text checking

This means either of the following are now allowed:

```csharp
class ClassName
{
  struct StructName
  {
    /// <summary>
    /// Initializes a new instance of the <see cref="StructName"/> struct.
    /// </summary>
    StructName(int argument)
    {
    }
  }
}
```

```csharp
class ClassName
{
  struct StructName
  {
    /// <summary>
    /// Initializes a new instance of the <see cref="ClassName.StructName"/> struct.
    /// </summary>
    StructName(int argument)
    {
    }
  }
}
```

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
@@ -382,10 +382,8 @@ class ClassName
 
         [Theory]
         [InlineData("class", "class", "class")]
-        [InlineData("class", "struct", "class")]
         [InlineData("class", "struct", "struct")]
         [InlineData("struct", "class", "class")]
-        [InlineData("struct", "struct", "class")]
         [InlineData("struct", "struct", "struct")]
         public async Task TestAllowedOuterQualifiedNames(string outerTypeKind, string nestedTypeKind, string describedTypeKind)
         {
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
@@ -143,73 +143,73 @@
         [Fact]
         public async Task TestNonPrivateConstructorCorrectDocumentationSimple()
         {
-            await this.TestConstructorCorrectDocumentationSimple("public", NonPrivateConstructorStandardText[0], NonPrivateConstructorStandardText[1], false).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


AArnott:
In two of your examples in the pull request description you use 'class' instead of 'struct' as the last word in the constructor summary, even though the constructor is for a struct. Is that intentional? That seems wrong.

sharwell:
@AArnott it is intentional. The analyzer allows this case (will not report a warning), even though the code fix would only ever generate the case where the keyword matches the type kind.

AArnott:
Why is it by design? The type named is not a class. It is most decidely a struct. And the only time I can think of that someone would say class instead of struct here is a copy and paste error from a class.

sharwell:
@AArnott The documentation for the StyleCop rule says to use "class". I was operating under the assumption that StyleCop enforced this so it was a compatibility measure, similar to the way we improved wording in #413 without breaking the old wording.

If StyleCop's documentation is simply incorrect (and it actually *enforced* the word `struct`), then we have no need to expand support to include both type kinds and we should make sure a bug is filed to report a warning if `class` is used in the documentation for a `struct`.

AArnott:
I just checked. StyleCop does _not_ permit the use of "class" for structs.

http://stylecop.codeplex.com/SourceControl/latest#Project/Src/AddIns/CSharp/Analyzers/DocumentationRules.cs
Look for the CheckConstructorSummaryText method.

sharwell:
Well that's a good thing. :smile: 

Would you be OK with addressing it as a separate issue/bug following this pull request?

AArnott:
yes

sharwell:
@AArnott oh well I fixed it. Can you review the change as a whole and give a :+1: or :-1: 

In return you'll get Alpha 5 :smile:

AArnott:
Great. I'll review now. You may want to update your pull request's description with the change to require 'struct'.

AArnott:
:+1:

PR #735, opened 2015-04-27
Improve performace of SystemDiagnosticsDebugDiagnosticBase by comparing strings

### LINKED ISSUE #733: SystemDiagnosticsDebugDiagnosticBase should check the method name before calling GetSymbolInfo
`GetSymbolInfo` is a (relatively) expensive call that appeared while investigating #732. We should check the text name of the method being invoked before calling `GetSymbolInfo` to filter out the vast majority of calls.

pdelvo:
Fixes #733.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/DebugMessagesUnitTestsBase.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/DebugMessagesUnitTestsBase.cs
@@ -340,6 +340,36 @@ public class Foo
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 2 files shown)


sharwell:
:exclamation: Can you add some unit tests that use `System.Debug` instead of just `Debug` to reference the class (since it's a slightly different tree)?

pdelvo:
This change will actually break Debug diagnostic in an edge case.
```csharp
using Foo = System.Diagnostics.Debug;

class Foo
{
    void Bar()
    {
        Foo.Assert(false);
    }
}
```

sharwell:
You sure about that? We are looking for `Assert`, not `Debug`.

sharwell:
:exclamation: Also add a check with `using static`, where the call appears as:

```csharp
using static System.Diagnostics.Debug;

...

void Bar()
{
    Assert(false);
}
```

pdelvo:
I implemented some unit tests and I fixed a bug i found while doing that

PR #743, opened 2015-04-27
Added WithSpan method to DiagnosticResult

### LINKED ISSUE #625: Consider testing complete diagnostic span instead of just start
The purpose of this is preventing regressions on issues like #618.

Currently the test framework does not support this, and we would need it to be relatively straightforward. I'm thinking an interesting possibility would be a `.WithSingleTokenSpan()` method for `DiagnosticResult` (expanding on #526).

vweijsters:
Fixes #625

Haven't added `.WithSingleTokenSpan()` because I can't really think of a proper way to do that at the moment.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
@@ -209,6 +209,16 @@ namespace TestHelper
             return new DiagnosticResult(descriptor);
         }
 
+        protected DiagnosticResult CSharpCompilerError(string errorIdentifier, string message)
+        {
+            return new DiagnosticResult
+            {
+                Id = errorIdentifier,
+                Severity = DiagnosticSeverity.Error,
+                Message = message,
+            };
+        }
+
         /// <summary>
         /// Create a project using the input strings as sources.
         /// </summary>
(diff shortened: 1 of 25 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -124,7 +126,7 @@ private static void VerifyDiagnosticResults(IEnumerable<Diagnostic> actualResult
                 var actual = actualResults.ElementAt(i);
                 var expected = expectedResults[i];
 
-                if (expected.Line == -1 && expected.Column == -1)
+                if (!expected.HasLocation)
```
sharwell: Nice.

vweijsters:
Updated to work with the latest master

vweijsters:
Rebased

pdelvo:
Can this be merged in? The code looks a lot cleaner.

vweijsters:
Rebased

pdelvo:
:+1:

StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -209,6 +209,16 @@ protected DiagnosticResult CSharpDiagnostic(DiagnosticDescriptor descriptor)
             return new DiagnosticResult(descriptor);
         }
 
+        protected DiagnosticResult CSharpCompilerError(string errorIdentifier, string message)
+        {
+            return new DiagnosticResult
+            {
+                Id = errorIdentifier,
+                Severity = DiagnosticSeverity.Error,
+                Message = message,
```
sharwell: :thought_balloon: It would be more consistent with the other `DiagnosticResult` factory methods if the message was omitted and left for a separate call to `WithMessage` instead.

PR #753, opened 2015-04-29
Modified analyzer to perform matching brace analysis at bit more efficient

### LINKED ISSUE #734: Rewrite SA1513ClosingCurlyBracketMustBeFollowedByBlankLine.IsOnSameLineAsOpeningBrace for performance
The line to `currentToken = currentToken.GetPreviousToken();` in this method is currently responsible for over 12% of the overall cost of running the currently available StyleCopAnalyzers analyzers. This issue was found while investigating #732.

vweijsters:
Fixes #734 

Haven't done the performance analysis, but the algorithm is much more efficient, so this should be good.

@sharwell How did you do the performance measurements?

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1513ClosingCurlyBracketMustBeFollowedByBlankLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1513ClosingCurlyBracketMustBeFollowedByBlankLine.cs
@@ -1,12 +1,15 @@
 ﻿namespace StyleCop.Analyzers.LayoutRules
 {
+    using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Linq;
+
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
     using Microsoft.CodeAnalysis.Text;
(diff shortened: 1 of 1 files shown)


sharwell:
I use the Visual Studio profiler. I initially start without profiling, and then after I load a solution I enable profiling and type for a while.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1513ClosingCurlyBracketMustBeFollowedByBlankLine.cs
```diff
@@ -66,204 +69,201 @@ public override void Initialize(AnalysisContext context)
             context.RegisterSyntaxTreeActionHonorExclusions(this.HandleSyntaxTreeAction);
         }
 
-        private void HandleSyntaxTreeAction(SyntaxTreeAnalysisContext context)
+        private async void HandleSyntaxTreeAction(SyntaxTreeAnalysisContext context)
```
sharwell: :exclamation: `async void` should be avoided.

vweijsters: Good point, I fixed it.

sharwell:
Verified the improvement. There are now 8 rules slower than this rule, and it's only 2.17% of the total time spent in StyleCopAnalyzers during a long profiling run of working on the Roslyn project.

PR #766, opened 2015-05-03
Fix that SA1404 does not support escaped identifiers

pdelvo:
Because everyone uses Justific\u0061tion


vweijsters:
:+1: Changes look good to me

PR #767, opened 2015-05-03
Update to 1.0.0-alpha006

sharwell:


pdelvo:
What does userInstalled="true" do? And why did you change the target framework?

sharwell:
@pdelvo Both of those changes were automatically getting applied by NuGet since CTP 6 (alpha 4). Previously I was manually backing them out before submitting a pull request.

pdelvo:
Looks good to me :+1:

PR #650, opened 2015-04-08
Code fix for using statements

vweijsters:
- [x] Fixes #291
- [x] Fixes #292
- [x] Fixes #293 
- [x] Fixes #294 
- [x] Fixes #295 
- [x] Fixes #1107 
- [x] Fixes #1108

PR #768, opened 2015-05-03
Implemented SA1216 + unit tests

### LINKED ISSUE #653: Using static directives must be placed after other using directives
I would like to propose a new diagnostic that covers a new language feature of C# 6:

**Category:** Ordering
**Rule:** Using static directives must be placed after other using directives
**Description:** A violation of this rule occurs when a `using static` directive is placed before a normal or an alias using directive. Placing the `using static` directives together below all other using-directives can make the code cleaner and easier to read, and can help make it easier to identify the static members used throughout the code.

**Rationale:** This proposed rule is an extension in line with rules SA1208 and SA1209 to cover a new language feature in C# 6. My proposal is to put the `using static` directives after the existing using directives in order to have minimal disturbance to patterns that developers are used to.

vweijsters:
Fixes #653

I also made a small change to SA1209, to prevent it from interfering with this rule.

There is no code fix yet, that should be covered by #650

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1216UsingStaticDirectivesMustBePlacedAtCorrectLocation.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1216UsingStaticDirectivesMustBePlacedAtCorrectLocation.cs
@@ -31,7 +31,7 @@
         private const string HelpLink = "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1216.md";
 
         private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.DisabledNoTests, Description, HelpLink);
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, true, Description, HelpLink);
 
         private static readonly ImmutableArray<DiagnosticDescriptor> SupportedDiagnosticsValue =
             ImmutableArray.Create(Descriptor);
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1216UsingStaticDirectivesMustBePlacedAtCorrectLocation.cs
```diff
@@ -0,0 +1,79 @@
...
+        private static readonly ImmutableArray<DiagnosticDescriptor> SupportedDiagnosticsValue =
+            ImmutableArray.Create(Descriptor);
+
+        /// <inheritdoc/>
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => SupportedDiagnosticsValue;
+
+        /// <inheritdoc/>
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionHonorExclusions(this.HandleUsingDirectiveSyntax, SyntaxKind.UsingDirective);
```
pdelvo: Have you considered looking at CompilationUnit/Namespace nodes instead? This rule might be a bit easier to implement that way and the algorithm could be O(n) instead of O(n^2) (if it even matters).

pdelvo:
It might be a good idea to add tests that verify the correct behavior if multiple using static directives are used.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1216UsingStaticDirectivesMustBePlacedAtCorrectLocation.cs
vweijsters: I haven't, I just copied most of the implementation from the SA1209 implementation. I'll have a look

pdelvo: SA1201 is looking at CompilationUnit/Namespace nodes.

sharwell: I created #776 to address SA1209.

vweijsters:
Updated after code review

pdelvo:
Very nice!

Looks good to me

sharwell:
Can you move the documentation to be a **SA1216.md** file within this repository? The wiki does not support pull requests, which makes it unnecessarily complex for external users to contribute to it.

vweijsters:
Moved the documentation.

PR #769, opened 2015-05-04
Implement proposed SA1028: NoTrailingWhitespace

### LINKED ISSUE #729: Rule proposal: NoTrailingWhitespace
As found [here](https://github.com/AArnott/nerdbank.stylecop.rules/blob/master/src/NerdBank.StyleCop.Rules/NerdBankRules.cs) and described [here](http://blog.nerdbank.net/2008/09/notrailingwhitespace-stylecop-rule-and.html), the NoTrailingWhitespace rule helps avoid code where there is invisible trailing whitespace at the end of a line of code. 

Git and some other tools will sometimes flag when there is trailing whitespace, so this helps detect it early.

This issue tracks porting this rule to StyleCop.Analyzers. Feel free to assign to me.

AArnott:
This includes the analyzer, the code fix provider, and the tests.

Of notable mention is that IIRC the original StyleCop did not notice trailing whitespace at the end of a single line or multi-line comment. I took some extra trouble to fill that gap. I believe this will find *all* trailing whitespace. Although we often want to emulate legacy StyleCop behavior, I believe this is simply filling a bug in the original and I expect especially given the auto fix provider that folks can quickly upgrade their code if necessary when applying this rule.

Fix #729

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
@@ -23,7 +23,10 @@ namespace TestHelper
         /// <returns>
         /// A new instance of the C# analyzer being tested.
         /// </returns>
-        protected abstract DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer();
+        protected virtual DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
+        {
+            return null;
+        }
 
         /// <summary>
         /// Called to test a C# <see cref="DiagnosticAnalyzer"/> when applied on the single input source as a string.
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
@@ -23,10 +23,7 @@ namespace TestHelper
         /// <returns>
         /// A new instance of the C# analyzer being tested.
         /// </returns>
-        protected virtual DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
-        {
-            return null;
-        }
+        protected abstract DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer();
 
         /// <summary>
         /// Called to test a C# <see cref="DiagnosticAnalyzer"/> when applied on the single input source as a string.
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
@@ -41,13 +41,18 @@
             {
                 var diagnosticSpan = diagnostic.Location.SourceSpan;
                 var trivia = root.FindTrivia(diagnosticSpan.Start, findInsideTrivia: true);
-                if (trivia.Kind() != SyntaxKind.None) // exclude XmlText node till we have support for it.
+                switch (trivia.Kind())
                 {
-                    context.RegisterCodeFix(
-                        CodeAction.Create(
-                            "Remove trailing whitespace",
-                            ct => RemoveWhitespaceAsync(context.Document, diagnostic, ct)),
-                        diagnostic);
+                    case SyntaxKind.WhitespaceTrivia:
+                    case SyntaxKind.SingleLineCommentTrivia:
+                    case SyntaxKind.MultiLineCommentTrivia:
+                    case SyntaxKind.MultiLineDocumentationCommentTrivia:
+                        context.RegisterCodeFix(
+                            CodeAction.Create(
+                                "Remove trailing whitespace",
+                                ct => RemoveWhitespaceAsync(context.Document, diagnostic, ct)),
+                            diagnostic);
+                        break;
                 }
             }
         }
@@ -87,9 +92,7 @@
                     newRoot = root.ReplaceTrivia(trivia, newTrivia);
                     break;
                 default:
-                    Debug.Assert(false, "Unexpected this is.");
-                    newRoot = root;
-                    break;
+                    return document;
             }
 
             var newDocument = document.WithSyntaxRoot(newRoot);
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
@@ -34,18 +34,22 @@
         }
 
         /// <inheritdoc/>
-        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
             foreach (var diagnostic in context.Diagnostics)
             {
-                context.RegisterCodeFix(
-                    CodeAction.Create(
-                        "Remove trailing whitespace",
-                        ct => RemoveWhitespaceAsync(context.Document, diagnostic, ct)),
-                    diagnostic);
+                var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken);
+                var diagnosticSpan = diagnostic.Location.SourceSpan;
+                var trivia = root.FindTrivia(diagnosticSpan.Start, findInsideTrivia: true);
+                if (trivia.Kind() != SyntaxKind.None) // exclude XmlText node till we have support for it.
+                {
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            "Remove trailing whitespace",
+                            ct => RemoveWhitespaceAsync(context.Document, diagnostic, ct)),
+                        diagnostic);
+                }
             }
-
-            return AnalyzerConstants.CompletedTask;
         }
 
         /// <summary>
@@ -83,8 +87,7 @@
                     newRoot = root.ReplaceTrivia(trivia, newTrivia);
                     break;
                 default:
-                    newRoot = root;
-                    break;
+                    throw new InvalidOperationException("Unexpected this is.");
             }
 
             var newDocument = document.WithSyntaxRoot(newRoot);
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Immutable;
     using System.Composition;
+    using System.Diagnostics;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
@@ -10,7 +11,6 @@
     using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.Text;
-
 
     /// <summary>
     /// Implements a code fix for <see cref="SA1028NoTrailingWhitespace"/>.
@@ -36,9 +36,9 @@
         /// <inheritdoc/>
         public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken);
             foreach (var diagnostic in context.Diagnostics)
             {
-                var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken);
                 var diagnosticSpan = diagnostic.Location.SourceSpan;
                 var trivia = root.FindTrivia(diagnosticSpan.Start, findInsideTrivia: true);
                 if (trivia.Kind() != SyntaxKind.None) // exclude XmlText node till we have support for it.
@@ -87,7 +87,9 @@
                     newRoot = root.ReplaceTrivia(trivia, newTrivia);
                     break;
                 default:
-                    throw new InvalidOperationException("Unexpected this is.");
+                    Debug.Assert(false, "Unexpected this is.");
+                    newRoot = root;
+                    break;
             }
 
             var newDocument = document.WithSyntaxRoot(newRoot);
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
@@ -9,6 +9,8 @@
     using Microsoft.CodeAnalysis.CodeActions;
     using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.Text;
+
 
     /// <summary>
     /// Implements a code fix for <see cref="SA1028NoTrailingWhitespace"/>.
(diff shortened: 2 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -23,10 +23,7 @@ public abstract partial class DiagnosticVerifier
         /// <returns>
         /// A new instance of the C# analyzer being tested.
         /// </returns>
-        protected virtual DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
-        {
-            return null;
-        }
+        protected abstract DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer();
```
vweijsters: A very good change, but I think it would be better to make a separate pull request for it.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+
+        /// <inheritdoc/>
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken);
```
vweijsters: This line should be moved outside of the foreach loop for performance reasons.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// Removes trailing whitespace.
+        /// </summary>
+        /// <param name="document">The document to be changed.</param>
+        /// <param name="diagnostic">The diagnostic to fix.</param>
+        /// <param name="cancellationToken">The cancellation token associated with the fix action.</param>
+        /// <returns>The transformed document.</returns>
+        private static async Task<Document> RemoveWhitespaceAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        {
+            var root = await document.GetSyntaxRootAsync(cancellationToken);
```
vweijsters: Passing the root as parameter to this method will be more efficient.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+            var trivia = root.FindTrivia(diagnosticSpan.Start, findInsideTrivia: true);
```
vweijsters: Passing the trivia as parameter to this method will be more efficient.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                    string oldTriviaContent = trivia.ToFullString();
+                    TextSpan diagnosticSpanWithinTrivia = TextSpan.FromBounds(diagnosticSpan.Start - triviaLocation.SourceSpan.Start, diagnosticSpan.End - triviaLocation.SourceSpan.Start);
+                    newTriviaContent = string.Concat(
+                        oldTriviaContent.Substring(0, diagnosticSpanWithinTrivia.Start),
+                        oldTriviaContent.Substring(diagnosticSpanWithinTrivia.End));
+                    newTrivia = SyntaxFactory.SyntaxTrivia(trivia.Kind(), newTriviaContent);
+                    newRoot = root.ReplaceTrivia(trivia, newTrivia);
+                    break;
+                default:
+                    throw new InvalidOperationException("Unexpected this is.");
```
vweijsters: How will VS2015 handle this exception? Is there a risk that it will disable the analyzer?

Maybe we should go with something like:
```C#
    Debug.Fail("Unexpected this is.");
    newRoot = root;
    break;
```

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028NoTrailingWhitespace.cs
```diff
@@ -0,0 +1,157 @@
...
+    public class SA1028NoTrailingWhitespace : DiagnosticAnalyzer
+    {
+        /// <summary>
+        /// The ID for diagnostics produced by the <see cref="SA1028NoTrailingWhitespace"/> analyzer.
+        /// </summary>
+        public const string DiagnosticId = "SA1028";
+        internal const string Title = "Code must not contain trailing whitespace";
+        internal const string MessageFormat = "Remove trailing whitespace";
+        internal const string Category = "Style";
+
```
vweijsters: Would be nice if you could provide a help link to the GitHub wiki for the project.
(See #768 for an example)

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028NoTrailingWhitespace.cs
```diff
@@ -0,0 +1,157 @@
...
+                        }
+
+                        if (reportWarning)
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(Rule, trivia.GetLocation()));
+                        }
+
+                        break;
+                    case SyntaxKind.SingleLineCommentTrivia:
+                        var text = root.GetText();
```
vweijsters: This can be done outside of the foreach loop, will improve performance.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028NoTrailingWhitespace.cs
```diff
@@ -0,0 +1,157 @@
...
+                        TextSpan trailingWhitespace = FindTrailingWhitespace(text, trivia.Span);
+                        if (!trailingWhitespace.IsEmpty)
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(Rule, Location.Create(context.Tree, trailingWhitespace)));
+                        }
+
+                        break;
+                    case SyntaxKind.MultiLineCommentTrivia:
+                    case SyntaxKind.SingleLineDocumentationCommentTrivia:
+                        text = root.GetText();
```
vweijsters: This can be done outside of the foreach loop, will improve performance.

vweijsters:
:+1: Looks good. I've made a few remarks (mostly performance related)

vweijsters:
We should definitely profile this one to see if there may be performance issues. Its scanning a lot of items each time the syntax tree is modified.

vweijsters:
@AArnott Should this rule also cover trailing whitespace after conditional directives (like `#if (Debug) `)?
I couldn't find any evidence that it does.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
sharwell: Agreed, and it should be done in the same pull request as addressing `GetCSharpCodeFixProvider`.

AArnott: Reverted. I'll be happy to send another pull request.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
AArnott: Quite right.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
AArnott: I considered it, but decided not to for a few reasons:

1. I expect Roslyn caches the result after the first call, making subsequent calls cheap anyway.
1. When that delegate is invoked, we're applying a code fix, which is a relatively expensive operation anyway, so this operation that was cheap enough to run thousands of times for lots of analyzers can certainly afford to run one more time as part of a user initiated action to apply a code fix.
1. Another parameter on a method makes it that much harder to call, and makes the method that much less helpful to its caller, since the caller has to obtain a value that the helper method could obtain itself.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
AArnott: Same answer as passing in the SyntaxRoot.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
AArnott: Ya, I wasn't sure what was the best way to go here. So I've made the change you suggest.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
sharwell: @AArnott The analyzers that don't have a code fix should technically extend `DiagnosticVerifier`, not `CodeFixVerifier`.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028NoTrailingWhitespace.cs
sharwell: I would prefer to see a **SA1028.md** file in this repository, since the wiki does not allow for pull requests.

AArnott: Done

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028NoTrailingWhitespace.cs
AArnott: fixed

AArnott:
Good catch on the condition directives. I'm working on adding that now.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+                    TextSpan diagnosticSpanWithinTrivia = TextSpan.FromBounds(diagnosticSpan.Start - triviaLocation.SourceSpan.Start, diagnosticSpan.End - triviaLocation.SourceSpan.Start);
+                    newTriviaContent = string.Concat(
+                        oldTriviaContent.Substring(0, diagnosticSpanWithinTrivia.Start),
+                        oldTriviaContent.Substring(diagnosticSpanWithinTrivia.End));
+                    newTrivia = SyntaxFactory.SyntaxTrivia(trivia.Kind(), newTriviaContent);
+                    newRoot = root.ReplaceTrivia(trivia, newTrivia);
+                    break;
+                default:
+                    Debug.Assert(false, "Unexpected this is.");
+                    newRoot = root;
```
sharwell: :bulb: I've seen every manner of horrible handling on the IDE side for this. It may have been fixed, but the sure way to not have it go wrong is to just `return document;` (the input unaltered).

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028NoTrailingWhitespace.cs
```diff
@@ -0,0 +1,160 @@
...
+        /// <summary>
+        /// The ID for diagnostics produced by the <see cref="SA1028NoTrailingWhitespace"/> analyzer.
+        /// </summary>
+        public const string DiagnosticId = "SA1028";
+        private const string Title = "Code must not contain trailing whitespace";
+        private const string MessageFormat = "Remove trailing whitespace";
+        private const string Category = "Style";
+        private const string Description = "There should not be any whitespace at the end of a line of code.";
+        private const string HelpLink = "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/wiki/SA1028";
+        private const bool IsEnabledByDefault = true;
```
sharwell: :bulb: It would make sense (as a separate pull request) to create a new property `AnalyzerConstants.EnabledByDefault` for this, and use it consistently through the project. I created issue #783 for this.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028NoTrailingWhitespace.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        private static DiagnosticDescriptor Rule { get; } =
```
sharwell: :bulb: This line doesn't follow the pattern used by all the other diagnostics in the project.

sharwell:
I'm hesitant to enable this by default prior to seeing it in action under the profiler, but I do want to get it included. Can you make sure the following items are clearly addressed in the documentation for the rule? For each case, please indicate whether or not the analyzer checks for trailing whitespace in this location, and also whether or not the code fix removes trailing whitespace in this location.

* :question: What is the behavior of the rule in disabled code (e.g. between `#if false` and `#endif`)?
* :question: What is the behavior of the rule within line comments?
* :question: What is the behavior of the rule within documentation comments?
* :question: What is the behavior of the rule within verbatim string literals?

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
AArnott: I'll fix that.

AArnott:
@sharwell I'm pleased to say that all your questions are already answered by tests I've written, with the exception of the first one (condition directives) which I'm adding now. But the answers so far, and planned for the first one, is that the analyzer is active in all those contexts. And I can certainly run a perf trace on a project with this turned on to see how much time it adds.

sharwell:
I'm really hoping the IDE will delay evaluation of this rule long enough that it's not constantly appearing at the caret location while typing code...

AArnott:
Good point, @sharwell. I wonder if there's a way to detect when there is an incomplete statement and skip it. No point in reporting trailing whitespace when the line itself is incomplete or wrong in some other way.

sharwell:
We could check if the `CompilationUnitSyntax`'s `ContainsDiagnostics` property is true, but that would cause *all* diagnostics in the file to disappear mid-statement while typing. It also doesn't cover typing in comments.

Best thing right now would probably be to experiment with it and see if it is annoying.

AArnott:
Aside: As usual, any time I'm using VS 2015, I find a bug. This time I found a gross memory leak. Just in working with StyleCopAnalyzers, devenv.exe is taking 1.2GB of RAM. Time to file another bug.

AArnott:
I've tested the performance of SA1028 and it's hard to even get it to show up on traces. And when it does, it's very small. So I think we're OK here. This doesn't surprise me because I was careful to avoid allocations or expensive work as much as possible.

I also tried the experience of simply typing code while this rule was turned on. It doesn't look like it pops up while I'm typing much. I can find it if I pause and look through the error list but otherwise it wasn't noticeable. I recommend we merge this.

PR #770, opened 2015-05-04
Additional tests and fixes for SA1005

### LINKED ISSUE #715: SA1005 does not allow indentation of comments
I don't know how the original StyleCop rule behaved, so this may or may not be by design. But even if it's by design, I'd like to challenge it to allow for a comment like this:

            // PERF: This has shown up on ETL traces as inefficient and expensive
            //       WithPart should call WithParts instead, and WithParts should
            //       execute a more efficient batch operation.

Currently, the rule complains that the second and third lines do not begin with "a space".
This certainly warrants discussion around exactly what the refined rule might be.

oatkins:
Now handles indented comments following a non-indented comment; comments starting with more than two dashes.

Modified diagnostic to address the above.

Fixes #715

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1005SingleLineCommentsMustBeginWithSingleSpace.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1005SingleLineCommentsMustBeginWithSingleSpace.cs
@@ -117,19 +117,25 @@
         private void HandleSingleLineCommentTrivia(SyntaxTreeAnalysisContext context, SyntaxTrivia trivia, bool isFirstSingleLineTrivia)
         {
             string text = trivia.ToFullString();
-            if (text.Equals("//"))
+            if (text.Equals(@"//"))
             {
                 return;
             }
 
             // special case: commented code
-            if (text.StartsWith("////", StringComparison.Ordinal))
+            if (text.StartsWith(@"////", StringComparison.Ordinal))
             {
                 return;
             }
 
             // Special case: multiple dashes at start of comment
-            if (text.StartsWith("//--", StringComparison.Ordinal))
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1005SingleLineCommentsMustBeginWithSingleSpace.cs
```diff
@@ -80,21 +81,40 @@ public override void Initialize(AnalysisContext context)
...
+            int newLineCount = 0;
+
             foreach (var trivia in root.DescendantTrivia())
             {
                 switch (trivia.Kind())
                 {
-                case SyntaxKind.SingleLineCommentTrivia:
-                    this.HandleSingleLineCommentTrivia(context, trivia);
-                    break;
+                    case SyntaxKind.SingleLineCommentTrivia:
```
pdelvo: Could you please revert these indentation changes?

pdelvo:
Besides the changes in indentation this looks good to me!

Do you think we should exclude //\ as well? I removed it from my list because I dont really like it. What do you think?

oatkins:
I've made the formatting change. Do I need to submit a new pull request?

Regarding "//\", I would tend to favour keeping the existing SC behaviour (and hence adjusting the code to accept this pattern) simply because someone, somewhere will be expecting it to work that way—and it is clearly the intended behaviour of the original code, not just a bug.

pdelvo:
No you dont need to send a new PR. If you add commits to the branch you opened the PR for the changes will show up automatically. You should however write a comment. Github unfortunately does not send notifications for commits in PRs.

Im okay with adding the //\ rule for compatibility.

oatkins:
Sorry, seems as though we're going round in circles on this one.

I //have// made the change to allow "//\..." (along with a matching test) but I'm not pressing for it if you think it's silly.

PR #786, opened 2015-05-05
Implemented SA1217 + unit tests

### LINKED ISSUE #652: Using Static directives must be ordered alphabetically
I would like to propose a new diagnostic that covers a new language feature of C# 6:

**Category:** Ordering
**Rule:** Using Static directives must be ordered alphabetically
**Description:** A violation of this rule occurs when `using static` directives are not sorted alphabetically by the fully qualifier class name. Sorting the `using static` directives alphabetically can make the code cleaner and easier to read, and can help make it easier to identify the namespaces that are being used by the code.

**Rationale:** This proposed rule is an extension in line with rules SA1210 and SA1211 to cover a new language feature in C# 6

vweijsters:
Fixes #652 

There is no code fix yet, that should be covered by #650

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1217UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1217UnitTests.cs
@@ -1,0 +1,131 @@
+﻿namespace StyleCop.Analyzers.Test.OrderingRules
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.OrderingRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1217UsingStaticDirectivesMustBeOrderedAlphabetically"/>.
+    /// </summary>
+    public class SA1217UnitTests : CodeFixVerifier
+    {
+        /// <summary>
+        /// Verifies that the analyzer will properly handle an empty source.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestEmptySourceAsync()
+        {
+            var testCode = string.Empty;
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will not produce diagnostics for correctly ordered using directives inside a namespace.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestValidUsingDirectivesInNamespaceAsync()
+        {
+            var testCode = @"namespace Foo
+{
+    using System;
+    using Execute = System.Action;
+    using static System.Array;
+    using static System.Math;
+}
+";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will not produce diagnostics for correctly ordered using directives inside a namespace.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestValidUsingDirectivesInMultipleNamespacesAsync()
+        {
+            var testCode = @"namespace Foo
+{
+    using System;
+    using Execute = System.Action;
+    using static System.Array;
+    using static System.Math;
+}
+
+namespace Bar
+{
+    using System;
+    using static System.Array;
+    using Execute = System.Action;
+    using static System.Math;
+}
+";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will not produce diagnostics for correctly ordered using directives in the compilation unit.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestValidUsingDirectivesInCompilationUnitAsync()
+        {
+            var testCode = @"using System;
+using static System.Array;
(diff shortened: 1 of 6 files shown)


pdelvo:
Within using directives System* should be at the top. Should this also apply for using static directives?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1217UsingStaticDirectivesMustBeOrderedAlphabetically.cs
```diff
@@ -0,0 +1,87 @@
...
+            UsingDirectiveSyntax lastStaticUsingDirective = null;
+
+            foreach (var usingDirective in usingDirectives)
+            {
+                if (usingDirective.StaticKeyword.IsKind(SyntaxKind.StaticKeyword))
+                {
+                    if (lastStaticUsingDirective != null)
+                    {
+                        var firstName = lastStaticUsingDirective.Name.ToFullString();
+                        var secondName = usingDirective.Name.ToFullString();
```
pdelvo: Why ToFullString? I dont think ToFullString would work correctly in this example:
```csharp
using static
//a
B;
using static
//b
A;
```

vweijsters:
I'm undecided whether or not the System.* stuff should be on top for the `using static` as well. My expectation is that `using static` will not be frequently used, so there would be limited benefit from readability point of view.

sharwell:
I'm undecided as well.

* What does a default installation of Visual Studio do?
* Does the "Place System using directives first" option affect `using static` directives?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1217UsingStaticDirectivesMustBeOrderedAlphabetically.cs
vweijsters: Just tested that and it will work correctly, but I will change the implementation to `.ToString()` anyway, just to be sure.

vweijsters:
Made a small improvement to properly handle aliases as well.
Also added a test case with inline comments.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NameSyntaxHelpers.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Class containing the extension methods for the <see cref="NameSyntax"/> class.
+    /// </summary>
+    internal static class NameSyntaxHelpers
+    {
+        private const string AliasSeparator = "::";
+
+        /// <summary>
+        /// Gets the name contained in the <see cref="NameSyntax"/>, without an alias prefix.
+        /// </summary>
+        /// <param name="nameSyntax">The <see cref="NameSyntax"/> to get.</param>
```
pdelvo: The param doc is a bit confusing

pdelvo:
The code looks good. This PR kind of blocks #790 because of the helper method that could be shared.

I also think that using static is not going to be used very frequently. It might be a good idea to add a rule that limits the number of using static directives in each scope.

sharwell:
@pdelvo I think the only such rule that would present a consistent message to users is "do not use `using static`". StyleCop generally leans towards qualifying references where possible.

vweijsters:
Improved ToUnaliasedString() documentation.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1217UnitTests.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestInvalidUsingDirectivesWithGlobalPrefixAsync()
+        {
+            var testCode = @"namespace Foo
+{
+    using System;
+    using Execute = System.Action;
+    using static System.Math;
+    using static global::System.Array;
```
sharwell: Visual Studio does not have an option to automatically sort using directives ignoring the alias, which means we're going to have some "annoyances" if we ignore them. In particular, a solution-wide "organize using operation" would incorrectly order using directives in files containing aliases. On the other hand if we do consider the using alias then we'll have to make a decision regarding the impact of an alias on SA1208 (SystemUsingDirectivesMustBePlacedBeforeOtherUsingDirectives).

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1217UsingStaticDirectivesMustBeOrderedAlphabetically.cs
```diff
@@ -0,0 +1,88 @@
...
+            foreach (var usingDirective in usingDirectives)
+            {
+                if (usingDirective.StaticKeyword.IsKind(SyntaxKind.StaticKeyword))
+                {
+                    if (lastStaticUsingDirective != null)
+                    {
+                        var firstName = lastStaticUsingDirective.Name.ToUnaliasedString();
+                        var secondName = usingDirective.Name.ToUnaliasedString();
+
+                        if (string.Compare(firstName, secondName, StringComparison.Ordinal) > 0)
```
sharwell: :exclamation: `OrdinalIgnoreCase`

documentation/SA1217.md
```diff
@@ -0,0 +1,30 @@
...
+</tr>
+<tr>
+  <td>CheckId</td>
+  <td>SA1217</td>
+</tr>
+<tr>
+  <td>Category</td>
+  <td>Ordering Rules</td>
+</tr>
+</table>
```
sharwell: If only we needed column headers...

```
| | |
| --- | --- |
| TypeName | SA1217UsingStaticDirectivesMustBeOrderedAlphabetically |
| CheckId | SA1217 |
| Category | Ordering Rules |
```

vweijsters: I know, I started out that way, but it didn't really look nice to me :wink:

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1217UnitTests.cs
vweijsters: StyleCop also ignores aliases when determining sort order for namespaces. I think we should stick to that behavior, to be as compatible as possible. That would mean adapting SA1208 (and potentially SA1209, SA1210 and SA1211 as well).

Besides that I think this will hardly ever be an issue, since aliases tend to be avoided by most developers because of the mess they can create.

vweijsters:
Made namespace comparison case insensitive

PR #787, opened 2015-05-05
Fix that SA1300 is reported for overridden members or members needed for...

### LINKED ISSUE #771: SA1300 misfires on members that implement interfaces
All rules that enforce naming conventions of members should *not* fire when the member provides an interface implementation. 

The rule *can* and probably should fire on the member in its interface declaration, but not on its implementations. For instance:

On the interface, the rule may fire:

```csharp
interface IVsCfg
{
    int get_DisplayName(out string name);
}
```

But don't blame the implementation of such a method:

```csharp
class MyVsCfg : IVsCfg
{
    public int get_DisplayName(out string name)
    {
        name = null;
        return 0;
    }
}
```

Another example of a place to avoid reporting a diagnostic is explicit interface implementations:

```csharp
class MyVsCfg : IVsCfg
{
    int IVsCfg.get_DisplayName(out string name)
    {
        name = null;
        return 0;
    }
}
```

pdelvo:
... implementing an interface

Fixes #771.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
@@ -1,6 +1,7 @@
 ﻿namespace StyleCop.Analyzers.Helpers
 {
     using System;
+    using System.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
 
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
```diff
@@ -91,5 +92,25 @@ internal static Location GetNameOrIdentifierLocation(SyntaxNode member)
...
+        /// If a derived class is implementing an interface and this member is required for it
+        /// this method will still return false.
+        /// </remarks>
+        /// <param name="memberSymbol">The member symbol that should be analyzed.</param>
+        /// <returns>true if the member is implementing an interface member, otherwise false.</returns>
+        internal static bool IsImplementingAnInterfaceMember(ISymbol memberSymbol)
+        {
+            var typeSymbol = memberSymbol.ContainingType;
+
+            return typeSymbol != null && typeSymbol.AllInterfaces
```
sharwell: I'm definitely going to want to find a more efficient way to handle this.

pdelvo: This is only called if the diagnostic would have been reported in the old version. So basicly in a project that follows this rule never.

pdelvo:
I improved the performance of this method a bit
1. Check if the symbol is a method, a property or an event. Otherwise it cant implement an interface
2. Check if it is implementing a member explicitly
3. Now that I only have to check implicit implementations I can pass the name of this method to GetMembers to reduce the number of checks I have to do.

sharwell:
I verified that it's fast by changing SA1300 locally to enforce a *lowercase* first letter instead of what it does now, which triggers a violation on nearly every member in the project. No noticeable impact resulted. :+1:

PR #788, opened 2015-05-06
Fix naming of variable

pdelvo:
Queue -> Stack.

This bothered me


vweijsters:
Good catch :smile: I guess you can easily determine what the initial type of that field was...

PR #804, opened 2015-05-09
Update all code fix providers following an API change

### LINKED ISSUE #803: Code fix providers do not properly set the name
In dotnet/roslyn@7c885ca20209ca95cfec1ed5bfaf1d43db06be99, the API for `ExportCodeFixProviderAttribute` changed. While the current use of this attribute still compiles, the semantics have become that each code fix supports two languages (one of which is the name of the code fix class), and do not have any name.

sharwell:
Fixes #803


vweijsters:
:+1: Looks good

PR #807, opened 2015-05-10
Updated installation scripts and package layout

### LINKED ISSUE #806: Make sure the analyzers work on asp.net 5
Currently the analyzers dont work when added to an asp.net 5 project.

I tried adding the nuget package to an asp.net 5 project. That added a reference under dependencies in the project.json file. The diagnostics do not trigger and I dont see a place to enalbe/disable analyzers.

Im not sure if analyzers are supported in asp.net 5 projects. The only thing I found was this issue(https://github.com/aspnet/Home/issues/196), but the issue is a bit old.

sharwell:
* Use the signed general-purpose **install.ps1** and **uninstall.ps1** from the Roslyn analyzers project
* Move the output assembly into a C#-specific subfolder in the NuGet package (should fix #806)

@pdelvo To test this, you can create a new **.nupkg** locally to move the assembly and use the new scripts, and then try toinstall that into an ASP.NET 5 project.


sharwell:
This did not solve the intended problem.

PR #811, opened 2015-05-10
Fix packaging

sharwell:
* Use the signed general-purpose **install.ps1** and **uninstall.ps1** from the Roslyn analyzers project
* Move the output assembly into a C#-specific subfolder in the NuGet package (prevents the analyzers from being enabled in VB projects)


pdelvo:
looks good

PR #828, opened 2015-05-11
SA1209 Efficiency Improvement

### LINKED ISSUE #288: Diagnostic unit tests: SA1209 (UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives)
Implement unit tests for #69.

### LINKED ISSUE #776: SA1209 (UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives) efficiency improvement
Currently SA1209 is implemented by registering a handler for using directives. This could be improved by instead registering a handler for compilation unit and namespaces, similar to the way SA1201 is implemented.

baks:
Fixes #288
Fixes #776 

Changed SA1209 implementation to register handlers for CompilationUnit
and NamespaceDeclaration instead of UsingDirectives.

Added tests for SA1209 rule.

Actually SA1209 doesn't report diagnostic if using alias directive is before/after using static. Should this be changed?

I didn't use helper method in @vweijsters PR #786 to get unaliased namespace name because it isn't merged to master. When it will be merged I also update my PR.


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1209UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives.cs
```diff
@@ -47,64 +51,67 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+        {
+            var namespaceDeclaration = context.Node as NamespaceDeclarationSyntax;
+
+            ProcessUsingsAndReportDiagnostic(namespaceDeclaration.Usings, context);
+        }
+
+        private static void ProcessUsingsAndReportDiagnostic(SyntaxList<UsingDirectiveSyntax> usings, SyntaxNodeAnalysisContext context)
+        {
+            UsingDirectiveSyntax usingAliasDirectivesShouldBePlacedAfterThis = null;
+            var usingAliasDirectivesToReport = new Lazy<List<UsingDirectiveSyntax>>();
```
vweijsters: I don't think its very useful to use Lazy<T> here. The collection is empty when created (so its cheap to construct) and it will be used almost immediately (if there is anything to report). In my opinion the benefits of using Lazy are outweighed here by the increased complexity (most developers I know will have to think twice when they see such a construction).

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1209UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives.cs
```diff
@@ -47,64 +51,67 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
-                    continue;
+            if (usingAliasDirectivesToReport.IsValueCreated && usingAliasDirectivesToReport.Value.Count > 0)
+            {
+                var unaliasedNamespaceName = GetNamespaceNameWithoutAlias(usingAliasDirectivesShouldBePlacedAfterThis.Name.ToString());
+                foreach (var usingAliasDirectiveToReport in usingAliasDirectivesToReport.Value)
+                {                    
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, usingAliasDirectiveToReport.GetLocation(), usingAliasDirectiveToReport.Alias.Name.ToString(), unaliasedNamespaceName));
                 }
+            }
+        }
```
vweijsters: I think the following case would produce wrong diagnostic messages for the namespace where the aliased using should be placed after:

```C#
using System;
using IO = System.IO;
using static System.Math;
using Threads = System.Threading
```

using static should be  ignored when setting `usingAliasDirectivesShouldBePlacedAfterThis`,

vweijsters: If the list of usings ends with several aliased usings, the message will suggest to place the reported aliased using (IO in the example) after the last aliased using statement (`System.Threading` in the example), instead of the alphabetically correct location (`System.Net` in the example).

Example:
```C#
using System;
using IO = System.IO;
using System.Net;
using Numerics = System.Numerics;
using Threads = System.Threading;
```

vweijsters: I would prefer that the diagnostic message contains the entire alias, when a suggestion is made to place the reported aliased using after another aliased using. I think it will improve the readability.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1209UnitTests.cs
```diff
@@ -0,0 +1,200 @@
...
+            };
+
+            var expectedForNamespaceDeclaration = new[]
+            {
+                this.CSharpDiagnostic().WithLocation("Test2.cs", 5, 5).WithArguments("N", "System.Threading.Tasks"),
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(usingsInCompilationUnit, expectedForCompilationUnit, CancellationToken.None);
+            await this.VerifyCSharpDiagnosticAsync(usingsInNamespaceDeclaration, expectedForNamespaceDeclaration, CancellationToken.None);
+        }
```
vweijsters: I find the format of the test case pretty difficult to read, due to the multiple source files used. I don't see a direct need for the defined namespaces, as they will not influence the code under test. In my opinion they can be replaced by system namespaces without losing anything from the test cases.

I would suggest something like this:
```C#
        [Fact]
        public async Task TestWhenUsingAliasDirectivesAreNotPlacedCorrectly()
        {
            var testCodeCompilationUnit = @"using TasksNamespace = System.Threading.Tasks;
using System.Net;
using System;
using System.IO;
using System.Numerics;
class A
{
}";

            var testCodeNamespace = @"namespace Test
{
    using System.Net;
    using System.Threading;
    using N = System.Numerics;
    using System.IO;
    using P = System.Threading.Tasks;
    class A
    {
    }
}"

            DiagnosticResult[] expectedForCompilationUnit =
            {
                this.CSharpDiagnostic().WithLocation(1, 1).WithArguments("TasksNamespace", "System.Numerics")
            };

            DiagnosticResult[] expectedForNamespaceDeclaration =
            {
                this.CSharpDiagnostic().WithLocation(5, 5).WithArguments("N", "System.Threading.Tasks")
            };

            await this.VerifyCSharpDiagnosticAsync(testCodeCompilationUnit, expectedForCompilationUnit, CancellationToken.None);
            await this.VerifyCSharpDiagnosticAsync(testCodeNamespace, expectedForNamespaceDeclaration, CancellationToken.None);
        }
```

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1209UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives.cs
baks: I looked at this again and you're right

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1209UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives.cs
baks: Just to clarify I wanna ask if ordering shouldn't be done by SA1211(UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName) ?

baks: @vweijsters you mean diagnostic message like that:
> Using alias directive for Numerics = System.Numerics must appear after directive for IO = System.IO

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1209UnitTests.cs
baks: @vweijsters good catch on creating array for expected diagnostics. 
I was using var and new[], but I see now, thanks to your example that better readability is when you explicitly declare array type.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1209UsingAliasDirectivesMustBePlacedAfterOtherUsingDirectives.cs
vweijsters: Yes

vweijsters: Ordering should be verified by SA1211 indeed, but if the message for SA1209 is suggesting where to place the alias using directive, then I think that it should at least suggest a location that is compatible with SA1211.

baks: makes sense :+1:

baks: I wonder about yet another case : 
'''csharp

sharwell: > Using alias directive for Numerics = System.Numerics must appear after directive for IO = System.IO

I think this will hurt readability. I would prefer to see the aliases alone:

> Using alias directive for 'Numerics' must appear after directive for 'IO'

sharwell: > Using alias directive for System.IO must appear after directive for using static System.Math

`using static` directives are last, so this message would never occur.

vweijsters:
I think we might be better off if we just change the message of SA1209 to something that's a bit more like the message that StyleCop reports:
> Using alias directives must be placed after all using namespace directives.

Get the right suggestion on where to place it correctly is difficult (as can been seen in the comments above) and it offers limited value. Most developers will instantly see the correct location anyway.

@sharwell How do you think about giving SA1209 a generic message without any target namespace names in it?

sharwell:
@vweijsters That would be fine with me :)

baks:
So, I changed message to be general message and deleted code associated with resolving correct namespace names

vweijsters:
:+1: Looks good to me

Issue #789, opened 2015-05-06
SA1600-1602: support SC rules for whether documentation is required for private and internal symbols


oatkins:
SC has a setting that relaxes these rules to apply to public types and members only. It's "documented" in [this bug report][1]. It's a really useful feature, because it helps ensure that the public API of a library is properly documented.

I'm not sure how this project is handling external settings files, but it would be really good to have something that maintains this feature.

One option that would avoid depending on external settings would be to have three analyzers for each of these rules—separating out private, internal and public symbols. e.g. SA1600_public, SA1600_internal and SA1600_private. Enabling all three would give the strictest behaviour.

[1]: http://stylecop.codeplex.com/workitem/6772

Issue #764, opened 2015-05-02
Code fix: SA1600 (ElementsMustBeDocumented)


sharwell:
Implement a code fix for #120.

In general the code fix would create the necessary XML elements in a new documentation comment, but not fill in the details. However, for certain cases we can fill in these details. For example:
- The `<summary>` element for constructors and finalizers can be created according to SA1642 and SA1643.
- If the member needing documentation overrides or implements a member from a base class or interface, the content can be simplified to just `<inheritdoc/>`.
- If the member returns a `Task`, the `<returns>` element should read as follows (note we may use different wording):
  
  ```
  A <see cre="Task"/> representing the result of the asynchronous operation.
  ```

oatkins:
Rusult!

sharwell:
@oatkins :expressionless:

jimmymain:
Hi,
I think sometimes the details can be filled in.
In the case of a properly named property it would be helpful.

```
public string VogonConstructorFleets { get; set; }
```

Could quite easily be documented as:
```
/// <summary>
/// Gets or sets the vogon constructor fleets
/// </summary>
```

I think there is value in this, particularly in the case of correctly named properties.
It's a real time saver.

thoughts?

sharwell:
@jimmymain It was my impression that several people working on this find automatically-generated documentation wrong (or useless) more frequently than it is correct (or useful). Right now I can't find the conversation I was thinking of. While they obviously have good intentions, I'm definitely not a fan of the behavior of tools like [GhostDoc](https://visualstudiogallery.msdn.microsoft.com/46A20578-F0D5-4B1E-B55D-F001A6345748).

That said, please keep the following in mind:

1. Feel free to create a proposal describing specific behavior you had in mind.
2. You can always implement a code fix for SA1600 separately from this project, and install it in the same project(s) where you use StyleCop Analyzers. For example, the InheritdocCodeFixProvider in this project handles CS1591 warnings, which are not produced by this project. Anyone who finds it useful can install and use it.

  :memo: While I would likely vote against adding this behavior directly to the StyleCopAnalyzers repository, I would not have any objection to creating an additional repository in the DotNetAnalyzers organization aiming to provide more comprehensive documentation generation features.

jimmymain:
On the whole, I agree, though my feeling is that my case for properties is pretty strong.
I will implement a fix separately and post a link to it.

pdelvo:
[...]

Issue #717, opened 2015-04-26
New rule proposal: Multi-line statements must be surrounded by braces


sharwell:
This rule is an intentional subset of the behavior of SA1503 (CurlyBracketsMustNotBeOmitted) targeting the behavior used by me in my own projects and now adopted by CoreFX.

**Category:** Layout
**Rule:** Multi-line statements must be surrounded by braces (improvements to this wording welcome)
**Description:** If the child of a compound statement does not fit on a single line, then curly braces must be used for the compound statement.

**Example:**

The following code does not trigger this rule:

```csharp
if (x == null)
    throw new ArgumentNullException(nameof(x), "some message");
```

The following code **does** trigger this rule:

```csharp
// The body of the `if` statement omits braces *and* spans multiple lines, thus
// triggering this diagnostic.
if (x == null)
    throw new ArgumentNullException(
        nameof(x),
        "some message);
```

To fix a violation of this rule, either insert curly braces or place the entire statement on a single line. The automatic code fix for this violation would always choose to insert braces.

AArnott:
:+1:

vweijsters:
:+1:

Przemyslaw-W:
+1

2015-05-20 17:32 GMT+02:00 Vincent Weijsters <notifications@github.com>:

Issue #716, opened 2015-04-26
New rule proposal: Use braces consistently


sharwell:
This rule is an intentional subset of the behavior of SA1503 (CurlyBracketsMustNotBeOmitted) targeting the behavior used by me in my own projects and now adopted by CoreFX.

**Category:** Layout
**Rule:** Use braces consistently
**Description:** For a sequence of `if`/`else if`/`else` control flow statements, if any element of the sequence includes curly braces for the block, then *all* elements of the sequence should include curly braces. Note that `else if` is a special case here - if `else` is immediately followed by `if`, then the `else` does not need braces.

vweijsters:
:+1: its nice to give people a bit more flexibility with regards to curly braces.

Przemyslaw-W:
+1

2015-05-20 17:32 GMT+02:00 Vincent Weijsters <notifications@github.com>:

PR #832, opened 2015-05-13
Split SA1600, SA1601 and SA1602 into three rules each

### LINKED ISSUE #829: SA1600-1602: Diagnostic should not be reported for elements inside a private class
As implemented at the moment, rules SA1600-1602 require documentation for everything except private elements. (My report #789 relates to the fact that this was actually configurable in SC, and the default //was// to require documentation even on private elements.)

The current code doesn't seem to properly account for how modifiers on parent elements determine the effective visibility of a child element. For example

    /// <summary>
    /// Some summary.
    /// </summary>
    public class Outer
    {
        private class inner
        {
            // A diagnostic is reported here.
            // It shouldn't be!
            public string SomeProperty { get; set; }
        }
    }

I'd also be interested in some feedback on [these new tests][1]. I tried to work with the existing tests, but found them a bit hard going. The linked branch contains an experimental new set of tests in `SA1600UnitTestsNew` that are parameter and use a helper class to generate sample code. They include the failing case shown above.

[1]: https://github.com/oatkins/StyleCopAnalyzers/blob/SA1600-1603-tests/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTestsNew.cs

oatkins:
- For public, internal and private members. Added code to resolve visibility correctly in scope of parent type.
- Updated tests.

This is a first pass at trying to resolve #829 and #789. I'm sure this will generate some discussion, but hopefully it's a step in the right direction.

SA1600 has a completely new set of tests, and uses a crazy-new-test-code-generator.
There are, of necessity, new diagnostic IDs. These have the suffices "In" and "Pr".

### DIFF
--- a/SA16X0.txt
+++ b/SA16X0.txt
@@ -1,0 +1,36 @@
+# SA16X0
+
+<table>
+<tr>
+  <td>TypeName</td>
+  <td>SA16X0NonPrivateElementsMustBeDocumented</td>
+</tr>
+<tr>
+  <td>CheckId</td>
+  <td>SA16X0</td>
+</tr>
+<tr>
+  <td>Category</td>
+  <td>Documentation Rules</td>
+</tr>
+</table>
+
+## Cause
+
+A non-private, non-public C# code element is missing a documentation header.
+
+## Rule description
+
+C# syntax provides a mechanism for inserting documentation for classes and elements directly into the code, through the use of Xml documentation headers.
+For an introduction to these headers and a description of the header syntax, see the following article: http://msdn.microsoft.com/en-us/magazine/cc302121.aspx.
+
+A violation of this rule occurs if an element is completely missing a documentation header, or if the header is empty. In C# the following types of elements
+can have documentation headers: classes, constructors, delegates, enums, events, finalizers, indexers, interfaces, methods, properties, and structs.
+
+No violation of this rule occurs for public elements; these are covered by [CS1591](https://msdn.microsoft.com/en-us/library/zk18c1w9.aspx).
+This rule specializes [SA1600](http://www.stylecop.com/docs/SA1600.html), [SA1601](http://www.stylecop.com/docs/SA1601.html) and [SA1602](http://www.stylecop.com/docs/SA1602.html)
+for non-private, non-public elements only.
+
+## How to fix violations
+
+To fix a violation of this rule, add or fill in the documentation header for this element.
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/DocumentationOptions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/DocumentationOptions.cs
@@ -1,0 +1,15 @@
+﻿namespace StyleCop.Analyzers.Test.DocumentationRules
+{
+    public enum DocumentationOptions
+    {
+        /// <summary>
+        /// Indicates that unit test code should contain sample documentation.
+        /// </summary>
+        WriteSampleDocumentation,
+
+        /// <summary>
+        /// Indicates that unit test code should not contain sample documentation.
+        /// </summary>
+        OmitSampleDocumentation
+    }
+}
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/DocumentationRuleTestSampleCodeBuilder.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/DocumentationRuleTestSampleCodeBuilder.cs
@@ -130,7 +130,7 @@
             this.Write(";\r\n");
         }
 
-        public void WriteFinalizer(string name, DocumentationOptions documentationOptions, ExpectedResult expectError)
+        public void WriteDestructor(string name, DocumentationOptions documentationOptions, ExpectedResult expectError)
         {
             switch (documentationOptions)
             {
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/DocumentationRuleTestSampleCodeBuilder.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/DocumentationRuleTestSampleCodeBuilder.cs
@@ -1,0 +1,306 @@
+﻿namespace StyleCop.Analyzers.Test.DocumentationRules
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Linq;
+
+    public sealed class DocumentationRuleTestSampleCodeBuilder : SampleCodeBuilder
+    {
+        private readonly ImmutableArray<string> diagnosticIds;
+
+        public DocumentationRuleTestSampleCodeBuilder()
+            : this(null)
+        {
+        }
+
+        public DocumentationRuleTestSampleCodeBuilder(IEnumerable<string> diagnosticIds)
+        {
+            this.diagnosticIds = diagnosticIds == null
+                ? ImmutableArray<string>.Empty
+                : diagnosticIds.Where(x => !string.IsNullOrEmpty(x)).Distinct(StringComparer.Ordinal).ToImmutableArray();
+        }
+
+        public void WriteClassStart(string name, DocumentationOptions documentationOptions, ExpectedResult expectError, params string[] modifiers)
+        {
+            this.WriteClassStart(name, documentationOptions, expectError, modifiers, null);
+        }
+
+        public void WriteClassStart(string name, DocumentationOptions documentationOptions, ExpectedResult expectError, string[] modifiers, string[] inherits)
+        {
+            switch (documentationOptions)
+            {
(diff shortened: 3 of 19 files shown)


oatkins:
I'd be interested to get some feedback on this. I know the test code is a bit weird and wonderful (OK: maybe just weird), but I think the logic changes to the rules themselves make the doc rules more useful—especially the changes to take into account the effective visibility of elements based on the visibility of their parent types.

vweijsters:
I just did a quick review of the code and most of it looks good.

I do have an issue with the the way the tests are set up.
In my opinion the new setup has a serious issue with readability. I find it quite frustrating that I have to analyze each test case in detail to see what the intended test case is. 
The new test cases will probably take slightly less time to code, but reviewing them and determining if they are correct has become a much more difficult job.

vweijsters:
Besides that I also think that having a StyleCop (like) config file would be a much better option than having separate rules for public / internal / private.
I appreciate the effort to get things working right now, but I think in the long run the project would be better off with a config file mechanism.

oatkins:
@vweijsters - Thanks for taking a look. I'm happy to try to re-work the tests.

The comment about the settings file makes sense. I went this way after reading comments by @sharwell on #422. However, I can keep this in a private build for my own use (since the current versions of these rules throw up a lot of false positives for us) up until settings files have been implemented. Then it should be fairly easy to integrate those changes into this branch.

sharwell:
@oatkins I ran into a similar problem with SA1503 - I wanted some but not all of its behavior. We resolved this without configuration by proposing (#716, #717) and finally implementing (#861) two brand new StyleCop rules.

Suppose now that you want to split rule SA1600 into *two* categories (for simplifying the example): publicly-accessible members, and non-publicly-accessible ("internal") members. In this case you could propose a rule for either of these two. Ideally you would propose a rule for publicly accessible members, which means two things:

1. The new rule would be enabled by default. Users who suppressed SA1600 due to its scope would see new warnings, but only on publicly-accessible members. They would then have the option of disabling the new rule as well.
2. If the new rule is disabled, but SA1600 is enabled, the behavior of SA1600 wouldn't change at all.

If you break SA1600 into more than two categories, you will need to be very specific in your descriptions with respect to nested types and members.

oatkins:
@sharwell Thanks for the tips. I'm still not 100% clear on where to go with this, so I hope you'll excuse a couple more questions.

The StyleCop behaviour for SA160x is, by default (i.e. no settings file, no suppressions), to require comments on *all* elements, whether public, internal or private. This can be modified (relaxed) by two properties in the settings file: one to disable the requirement for documentation on private elements, and the other for internal elements.

Here are two alternatives that come to mind for how this could be handled with new rules (taking SA1600 as the example case):

# Option 1

- SA1600 functions as per the SC defaults, requiring documentation on all elements.
- Three new rules, disabled by default, break out the results at the three accessibility levels.
- It would never make sense to enable SA1600 *and* one or more of the new rules.
- When configuration is supported later, the legacy SC configuration properties could be supported on SA1600 but ignored by the new rules.

# Option 2

- SA1600 functions as if a settings file had enabled it for public members only.
- Two new rules add support for requiring comments on internal and private members.
- The legacy SC behaviour would be obtained by enabling all the rules.
- When configuration is supported later, the two new rules could respect the configuration properties too.

To my mind the second option is more straightforward because enabling all the rules wouldn't give duplicate diagnostic messages. However, the first option mirrors the legacy SC behaviour more precisely.

What do you think?

vweijsters:
How about option 3: Give me a couple of days to get the configuration stuff working and go from there?

sharwell:
I would pick option 4:

* SA1600 reports anything *not* reported by another rule. Similar to how SA1503 behaves today with respect to SA1519 and SA1520.
* Two new rules, one for publicly-accessible signatures, and one for non-private but also non-public signatures.
* All rules can be enabled by default without conflicts.

vweijsters:
I'm not so sure about option 4 yet. There will be a large number of additional rules needed to support all scenarios, especially if we also want take into account the 'ignore for fields' setting.

The list below would be all rules that are needed:

SA1600, SA1600 public only, SA1600 internal only, SA1600 no fields, SA1600 public only - no fields, SA1600 internal only - no fields
SA1601, SA1601 public only, SA1601 internal only, SA1601 no fields, SA1601 public only - no fields, SA1601 internal only - no fields
SA1602, SA1602 public only, SA1602 internal only, SA1602 no fields, SA1602 public only - no fields, SA1602 internal only - no fields

(`protected` would be considered the same as `public`, as would `public internal`)

I would very much prefer to handle that with a settings file, instead of having to fill-in the correct puzzle in the rule set file. 

The use case here is having settings that are the same for a number of different analyzers. We could solve that by creating individual analyzers to cover each setting, but I think we should not. I like the pattern that was implemented for SA1503, but this is a different case and therefore it warrants a different solution.

oatkins:
@sharwell I agree that there are definite benefits to avoiding reliance on config files where possible—though this probably won't be possible for, say, [SA1636](http://stylecop.soyuz5.com/SA1636.html).

@vweijsters I completely agree with the comment,

> `protected` would be considered the same as `public`, as would `public internal`.

though I suspect you intended to write `protected internal` in that last clause. This makes perfect sense from a documentation perspective, and makes it much easier to reason on the rules.

I don't see why fields should be excluded. Some APIs make extensive use of fields, e.g. WPF. StyleCop doesn't make such an exclusion.

@sharwell  I can't quite figure out the rationale behind "option 4." It seems to be the same as 2, but with SA1600 and one of the new rule proposals transposed.

It'll certainly be helpful for the rules to be mutually exclusive, but by the logic that, "SA1600 reports anything *not* reported by another rule" (which is logical) it seems as though SA1600 would end up being for private elements only. Since this is perhaps the *least* useful case (and least frequently enabled in legacy SC) it seems rather counter-intuitive.

Here is my (over-)simplified understanding of your two proposed *new* rules (again, treating `protected` and `protected internal` as `public`):

- publicly-accessible elements, and
- internal elements (i.e. non-public, non-private).

That leaves only "private" for the existing rule.

What am I missing? btw Sorry to be a pain with this. Just want to make sure I fully understand before making the changes.

vweijsters:
@oatkins The option to exclude fields is present in StyleCop and definitely handy if you want private / internal stuff documented, except for the fields.

vweijsters:
@oatkins Option 4 is about SA1600 reporting everything, unless the two new rules are enabled. In that case SA1600 will only report for the private stuff. It is a mechanism that we introduced recently to suppress duplicate messages. Have a look at the implementation of SA1503, SA1519 and SA1520 for details.

sharwell:
@vweijsters I disagree with your breakdown. In particular, we could get away with only the following additional rules:

1. Publicly-visible elements must be documented. This single rule covers the relevant subset of SA1600, SA1601, and SA1602. For discussion below, call this SA16X1.
2. Non-private elements must be documented. This single rule covers the relevant subset of SA1600, SA1601, and SA1602. For discussion below, call this SA16X2.

I do not believe fields need to receive special consideration outside of the above rules. If you don't want to document fields, enable SA16X1, SA16X2, and SA1401 (FieldsMustBePrivate). If you want to document public members only, enable SA16X1 and disable SA1600, SA1601, SA1602, and SA16X2.

As it turns out, SA16X1 as described above already exists - it's CS1591. This means only one additional rule is needed to enable all scenarios.

vweijsters:
@sharwell Good analysis :smile: 

The only minor case that is missing is when private fields should not be documented, but private methods / properties should be. 

Another minor issue might be that CS1591 will only be enabled when the /doc compiler option is specified. It might be handy to still implement SA16X1 to work around that.

oatkins:
Great. The bit I hadn't grasped is the way SA160x changes its behaviour depending on whether the other rules are enabled. Ingenious. I will take a look at SA1503 before coding anything.

sharwell:
@vweijsters It's important to always enable the documentation compiler, otherwise all semantic information in documentation comments is ignored and refactoring operations in the IDE will do the wrong thing. This Best Practice is outside the scope of StyleCopAnalyzers. See dotnet/roslyn#2773 for more information.

sharwell:
> The only minor case that is missing is when private fields should not be documented, but private methods / properties should be.

The reason I think we can ignore that now is this generally applies to cases where a private field simply backs a property. With auto-properties, and now read-only auto-properties, I think the demand for such a configuration option is rapidly dropping.

oatkins:
Hopefully this represents the latest suggestions on how these rules should function.

I've not done the work to change the tests so they don't depend on my little test code factory class yet. Also, the rules now change their behaviour depending on whether *other* rules are enabled. I can't see how to write tests that take that into account, so for now the tests only exercise the behaviour of the analyzers when all rules are enabled.

oatkins:
If we rely on CS1591, does this mean that we might end up with a code fix (#764) that works for private elements but not public ones? That would be a bit weird.

sharwell:
Code fixes can be applied to any warning identifier, including ones provided by the compiler. :smile: 

Doesn't mean we won't provide our own ID for it. It's just one option.

sharwell:
Superseded by #1278.

PR #839, opened 2015-05-15
Implemented file header diagnostics + tests

### LINKED ISSUE #153: SA1633: FileMustHaveHeader
http://www.stylecop.com/docs/SA1633.html

### LINKED ISSUE #154: SA1634: FileHeaderMustShowCopyright
http://www.stylecop.com/docs/SA1634.html

### LINKED ISSUE #155: SA1635: FileHeaderMustHaveCopyrightText
http://www.stylecop.com/docs/SA1635.html

### LINKED ISSUE #157: SA1637: FileHeaderMustContainFileName
http://www.stylecop.com/docs/SA1637.html

### LINKED ISSUE #158: SA1638: FileHeaderFileNameDocumentationMustMatchFileName
http://www.stylecop.com/docs/SA1638.html

### LINKED ISSUE #159: SA1639: FileHeaderMustHaveSummary
http://www.stylecop.com/docs/SA1639.html

### LINKED ISSUE #160: SA1640: FileHeaderMustHaveValidCompanyText
http://www.stylecop.com/docs/SA1640.html

vweijsters:
Fixes #153
Fixes #154 
Fixes #155 
Fixes #157 
Fixes #158 
Fixes #159 
Fixes #160 

I implemented the entire bunch in one go for efficiency reasons.

For SA1639 (which is disabled by default according to StyleCop documentation) I added AnalyzerConstants.DisabledByDefault and I modified DiagnosticVerifier.Helper to make sure that disabled by default analyzers can be unit tested.

I skipped SA1636 because it needs configuration files to work properly, which is still unclear to me on how we will solve this.

There is probably a possibility to add some form of caching to the FileHeaderHelpers, inline with the caching currently performed for the auto generated determination, but I'm not quite sure what the impact of that would be and if it is needed, so I would prefer to discuss this first before adding it.

I skipped the code fixes, because I think we can create a single code fix for headers, as soon as we've solved the configuration issue for SA1636. Lets discuss this as well.


pdelvo:
Wow very cool! Im going to have a look at this when Im home

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1638FileHeaderFileNameDocumentationMustMatchFileName.cs
```diff
@@ -40,30 +44,58 @@ public class SA1638FileHeaderFileNameDocumentationMustMatchFileName : Diagnostic
...
+            var fileAttribute = copyrightElement.Attribute("file");
+            if (fileAttribute == null)
+            {
+                // this will be handled by SA1637
+                return;
+            }
+
+            var fileName = Path.GetFileName(context.Tree.FilePath);
+
+            if (!fileAttribute.Value.Equals(fileName, StringComparison.OrdinalIgnoreCase))
```
pdelvo: Do we really want to ignore the case here? If the code is used on an unix system case matters.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeader.cs
```diff
@@ -0,0 +1,125 @@
...
+        internal bool IsMalformed { get; private set; }
+
+        /// <summary>
+        /// Gets a XML element from the file header with the given tag name.
+        /// </summary>
+        /// <param name="tagName">The tag name for the node.</param>
+        /// <returns>The requested node, or null if the node could not be found.</returns>
+        internal XElement GetElement(string tagName)
+        {
+            return this.headerXml.Descendants().FirstOrDefault(e => e.Name.LocalName.Equals(tagName, StringComparison.Ordinal));
```
pdelvo: What does stylecop do if there are multiple copyright/summary elements?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeader.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <param name="syntaxTree">The syntrax tree to use for generating the location.</param>
+        /// <returns>The location representing the start of the file header.</returns>
+        internal Location GetLocation(SyntaxTree syntaxTree)
+        {
+            if (this.IsMissing || this.IsMalformed)
+            {
+                return Location.Create(syntaxTree, new TextSpan(0, 0));
+            }
+
+            return Location.Create(syntaxTree, TextSpan.FromBounds(this.fileHeaderStart, Math.Min(this.fileHeaderStart + 2, this.fileHeaderEnd)));
```
pdelvo: Why fileHeaderStart + 2? is it possible to have a file header with a length less then two chars which is not malformed?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeader.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        /// <summary>
+        /// Gets the location representing the position of the given element in the source file.
+        /// </summary>
+        /// <param name="syntaxTree">The syntrax tree to use for generating the location.</param>
+        /// <param name="element">The XML element to get the location of.</param>
+        /// <returns>The location representing the position of the given element in the source file.</returns>
+        internal Location GetElementLocation(SyntaxTree syntaxTree, XElement element)
```
pdelvo: This method can be buggy. Example:
```csharp
// <summary>
// <![CDATA[<copyright>]]>
// </summary>
// <copyright>
// </copyright>
// ...
```

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeaderHelpers.cs
```diff
@@ -0,0 +1,96 @@
...
+                    case SyntaxKind.WhitespaceTrivia:
+                        endOfLineCount = 0;
+                        break;
+                    case SyntaxKind.SingleLineCommentTrivia:
+                        endOfLineCount = 0;
+
+                        var commentString = trivia.ToFullString();
+
+                        // ignore borders
+                        if (commentString.StartsWith("//-", StringComparison.OrdinalIgnoreCase))
```
pdelvo: Does StyleCop support bords in the middle of a file header?

pdelvo:
Does StyleCop support multiline comments for file headers?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1638FileHeaderFileNameDocumentationMustMatchFileName.cs
vweijsters: No, we probably don't. I'll change the code.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeader.cs
vweijsters: StyleCop only processes the first element, same as the behavior in this analyzer.

We should probably add another rule if we want to catch multiple. @pdelvo @sharwell what do you think?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeader.cs
vweijsters: The fileHeaderStart + 2 is intended to limit the squiggles to the first `//` pair of the entire file header. 

The minimal non-malformed file header would be `// <a/>`, so that's definitely more than 2 characters, I'll remove the `Math.Min` stuff.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeaderHelpers.cs
vweijsters: Yes, it does.

vweijsters:
StyleCop does not support multiline comments for file headers.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/FileHeader.cs
vweijsters: I know, but it will be really hard to get the position right in any other way (other than writing a full blown XML parser). I've tried `IXmlLineInfo`, but I couldn't find any functionality in roslyn to map a line number, character number back to `Location` object that is still linked to the syntax tree.

vweijsters:
Updated to use #841

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/FileHeaderTestBase.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// </summary>
+    public abstract class FileHeaderTestBase : DiagnosticVerifier
+    {
+        /// <summary>
+        /// Gets the diagnostics that will be checked in the <see cref="TestEmptySourceAsync"/> test case.
+        /// </summary>
+        /// <value>
+        /// The diagnostics that should occur when the test case is executed.
+        /// </value>
+        protected virtual DiagnosticResult[] EmptySourceDiagnostics
```
sharwell: :exclamation: An empty source file should not produce diagnostics. This may deviate from the behavior of the reference StyleCop implementation, but it does not make sense to report diagnostics for an empty file.

vweijsters:
Updated to no longer produce a diagnostic for an empty file.

Issue #678, opened 2015-04-22
Code fix: SA1615 (ElementReturnValueMustBeDocumented)


sharwell:
Implement a code fix for #135.

PR #843, opened 2015-05-16
Fix StyleCop warnings

AArnott:
I remove verification of results being defined because that irritatingly requires that you document async Task methods' return values. And for a test project that hardly seems worthwhile.


sharwell:
:exclamation: The return value for an asynchronous test method is documented as follows:

```cs
```

AArnott:
It wasn't a matter of me not being able to document it. It's simply that I see no value in it. And in general the value of `<returns>` is highly questionable IMO in a unit test project.
But if you'd rather I fix it by documenting it, sure, I can do that.

sharwell:
I'm working on #678, which will have special handling for this case. :+1:

AArnott:
I fixed the commit message, btw.

sharwell:
:exclamation: I would prefer to leave that diagnostic enabled, and let #846 take care of the tedious cases.

AArnott:
Done

sharwell:
Rebased to e0583e1dacd2e2542a4f0eb8dba63791a8d4c161 and merged in 98aec9dfee4fdc73f3239a1d645436413235c882.

PR #844, opened 2015-05-16
Implement alternative rules SX1309 and SX1309S

### LINKED ISSUE #837: SX1309: FieldNamesMustBeginWithUnderscore
This is the standard alternative rule to SA1309 (FieldNamesMustNotBeginWithUnderscore).

A violation of this rule occurs when a field name does not begin with an underscore. This rule only examines fields which are *all* of the following:

* `private`
* Instance (not `static`)
* Not `const`

### LINKED ISSUE #838: SX1309S: StaticFieldNamesMustBeginWithUnderscore
This rule expands on the coverage of SX1309 (FieldNamesMustBeginWithUnderscore) (#837) to cover static fields in addition to instance fields.

A violation of this rule occurs when a field name does not begin with an underscore. This rule only examines fields which are *all* of the following:

* `private`
* `static`
* Not `const`
* Not `readonly`

sharwell:
Fixes #837
Fixes #838

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SX1309SUnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SX1309SUnitTests.cs
@@ -1,0 +1,172 @@
+﻿namespace StyleCop.Analyzers.Test.NamingRules
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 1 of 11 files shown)


vweijsters:
I think it would be a good idea to create a test base class, as both test classes are large duplicates of each other.

vweijsters:
Both analyzers are very similar as well and probably could benefit from a base class as well, most likely even combined with SA1309.

vweijsters:
:+1: Besides the two minor remarks above, it looks very good.

sharwell:
Sharing a test base class doesn't share nearly as much as was hoped, because the `[MemberData]` attribute only works with static members.

vweijsters:
That will make things very difficult indeed and not worth the effort for just two classes.

PR #846, opened 2015-05-16
Implement a code fix for SA1615 and SA1616

### LINKED ISSUE #678: Code fix: SA1615 (ElementReturnValueMustBeDocumented)

sharwell:
@AArnott you should appreciate this one :laughing: 

Fixes #678

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1615UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1615UnitTests.cs
@@ -29,8 +29,10 @@
             {
                 yield return new[] { "    public          Task      MethodAsync(string foo, string bar) { return null; }" };
                 yield return new[] { "    public          Task<int> MethodAsync(string foo, string bar) { return null; }" };
+                yield return new[] { "    public          TASK      MethodAsync(string foo, string bar) { return null; }" };
                 yield return new[] { "    public delegate Task      MethodAsync(string foo, string bar);" };
                 yield return new[] { "    public delegate Task<int> MethodAsync(string foo, string bar);" };
+                yield return new[] { "    public delegate TASK      MethodAsync(string foo, string bar);" };
             }
         }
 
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1615UnitTests.cs
```diff
@@ -106,14 +133,106 @@ public class ClassName
...
+using System.Threading.Tasks;
+/// <summary>
+/// Foo
+/// </summary>
+public class ClassName
+{
+    /// <summary>
+    /// Foo
+    /// </summary>
+    /// <returns><placeholder>A <see cref=""Task""/> representing the asynchronous operation.</placeholder></returns>
```
AArnott: What does `<placeholder>` signify?

sharwell: It's a [SHFB MAML element](http://ewsoftware.github.io/MAMLGuide/html/d995d093-ce18-4a7e-bd03-82c6c0b4fd72.htm) that also works in XML documentation comments. In combination with #536, it plays a vital role in allowing us to generate certain documentation comment (e.g. here and in #606) without compromising the quality of project documentation.

AArnott: I don't understand yet (sorry). I read the link and it's very abstract. Can you give me an example of how this tag might behave or be useful?
What do you mean it allows you to generate a documentation comment without compromising the quality of the project documentation?

sharwell: The code fix for #606 generates content for a `<value>` documentation element according to the current content of the `<summary>` element. It also supports the `BatchFixAllProvider` to quickly create `<value>` documentation for all properties that already have `<summary>` documentation. However, this documentation is only a guess of what the user intended, so we wrap this content in a `<placeholder>` element which indicates that the content still needs to be reviewed for accuracy. After reviewing the content, the `<placeholder>` element is removed leaving just its content (the code fix in #604 can do this automatically).

The `<placeholder>` element is not intended to affect the rendered documentation. Its entire purpose is to mark sections of documentation as "needing review", so we use it any time content is generated for a comment by an analyzer.

AArnott: Ah, I see thanks.
Did you really mean *anytime* doc is generated by an analyzer? That seems like it would apply to the boilerplate doc produced for the `<summary>` tag of a constructor, and that bolierplate is adequate IMO as-is, and adding `<placeholder>` around it every time the code fixer spits it out would make it more work for the user every time it was produced.

sharwell: No, in that case (constructors and finalizers) the rule actually defines the text, so when we generate it we are certain it's correct.

pdelvo:
It might be good to include a test that verifies, that the code fix still works if Task is not in scope.

sharwell:
@pdelvo Good idea. I'll move this back to Alpha 8 so I can get that in.

pdelvo:
A test that verifies that the check if the return type is "Task" is semantically would be usefull, so if I have a using Foo = Task; it should still work

pdelvo:
Otherwise this looks good to me

sharwell:
@pdelvo I added some tests with a using alias directive.

pdelvo:
:shipit:

PR #847, opened 2015-05-17
Improve version stamps of builds

### LINKED ISSUE #793: Local StyleCop builds should be identical to public builds
StyleCop.Analyzers on dev boxes builds to 1.0.0-dev while the publicly released builds take the form of 1.0.0-alpha000. This makes it difficult or impossible to make a fix locally and publish the package to some internal package store, repeatedly, and have the consuming projects track which private build they're using at the time.

One pattern that we use in some places internally at Microsoft that I think would be beneficial here is for the package versions to take this form:

1.0.0-pre-ga1f23cd3

The idea here is that the version spec includes the git commit id (after the -g for 'git') allowing folks to immediately identify the version of source code in git that went behind a build (even without producing tags for every single one) and most importantly, allows a local build to exactly reproduce what the public build will be if the commit is the same. 

We could of course make this more friendly to nuget upgrades by adding the alpha000 back:

1.0.0-alpha008-ga1f23cd3

But we'd need some file in SCC to name the alpha000 number being built so that anyone who builds the commit can reproduce the package without any special global properties, etc. When I've done this myself, my pattern is that the version.txt file will have 'alpha008' in it as soon as alpha007 ships. So that all subsequent builds claim to be alpha008 to match the next public release.

What do you think? If you agree, I can prepare a pull request to add this.

AArnott:
With this change, it is no longer necessary to make commits that change version data just to prepare for an official release (like alpha006) and revert it back immediately after so that local builds are versioned as 1.0.0-dev.

With the old scheme, since everyone's build was always 1.0.0-dev it made it impossible to consume the NuPkg produced by a build anywhere because 1.0.0-dev is a constant that doesn't actually point at any specific version. Now, anyone's non-official build will be unique with the commit ID where it was built, such as:

    1.0.0-alpha-g9a7eb6c819

In this way, anyone with a fork of the repo can build their own package, upload it to their private nuget store (myget.org, a network share, etc.) and consume it immediately.

When shipping officlal releases, simply build with the /p:UseNonZeroBuildNumber=true switch to produce a build number like this:

    1.0.15136-alpha

Note that this no longer uses the "-alpha007" syntax, with its fixed number of leading zeros. Instead, it just uses the PATCH component of the version as the incrementing integer for proper semver-sorting and easier reading that this is an alpha.
This is a safe upgrade from what we were doing because this new format is seen as 'newer' than the old one, so upgraders will get the new versions as expected.

Fixes #793


sharwell:
If you look at the GitHub releases page, you'll see that at the bottom of the information for the Alpha 6 build is a zip file (or tar.gz file). GitHub prepares these files automatically based on the tag I created. In the current implementation, if you download one of those files and extract it, and then run **build\build.ps1**, it will build a NuGet package matching the official distribution from those tags (it's not a binary match, but only because the C# compiler is not deterministic). This pull request appears to eliminate our ability to provide that functionality for two reasons:

1. The build version is now tied to `DateTime.Now`
2. The build version is tied to the existence of Git repository information, which is not contained in those archives.

AArnott:
Yes, those are good points. On the other hand, reproducing builds when you are using git (instead of the tarball) is easier now. Do you know how many people use the tarball to produce their own builds vs. fork the git repo and produce them there instead? My change is catering to the git forking crowd, to be sure.

If you'd like to keep the tarball approach working, do you have any proposal for a compromise so that we can improve the git fork story without degrading the tarball so much? Getting rid of the jdate and instead using the PATCH component in the version.txt file is certainly one possibility. That would make it readily reproduce official builds when you download it, minus the git commit ID part of the assembly informational version. But that won't cause any issue with them switching to their own build over yours since the information version doesn't fit into binding rules.

AArnott:
[See here](https://www.myget.org/feed/aarnott/package/nuget/StyleCop.Analyzers) for an example of how I'm leveraging this feature. Note that the version of my package will never conflict with yours, allowing myself to easily jump between my builds and your official builds as  necessary.

sharwell:
I've spent a great deal of time over the past few years thinking about challenges surrounding versioning and the community development efforts. My focus, and the reference for my current practices is:

1. **Usable releases:** Ensuring it is easy for new users to use a reference release of the project (i.e. a tagged commit). This is reflected in at least the following:
  * The [release notes](), which aren't perfect but are arguably above average
  * The fact that you can build a reference release from the **.zip** file attached to the release
  * Uploading reference releases to NuGet, along with sources and symbols to Symbol Source
1. **Dependable stable releases:** Ensuring that users with strong-named assemblies referencing a "stable" build (i.e. a tagged commit without a suffix) have *absolute dependability* at runtime, even if another assembly in the same process is referencing a pre-release build of the library. This is less of a concern for this project, but you'll note that the builds do follow most of the versioning practices described here, including the use of a different strong name key for stable releases.
1. **Open contributor policy:** Ensuring it is easy for new users to start contributing to the project. This affects technical aspects of the project, such as [project setup]() and strong name keys for clone-and-go support. It also affects non-technical aspects such as [CONTRIBUTING.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/CONTRIBUTING.md).

I believe that the current versioning system employed by the project solves these three problems well. Furthermore, I want to focus on providing timely releases to address the majority of concerns surrounding "access to new features".

For now, I'm closing this pull request as **wontfix**. This does not mean I won't be revisiting this issue in the future. The following are the reasons I'm closing it:

1. It is not clear to me that the change is beneficial when measured by the goals outlined above. Even if the problems mentioned in previous comments are addressed (e.g. the use of `DateTime.Now` could easily be changed to use the commit time for build stability), the goals of the versioning change differ from the goals which led to the current versioning strategy used by the project.
2. For users who currently need to maintain forks of this project due to technical limitations (e.g. when working with a different build of Roslyn, the analyzers may need to be rebuilt against that release), this pull request does not change that requirement. Users will still need to fork the project. If users already need to maintain a separate fork, then it is easy to incorporate this change into the fork.

Issue #821, opened 2015-05-11
SA1028 (CodeMustNotContainTrailingWhitespace) does not properly handle multi-line documentation comments


sharwell:
Currently the SA1028 analyzer does not report warnings inside a multi-line documentation comment, such as the following:

```csharp
/**
 * <summary>  
 * Summary.    
 * </summary>  
 */
```

In addition, the code fix for SA1028 treats `SyntaxKind.MultiLineDocumentationCommentTrivia` the same way it treats `SyntaxKind.MultiLineCommentTrivia`. This may be incorrect since these comments are documentation comments with internal structure similar to `SyntaxKind.SingleLineDocumentationCommentTrivia`.

vweijsters:
Grabbing this

PR #852, opened 2015-05-17
Implement code fix for trailing whitespace after xml doc comments

### LINKED ISSUE #795: SA1028 (CodeMustNotContainTrailingWhitespace) code fix does not support documentation comments
This is a problem for the following reasons:

* Documentation comments are already analyzed, so warnings will be reported.
* The default smart indent implementation in Visual Studio for documentation comments includes trailing whitespace on new lines, as opposed to most other cases where virtual space is used to avoid these spaces.

AArnott:
Fix #795

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
@@ -8,6 +8,7 @@
     using Microsoft.CodeAnalysis.CodeActions;
     using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Text;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


sharwell:
Can you add a test which covers a documentation comment in the following form?

```csharp
/**
 * <summary>  
 * Some description    
 * </summary>
 */
class Foo { }
```

AArnott:
I suppose. Do folks actually use that syntax? I just tried it and the language service seems to make it rather difficult.

AArnott:
Hmmmm... it seems #821 which you already linked to tracks support for this in general for this rule. It seems best to keep that as a separate pull request IMO.

sharwell:
I'm OK with that.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1028CodeFixProvider.cs
```diff
@@ -71,27 +86,55 @@ public sealed override FixAllProvider GetFixAllProvider()
...
+                                    newTriviaContent = string.Concat(
+                                        oldTriviaContent.Substring(0, diagnosticSpanWithinTrivia.Start),
+                                        oldTriviaContent.Substring(diagnosticSpanWithinTrivia.End));
+                                    var newToken = SyntaxFactory.Token(textToken.LeadingTrivia, textToken.Kind(), newTriviaContent, newTriviaContent, textToken.TrailingTrivia);
+                                    newTextTokens = newTextTokens.Replace(textToken, newToken);
+                                    break;
+                                }
+                            }
+
+                            var newNode = xmlNode.Update(newTextTokens);
```
sharwell: What is the difference between `Update` and `WithTextTokens`?

sharwell: Answer: The name. `WithTextTokens` is implemented by calling `Update`.

AArnott: And also: I couldn't find the `WithTextTokens` method for the life of me at the time I was coding this. I suspect I was trying before adding the downcast code I guess.

PR #856, opened 2015-05-17
Enable the batch fixer for SA1651 (DoNotUsePlaceholderElements)

sharwell:
In the initial implementation of SA1651, the batch fixer was explicitly disabled in order to require users manually review each placeholder tag for accuracy. This limitation is not required, because even with the batch fixer enabled, there are three other points where manual review is involved:

1. The UI for fixing multiple instances shows a list of changes that will occur, and allows the user to deselect items which should not be changed.
2. When committing changes to source control.
3. During the code review process for the team maintaining the affected code.

Disabling the batch fixer serves only to slow down the process of correcting multiple instances of this violation. This pull request enables the batch fixer for the SA1651 code fix provider.


vweijsters:
:+1: 

While point 2 will not be valid for all organizations the other two should be enough safe guard. I think it is good to give the control to the developer, it will reduce potential frustration.

PR #858, opened 2015-05-17
Improve IsImplementingAnInterfaceMember

### LINKED ISSUE #822: Revisit IsImplementingAnInterfaceMember for maximum performance
Currently `IsImplementingAnInterfaceMember` still has some room for improvement. This is a minor issue since the performance is not causing user problems at this time.

1. Return false immediately if `memberSymbol` is static
1. Perform the checks for `IMethodSymbol` immediately if `methodSymbol` is not null, rather than performing all the casts up front.

Maxwe11:
- Return false immediately if `memberSymbol` is static
- Eliminate unnecessary casts and checks on null

Closes #822


pdelvo:
Nice :+1:

Looks good to me.

PR #861, opened 2015-05-18
Implement new diagnostic SA1519 CurlyBrackets and SA1520 UseCurlyBracketsConsistently

### LINKED ISSUE #716: New rule proposal: Use braces consistently

### LINKED ISSUE #717: New rule proposal: Multi-line statements must be surrounded by braces

sharwell:
The SA1519 rule still needs a good name, but everything else should be in place. :smile: 

I updated SA1503 to not report warnings that are now reported as SA1519 or SA1520.

Fixes #716
Fixes #717

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1519UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1519UnitTests.cs
@@ -1,0 +1,627 @@
+﻿namespace StyleCop.Analyzers.Test.LayoutRules
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.LayoutRules;
+    using TestHelper;
+    using Xunit;
+
+    public class SA1519UnitTests : CodeFixVerifier
+    {
+        /// <summary>
+        /// Gets the statements that will be used in the theory test cases.
+        /// </summary>
+        /// <value>
+        /// The statements that will be used in the theory test cases.
+        /// </value>
+        public static IEnumerable<object[]> TestStatements
+        {
+            get
+            {
+                yield return new[] { "if (i == 0)" };
+                yield return new[] { "while (i == 0)" };
+                yield return new[] { "for (var j = 0; j < i; j++)" };
+                yield return new[] { "foreach (var j in new[] { 1, 2, 3 })" };
+            }
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will properly handle an empty source.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestEmptySourceAsync()
+        {
+            var testCode = string.Empty;
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that a statement followed by a single-line statement without curly braces will not produce a
+        /// warning.
+        /// </summary>
+        /// <param name="statementText">The source code for the first part of a compound statement whose child can be
+        /// either a statement block or a single statement.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Theory]
+        [MemberData(nameof(TestStatements))]
+        public async Task TestSingleLineStatementWithoutCurlyBracketsAsync(string statementText)
+        {
+            await this.VerifyCSharpDiagnosticAsync(this.GenerateSingleLineTestStatement(statementText), EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that a statement followed by a multi-line statement without curly braces will produce a warning.
+        /// </summary>
+        /// <param name="statementText">The source code for the first part of a compound statement whose child can be
+        /// either a statement block or a single statement.</param>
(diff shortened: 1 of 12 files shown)


sharwell:
This is currently blocked on approval for #716.

vweijsters:
The changes made to SA1503 will break SA1503 when SA1520 is disabled. I don't think that that is appropriate behavior.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1519UnitTests.cs
```diff
@@ -0,0 +1,627 @@
...
+}";
+
+            var expected = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(7, 13),
+                this.CSharpDiagnostic().WithLocation(10, 13)
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
```
vweijsters: I would have expected a code fix test here as well, as the test case for do ... while above provides.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1519UnitTests.cs
```diff
@@ -0,0 +1,627 @@
...
+}";
+
+            var expected = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(6, 21),
+                this.CSharpDiagnostic().WithLocation(6, 33)
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
```
vweijsters: I would expect a code fix here as well.

vweijsters: I found the code fix below, but it would make more sense to have it in this test case.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1520UnitTests.cs
```diff
@@ -0,0 +1,485 @@
...
+        if (i == 0)
+        {
+            Debug.Assert(true);
+        }
+        else
+        {
+            Debug.Assert(false);
+        }
+    }
+}";
```
vweijsters: would be nice to have some inline comments before and after statements to verify that they are handled properly.

vweijsters:
In general most of it looks good. 

I think we need a better mechanism for disabling certain diagnostics when others are present. 
In this pull request there are already issues when:
* SA1503 is enabled and SA1520 is not 
* SA1520 is enabled and SA1519 is not

There is no point in offer separated rules if there are these kinds of undocumented dependencies between rules.

sharwell:
> The changes made to SA1503 will break SA1503 when SA1520 is disabled. I don't think that that is appropriate behavior.

SA1520 would be enabled by default, and there would be no reason to disable it if SA1503 is enabled. Rather than report a violation as both SA1503 and SA1520, certain code would be reported as SA1520 *instead* of SA1503. Note that SA1520 does not report anything that wasn't previously reported as SA1503; it is a strict subset.

Another way to look at this is we are dividing SA1503 into multiple rules in order to better address the needs of project that historically needed to disable SA1503 and thus lost *all* validation of curly brackets, when in reality some validation was desired.

vweijsters:
Whether or not there is a reason to disable SA1520 is besides the point. If some developer decides he doesn't want SA1520 enabled, the functionality of SA1503 is reduced (this point isn't about the new rules at all, there's more examples of this in other existing rules).

I think that developers should be able to consider our set of analyzers as an orthogonal set of analyzers where they can pick whatever analyzers they want. If we want to minimize the amount of diagnostics for places where there is overlap, we should do so without reducing the functionality of the original analyzer, if the overlapping analyzer is disabled for some reason.

I would prefer a situation where SA1503 would check if SA1520 is active, and if so, suppress one of its ReportDiagnostic calls. I think it would be doable to determine such a situation from the default state of a rule combined with the `SpecificDiagnosticOptions` from the project containing the source file.

sharwell:
@vweijsters Something like that?

vweijsters:
:+1:  That looks really good, its exactly what I had in mind.

I think it would also be nice to wrap the entire check in an extension method. Something like:
```C#
if (context.IsDiagnosticEnabled(SA1519CurlyBrackets.DiagnosticId))
{
    ...
}
```

I'll be happy to do that later as a refactoring as well.

vweijsters:
If you could also add a check for SA1519 enabled to `SA1520UseCurlyBracketsConsistently.cs` line 113-117, that would be great as well.

PR #866, opened 2015-05-19
Fix SA1211 + unit tests

### LINKED ISSUE #290: Diagnostic unit tests: SA1211 (UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName)
Implement unit tests for #71.

### LINKED ISSUE #612: SA1211 reported for incorrect comparison
Currently SA1211 is reported for the first of the following using alias directives:

```csharp
using StringBuilder = System.Text.StringBuilder;
using StringWriter = System.IO.StringWriter;
```

It appears the analysis is examining the target name, not the alias name.

### LINKED ISSUE #779: SA1211 (UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName) efficiency improvement
Currently SA1211 is implemented by registering a handler for using directives. This could be improved by instead registering a handler for compilation unit and namespaces, similar to the way SA1201 is implemented.

Maxwe11:
Fixes #290
Fixes #612
Fixes #779

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1211UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1211UnitTests.cs
@@ -1,0 +1,149 @@
+﻿namespace StyleCop.Analyzers.Test.OrderingRules
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.OrderingRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1211UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName"/>.
+    /// </summary>
+    public class SA1211UnitTests : DiagnosticVerifier
+    {
+        /// <summary>
+        /// Verifies that the analyzer will properly handle an empty source.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestEmptySourceAsync()
+        {
+            var testCode = string.Empty;
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will not produce diagnostics for correctly ordered using directives inside a namespace.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestValidUsingDirectivesInNamespaceAsync()
+        {
+            const string testCode = @"namespace Foo
+{
+    using System;
+    using Character = System.Char;
+    using @int = System.Int32;
+    using StringBuilder = System.Text.StringBuilder;
+    using StringWriter = System.IO.StringWriter;
+}
+";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will not produce diagnostics for correctly ordered using directives in multiple namespaces.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestValidUsingDirectivesInMultipleNamespacesAsync()
+        {
+            const string testCode = @"namespace Foo
+{
+    using System;
+    using character = System.Char;
+    using \u0069nt = System.Int32;
+}
+
+namespace Bar
+{
+    using System;
+    using MemoryStream = System.IO.MemoryStream;
+    using Stream = System.IO.Stream;
+    using StringBuilder = System.Text.StringBuilder;
+    using StringWriter = System.IO.StringWriter;
+}
+
+namespace Spam
+{
+    using System;
+    using Character = System.Char;
+    using @int = System.Int32;
+}
+";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will not produce diagnostics for correctly ordered using directives in the compilation unit.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1211UnitTests.cs
```diff
@@ -0,0 +1,128 @@
...
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.OrderingRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1217UsingStaticDirectivesMustBeOrderedAlphabetically"/>.
```
Maxwe11: Oops...will fix with addressing feedback

sharwell:
Also fixes #290. :smile:

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1211UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName.cs
```diff
@@ -47,54 +49,66 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
 
-                string alias = syntax.Alias.Name.ToString();
-                string precedingAlias = usingDirective.Alias.Name.ToString();
-                if (string.Compare(alias, precedingAlias, StringComparison.OrdinalIgnoreCase) >= 0)
+                string currentAliasName = usingDirective.Alias.Name.ToString();
+                if (prevAliasUsingDirective != null)
                 {
-                    continue;
+                    string prevAliasName = prevAliasUsingDirective.Alias.Name.ToString();
+                    if (string.Compare(prevAliasName, currentAliasName, StringComparison.OrdinalIgnoreCase) > 0)
```
sharwell: :bulb: This currently sorts the following incorrectly:

```csharp
using _DTE = EnvDTE._DTE;
using DTE = EnvDTE.DTE;
```

To correct the situation, I used the following:

```csharp
string.Compare(prevAliasName.ToLowerInvariant(), currentAliasName.ToLowerInvariant(), StringComparison.Ordinal)
```

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1211UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName.cs
```diff
@@ -47,54 +49,66 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                    {
+                        // Find alias before which should be placed current alias
+                        string alias = usingAliasNames.FirstOrDefault(aliasName =>
+                        {
+                            return string.Compare(aliasName, currentAliasName, StringComparison.OrdinalIgnoreCase) > 0;
```
sharwell: :bulb: This suffers from the same issue described above.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1211UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName.cs
```diff
@@ -47,54 +49,66 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
                 // only interested in using alias directives
                 if (usingDirective.Alias?.Name?.IsMissing != false)
                 {
                     continue;
                 }
 
-                string alias = syntax.Alias.Name.ToString();
-                string precedingAlias = usingDirective.Alias.Name.ToString();
-                if (string.Compare(alias, precedingAlias, StringComparison.OrdinalIgnoreCase) >= 0)
+                string currentAliasName = usingDirective.Alias.Name.ToString();
```
sharwell: :exclamation: I think this needs to be `usingDirective.Alias.Name.Identifier.ValueText`. To verify, add unit tests with the following:

1. A using alias with an `@` sign. The following should *not* produce a warning (and the reverse should).

    ```csharp
    using Character = System.Char;
    using @int = System.Int32;
    ```

2. A using directive with an escape sequence. The following should *not* produce a warning (and the reverse should).

    ```csharp
    using character = System.Char;
    using \u0069nt = System.Int32;
    ```

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1211UsingAliasDirectivesMustBeOrderedAlphabeticallyByAliasName.cs
```diff
@@ -47,54 +49,66 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                if (prevAliasUsingDirective != null)
                 {
-                    continue;
+                    string prevAliasName = prevAliasUsingDirective.Alias.Name.ToString();
```
sharwell: :exclamation: I think this needs to use `ValueText` as well. See the previous remark.

sharwell:
Aside from the above, eventually this diagnostic will need to "reset" its state when it crosses an `#if`/`#elif`/`#else`/`#endif`, or you get a slew of diagnostics for something like this:

https://github.com/tunnelvisionlabs/InheritanceMargin/blob/566f0d677cccc431a2f146a9f567b3eec3e97485/Tvl.VisualStudio.InheritanceMargin.CSharp/CSharpInheritanceAnalyzer.cs#L3-L44

sharwell:
:memo: I completed my review, and marked this **do not merge** while you update it. Make sure to comment here when you finish updating the request, since GitHub won't notify me if you only push to the branch.

Maxwe11:
@sharwell thanks for the review. I'll take a look, but on the next week(

Maxwe11:
@sharwell updated

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1211UnitTests.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    using System;
+    using @int = System.Int32;
+    using Character = System.Char;
+}
+";
+
+            DiagnosticResult[] expectedDiagnostics =
+            {
+                this.CSharpDiagnostic().WithLocation(5, 5).WithArguments("character", "\u0069nt"),
```
sharwell: :bulb: The second argument can be written as just `"int"`; the escape sequence is misleading.

Maxwe11: fixed

PR #869, opened 2015-05-22
Fix that SA1304 is reported for accessible fields

### LINKED ISSUE #574: Accessible readonly fields are reported twice
Currently an externally-accessible `readonly` fields is reported as both SA1304 and SA1307. In this case, only SA1307 should be reported (see #557 for more information).

I'm marking this as enhancement instead of bug because the reference implementation of StyleCop reports both warnings for this case. We are only reporting one to simplify the presentation and code fix for users.

pdelvo:
We should make it possible to test if the diagnostic works if SA1307 is disabled. I dont think our testing framework currently supports that.

Fixes #574.


sharwell:
> We should make it possible to test if the diagnostic works if SA1307 is disabled

Sounds like a good new issue to create. :smile:

vweijsters:
I think that #522 is all that is needed to be able to test it.

PR #872, opened 2015-05-24
Disable alternative rules by default

sharwell:
I unintentionally enabled alternative rules by default when I was looking for a solution for #841. This pull request corrects that mistake.


sharwell:
@pdelvo This change may break your site's status reporting for SX* rules.

PR #877, opened 2015-05-25
Update to the latest release (alpha 7) of StyleCop.Analyzers

sharwell:


pdelvo:
Looks good to me. We might want issues for enabling SA1512 and SA1515 in the next release.

sharwell:
Yes, I certainly plan to do that.

PR #882, opened 2015-05-25
Fix that SA1612 does not take order into account

### LINKED ISSUE #867: SA1612 (ElementParameterDocumentationMustMatchElementParameters) Does not take parameter order into account
SA1612 (ElementParameterDocumentationMustMatchElementParameters) Is not checking the order of the param documentation is in the same order as the method parameters.

The following fails to report SA1612:	

    /// <summary>
    /// foo
    /// </summary>
    /// <param name="bar"></param>
    /// <param name="foo"></param>
    public bool Method(string foo, string bar)
    {
	    return true;
    }

pdelvo:
Fixes #867.

The implementation for this rule is now symmetric to the one I did for SA1620.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1612ElementParameterDocumentationMustMatchElementParameters.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1612ElementParameterDocumentationMustMatchElementParameters.cs
@@ -1,5 +1,6 @@
 ﻿namespace StyleCop.Analyzers.DocumentationRules
 {
+    using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Linq;
     using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1612ElementParameterDocumentationMustMatchElementParameters.cs
```diff
@@ -36,11 +37,17 @@ public class SA1612ElementParameterDocumentationMustMatchElementParameters : Dia
         private const string Description = "The documentation describing the parameters to a C# method, constructor, delegate or indexer element does not match the actual parameters on the element.";
         private const string HelpLink = "http://www.stylecop.com/docs/SA1612.html";
 
-        private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
+        private const string MissingParamForDocumentationMessageFormat = "The parameter '{0}' does not exist.";
```
sharwell: :bulb: It seems like these replace the `MessageFormat` field; it could be removed.

pdelvo: Yes the warning is only for that case. The location of xxx is the location of the diagnostic, so I think the message should be okay. Or do you have a better message for it?

Im going to remove the unused constant

sharwell: > Yes the warning is only for that case. The location of xxx is the location of the diagnostic, so I think the message should be okay. Or do you have a better message for it?

That's fine. I was originally thinking it was warning that the `<param>` element was missing for a parameter `xxx`.

pdelvo:
I removed the unused constant

PR #889, opened 2015-05-27
Fix exception in SA1619

### LINKED ISSUE #888: SA1619 (GenericTypeParametersMustBeDocumentedPartialClass) NullReferenceException for non-generic partial class
SA1619 throws a `NullReferenceException` when analyzing certain non-generic partial classes. For example:

```csharp
/// <summary>
/// Summary Text.
/// </summary>
internal partial class ClassName
{
}
```

This error is due to the fact that `typeDeclaration.TypeParameterList` is null for this type on [this line](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/bfc8c85adf5dfc481222786d2422dde7929057da/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1619GenericTypeParametersMustBeDocumentedPartialClass.cs#L144).

sharwell:
This change includes two parts:

* Pass through diagnostic AD0001 as an error during testing This change ensures exceptions thrown during the evaluation of a test are properly detected by the test infrastructure.
* (In progress) Fixes #888

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
@@ -65,6 +65,8 @@ namespace TestHelper
             }
 
             var supportedDiagnosticsSpecificOptions = new Dictionary<string, ReportDiagnostic>();
+            // Report exceptions during the analysis process as errors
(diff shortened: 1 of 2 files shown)


sharwell:
Fun fact: If you update the pull request before AppVeyor builds it, it will associate the build result with the wrong commit. The commit in the following image should have failed, but AppVeyor actually built ebf0419 when it ran the test.

image

sharwell:
I'm correcting it by force-pushing the first commit as the pull request, and then I'll update it again after the build fails.

PR #891, opened 2015-05-27
Update to the latest release (alpha 8) of StyleCop.Analyzers

sharwell:
Notes:

* SA1512 is now enabled (bugs fixed during the Alpha 8 release allowed this)
* SA1515 is now enabled (bugs fixed during the Alpha 8 release allowed this)
* SA1633 is disabled (we haven't settled on a specific file header at this point, plus I don't want to enable it prior to a code fix being available)


vweijsters:
:+1: Looks good to me

PR #893, opened 2015-05-27
Implements a Settings helper class

### LINKED ISSUE #892: Proposal: Settings file
### Scope
This issue describes a proposal for settings files for the StyleCopAnalyzers analyzers. It is similar to the original StyleCop settings files, but deliberately different, to signify the differences between the two projects. 

The settings files for the StyleCopAnalyzers analyzers will only feature settings that can be used to configure certain diagnostics. There will be no mechanism to disable analyzers, the standard rule set mechanism from Visual Studio can be used for that.

The settings file for the StyleCopAnalyzers will be stored using the JSON data-interchange format.

### Example
The example below shows the content of a possible settings file. This example also shows all currently identified settings.

```JSON
{
  "mergeMode" : "parentFolder",
  "settings" :
  {
    "documentationRules" :
    {
      "companyName" : "FooBar Inc.",
      "copyrightText" : "Copyright (c) 2015 FooBar Inc. All rights reserved.",
      "ignorePrivateDeclarations" : true,
      "ignoreInternalDeclarations" : true,
      "ignoreFields" : true
    }
  }
}
```
### Naming
The settings file shall be named `Settings.StyleCopAnalyzers`. This name was chosen to be similar to the original StyleCop name, to indicate to developers its intended use, but also different to signify that the contents are not compatible.

### Locating the settings file
When analyzing a source file, the analyzer will look in the folder where the source file resides for a `Settings.StyleCopAnalyzers` file. If no settings file is present, the parent folder is searched for a settings file. If no settings file is present there, the parent of the parent folder is searched, and so on.
(See the section on merging for more details)

If no settings files are found, the default settings are used.

### Validation
The settings file shall be validated as being proper JSON and a diagnostic must be produced to signal an invalid settings file. Whenever an invalid settings file is encountered, all valid settings from already parsed (and merged) settings files must be retained. The merge process must stop immediately when an invalid settings file is encountered.

If the settings file contains unknown name/value pairs, they shall be silently ignored.

### Merging
To allow reuse of settings files, it is useful to have a merging scheme available, where multiple settings files will be combined into a single set of settings. When merging the current set of  settings with another settings file, the current set of settings will override any settings defined in the file being merged.

The proposed merge scheme is similar to the one that StyleCop is using. It will however not have a 'master' settings file in the installation folder.

The merge process will be controlled by the optional `<merge>` element within each settings file. One (and only one) of three merge modes can be specified:
* Do not merge this settings file. This will stop the merge process. 
```JSON
  "mergeMode" : "none"
```
* Merge with a settings file present in a parent folder. This does not have to be the direct parent folder, the entire tree towards the root of the filesystem is traversed until a settings file is found. This is the default option, if no merge element is specified.
```JSON
  "mergeMode" : "parentFolder"
```
* Merge with a specified settings file, and continue the merge process from there. 
```JSON
  "mergeMode" : "file"
  "mergeFile" : "D:\\GitHub\\Settings.StyleCopAnalyzers"
```

### Related issues
This would be used for implementing the following issues: #156, #161, and #789.
It is related to #410 and #422
It would also influence pull request #832

vweijsters:
Fixes #892 

This is work in progress. I created the pull request to get early feedback.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsUnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsUnitTests.cs
@@ -1,0 +1,114 @@
+﻿namespace StyleCop.Analyzers.Test.Settings
+{
+    using System.Collections.Immutable;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.CodeAnalysis.Text;
+    using Xunit;
+
+    public class SettingsUnitTests
+    {
+        private const string TestProjectName = "TestProject";
+
+        /// <summary>
+        /// Verifies that proper defaults are returned when there are no additional files.
+        /// </summary>
+        [Fact]
+        public void VerifySettingsDefaults()
+        {
+            var styleCopSettings = SettingsHelper.GetStyleCopSettings(default(SyntaxTreeAnalysisContext));
+
+            Assert.Equal("PlaceHolderCompany", styleCopSettings.CompanyName);
+            Assert.Equal("Copyright (c) PlaceHolderCompany. All rights reserved.", styleCopSettings.CopyrightText);
+        }
+
+        /// <summary>
+        /// Verifies that the settings are properly read.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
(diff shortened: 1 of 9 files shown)


vweijsters:
@sharwell @pdelvo Can you have a look and check if I'm going in the right direction with this?

vweijsters:
Updated to match the current state of #892

sharwell:
The implementation looks fine overall. Can you do the following?

1. Rebase/squash to a commit on **master**.
2. Update **StyleCop.Analyzers.nuspec** to include the Json.NET dependency (under binaries and symbols).

vweijsters:
@sharwell I've made the requested changes.

PR #896, opened 2015-05-29
SA1210 Efficiency improvement

### LINKED ISSUE #289: Diagnostic unit tests: SA1210 (UsingDirectivesMustBeOrderedAlphabeticallyByNamespace)
Implement unit tests for #70.

### LINKED ISSUE #778: SA1210 (UsingDirectivesMustBeOrderedAlphabeticallyByNamespace) efficiency improvement
Currently SA1210 is implemented by registering a handler for using directives. This could be improved by instead registering a handler for compilation unit and namespaces, similar to the way SA1201 is implemented.

baks:
Fixes #778 
Fixes #289 

Changed SA1210 code to process compilation unit and namespace declaration instead of using directives.

Created `IsSystemUsingDirective` extension method for `UsingDirectiveSyntax`, and modified SA1208/SA1210 rule to use that method.
Also added unit tests for SA1210 rule.

I didn't address issue with preprocessor conditional directives that @sharwell mentions in #866.
I think for that problem there is a need to create issues similar as for efficiency improvement in using directives.


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
```diff
@@ -50,108 +56,60 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
         /// <inheritdoc/>
         public override void Initialize(AnalysisContext context)
         {
-            context.RegisterSyntaxNodeActionHonorExclusions(this.HandleUsingDirectiveSyntax, SyntaxKind.UsingDirective);
+            context.RegisterSyntaxNodeActionHonorExclusions(this.HandleCompilationUnit, SyntaxKind.CompilationUnit);
+            context.RegisterSyntaxNodeActionHonorExclusions(this.HandleNamespaceDeclaration, SyntaxKind.NamespaceDeclaration);
         }
 
-        private void HandleUsingDirectiveSyntax(SyntaxNodeAnalysisContext context)
+        private void HandleCompilationUnit(SyntaxNodeAnalysisContext context)
```
Maxwe11: Why not make this method `static`?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
```diff
@@ -50,108 +56,60 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
-            }
-
-            foreach (var usingDirective in usingDirectives)
-            {
-                // we are only interested in nodes before the current node
-                if (usingDirective == syntax)
-                {
-                    break;
-                }
+            var compilationUnit = (CompilationUnitSyntax)context.Node;
```
Maxwe11: It's very very unlikely that this method will be invoked with passed non `CompilationUnitSyntax` as argument but it's possible. I'd prefer `as` cast and null check.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
baks: It should be static, I didn't noticed that it isn't :(

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
```diff
@@ -50,108 +56,60 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
-                if (string.Compare(fullyQualifiedNamespace, precedingFullyQualifiedNamespace, StringComparison.OrdinalIgnoreCase) >= 0)
-                {
-                    continue;
-                }
+            ProcessUsings(namespaceDeclaration.Usings, context);
+        }
 
-                string @namespace = namespaceSymbol.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat);
-                string precedingNamespace = precedingNamespaceSymbol.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat);
+        private static void ProcessUsings(SyntaxList<UsingDirectiveSyntax> usings, SyntaxNodeAnalysisContext context)
```
Maxwe11: I would swap method arguments. Implemented rules in project  that I've seen before accepts `context` as first argument. I think we should be consistent in such things.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
```diff
@@ -50,108 +56,60 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
-            if (namespaceSymbol == null)
+            if (usings.Count <= 1)
             {
-                return null;
+                return;
             }
 
-            string fullyQualifiedName = namespaceSymbol.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat);
-            string name = fullyQualifiedName;
+            var properPositionsOfUsingDirectives = usings.OrderBy(directive => directive.Name.ToUnaliasedString()).Select((directive, idx) => new { Directive = directive, Index = idx }).ToDictionary(item => item.Directive, item => item.Index);
```
Maxwe11: I'm not sure but could we replace workarounds with positions by comparing existing using with ordered and in case of not equality report diagnostic.

Maxwe11: I mean something like this:
```C#
var oredered = usings.OrderBy(directive => directive.Name.ToUnaliasedString()).ToList();
            var en = usings.GetEnumerator();
            
            foreach (UsingDirectiveSyntax usingDirective in oredered)
            {
                en.MoveNext();
                if (!object.ReferenceEquals(usingDirective, en.Current))
                {
                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, usingDirective.GetLocation(), usingDirective.Name.ToUnaliasedString(), en.Current.Name.ToUnaliasedString()));
                    break;
                }
            }
```

but with this code two unit test fail

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
baks: I just think using cast in that situation better state my intent because I'm sure about the object type. If I used `as` operator it would be mean that I'm not sure.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
baks: Yeah, good catch :+1:

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
Maxwe11: :+1: got it

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
baks: So, I used number of order position for given using directive because it helps me check if in previous directives is any directive that has greater order position  than the directive that is currently checked.
I think checking equality between lists as in your code, would not be enough to generate proper diagnostics.

For now, I can't figure out something different, but maybe someone came up with better idea :+1:

Maxwe11: Yeah, I missed that. I think we should stop analyzing at the first occurrence of SA1210.
I suggest following implementation.
```c#
private static void CheckOrderAndReportDiagnostic(SyntaxNodeAnalysisContext context, List<UsingDirectiveSyntax> usings)
{
    if (usings.Count <= 1)
    {
        return;
    }

    UsingDirectiveSyntax prevUsingDirective = null;
    var usingsNames = new List<string>(usings.Count);

    foreach (var usingDirective in usings)
    {
        string currentDirectiveName = usingDirective.Name.ToUnaliasedString();
        if (prevUsingDirective != null)
        {
            string prevDirectiveName = prevUsingDirective.Name.ToUnaliasedString();
            if (string.CompareOrdinal(prevDirectiveName, currentDirectiveName) > 0)
            {
                // Find directive before which should be placed current directive
                foreach (string usingName in usingsNames)
                {
                    if (string.CompareOrdinal(usingName, currentDirectiveName) > 0)
                    {
                        prevDirectiveName = usingName;
                        break;
                    }
                }

                // Using directive for '{currentDirectiveName}' must appear before directive for '{prevDirectiveName}'
                context.ReportDiagnostic(Diagnostic.Create(Descriptor, usingDirective.GetLocation(), currentDirectiveName, prevDirectiveName));
                return;
            }
        }

        usingsNames.Add(currentDirectiveName);
        prevUsingDirective = usingDirective;
    }
}
```
I intentionally avoid LINQ because it requires closure allocations.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
sharwell: > It's very very unlikely that this method will be invoked with passed non `CompilationUnitSyntax` as argument but it's possible. I'd prefer `as` cast and null check.

This came up in another review recently. If this cast ever fails, it's a bug in the Roslyn framework not our code. The cast is preferable when only one input is valid.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
baks: I think before we can go any further it's good to clarify how that rule should behave.

Given that example:
```csharp
using System.Threading;
using System.IO;
using System;
using System.Linq;
```

In my opinion it should returns 3 diagnostics:
`using System.IO must appear before System.Threading`
`using System must appear before System.IO`
`using System.Linq must appear before System.Threading`

if someone apply diagnostics in order they appear, he obtains correctly ordered usings.
I'm not sure about second rule because it might be confusing, if somebody is randomly checking diagnostics.

So, if previous is a bit confusing, the diagnostics can be:
`using System.IO must appear before System.Threading`
`using System must appear before System.Threading`
`using System.Linq must appear before System.Threading`

@sharwell , @Maxwe11, what are your thoughts about that?

Maxwe11: >if someone apply diagnostics in order they appear

Key word **if** :smile: I think diagnostics shouldn't depend on order.
e.g. [similar SA1217 implemetation](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/786/files#diff-d0889bf2040e93dea98e68c5b0e84345R79).

vweijsters: I agree with @Maxwe11 here. 

I think the discussion in #828 basically holds for this one as well. We should use generic messages, because it will be very difficult to get the order correct in all cases and not worth it at all imho.

I ran the example of @baks through StyleCop and it gives diagnostics for the first two using statements only, which seems proper to me

baks: Makes sense.
It looks like StyleCop gives generic message for every using directive that following directive is less than:

```csharp
    using System.Threading;
    using System.Linq;
    using System.IO;
    using System.Threading.Tasks;
    using System.Security;
```
For that example StyleCop generates message for `System.Threading`, `System.Linq`, `System.Threading.Tasks`.

Just to clarify, should we go the same?

vweijsters: Yes, I think we should.

sharwell: Yes, this makes it easy to do one pass through the list, and if only one item is out of place we'll only report one diagnostic.

baks:
I changed SA1210 rule to generate diagnostics like StyleCop.
Also I changed ToUnaliasedString extension method to return names correctly when there is unicode escape sequence.

baks:
@sharwell, @Maxwe11, @vweijsters 
Any thoughts about last commit?

sharwell:
Last commit looks fine. Haven't been through the rest yet.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1210UnitTests.cs
```diff
@@ -0,0 +1,188 @@
...
+{
+    using global::Foo;
+    using System;
+}";
+
+            DiagnosticResult[] expected = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(2, 1),
+                this.CSharpDiagnostic().WithLocation(4, 1),
+            };
```
vweijsters: This is not correct. It should report diagnostics for lines 2, 3, and 4.
I verified this against the StyleCop implementation as well.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
```diff
@@ -50,108 +54,52 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+            ProcessUsings(context, namespaceDeclaration.Usings);
+        }
 
-                string @namespace = namespaceSymbol.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat);
-                string precedingNamespace = precedingNamespaceSymbol.ToDisplayString(SymbolDisplayFormat.CSharpErrorMessageFormat);
+        private static void ProcessUsings(SyntaxNodeAnalysisContext context, SyntaxList<UsingDirectiveSyntax> usings)
+        {
+            var usingsDirectivesWithoutAliasAndStatic = usings.Where(ExcludeAliasAndStaticUsingDirectives);
+            var systemUsingDirectives = usingsDirectivesWithoutAliasAndStatic.Where(directive => directive.IsSystemUsingDirective());
+            var usingDirectives = usingsDirectivesWithoutAliasAndStatic.Where(directive => !directive.IsSystemUsingDirective());
```
vweijsters: :bulb: This is not really efficient, as it is doing multiple iterations over the same (part of the) collection.

I would suggest something like:
```C#
var usingDirectives = new List<UsingDirectiveSyntax>();
var systemUsingDirectives = new List<UsingDirectiveSyntax>();

foreach (var usingDirective in usings)
{
    if (IsAliasorStaticUsingDirective(usingDirective))
    {
        continue;
    }

    if (usingDirective.IsSystemUsingDirective())
    {
        systemUsingDirectives.Add(usingDirective);
    }
    else
    {
        usingDirectives,Add(usingDirective);
    }
}

CheckIncorrectlyOrderedUsingsAndReportDiagnostic(context, usingDirectives);
CheckIncorrectlyOrderedUsingsAndReportDiagnostic(context, systemUsingDirectives);

```

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1210UnitTests.cs
sharwell: :bulb: What should the correct order be? If you configure Visual Studio to place System using directives first, it will sort them in the following order:

```cs
using System.Threading;
using global::System;
using global::System.IO;
using global::System.Linq;
```

This would suggest the only warning be given on line 3. Any other ordering means you can no longer use Visual Studio's built-in operation to sort usings, and doing so could result in compiler warnings for some users.

vweijsters: I think the correct order would be:
```C#
using global::System;
using global::System.IO;
using global::System.Linq;
using System.Threading;
```
Basically that means ignoring the `global::` prefix which seems logical to me. 

If we would follow Visual Studio's way of ordering, diagnostics should be produced for both lines 3 and 4, as they both should be after the `global:System`.

@sharwell Do you know if there is a way to replace the built-in sort usings operator? If so, we could adapt the sort order, if not we should indeed seriously consider to follow Visual Studio's ordering.

sharwell: > diagnostics should be produced for both lines 3 and 4

To control the number of warnings reported and to simplify analysis, I think the analyzers which examine `using` directives should follow this one simple rule: Report a warning on the current using directive if the current using directive should appear after the next using directive. The code fix would operate by reordering all using directives to match the rules.

> Do you know if there is a way to replace the built-in sort usings operator?

I don't think we can, at least not easily. Since most users do not globally-qualify using directives, I don't think the deviation from StyleCop will be a pain point.

vweijsters: ok, that sounds fine to me.

Do we want to create an updated documentation page for this rule as well?

baks: @sharwell but if you turn off that option, Visual Studio sort usings in that way: (same as @vweijsters wrote)
```csharp
using global::System;
using global::System.IO;
using global::System.Linq;
using System.Threading;
```

@sharwell, currently SA1210 analyzer is working same as you wrote, it reports warning when the current using directive should be after the next using directive.

So, in that unit tests where should diagnostics be reported?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1210UsingDirectivesMustBeOrderedAlphabeticallyByNamespace.cs
baks: Ok, I'll fix that

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1210UnitTests.cs
sharwell: When you turn off that option, it changes the behavior `using Microsoft.Win32.SafeHandles`.

I think a warning should only appear line 3:

* `using System` should be before `using global::System.IO`, therefore no warning on line 1.
* `using global::System.IO` should be before `using global::System.Linq`, therefore no warning on line 2.
* `using global::System.Linq` should be after `using global::System` &rarr; **warning on line 3**
* `using global::System` is last, therefore no warning on line 3.

baks: @sharwell, in this example on first line we have `using System.Threading` not `using System`

sharwell: @baks You are correct, but it does not change the outcome.

sharwell: Actually there are two possibilities. I'm in favor of following Visual Studio on this one and not reporting a warning, but the alternative is also a reasonable interpretation of the StyleCop rule so we need to make a decision.

vweijsters: My preference would also be to follow Visual Studio, as that would make things more smooth for most developers.

baks: So, I'll change SA1210 analyzer to follow Visual Studio

baks: Just to clarify one more example:
```csharp
using System.Threading;
using global::System.IO;
using global::System.Linq;
using global::System;
using global::Abc;
```

According to how Visual Studio order this, we should return warnings on line 3 and 4?

sharwell: @baks That's correct.

baks:
@sharwell , @vweijsters  I made changes to SA1210 rule

vweijsters:
:+1: Looks good to me

felickz:
@vweijsters mentioned updating documentation, also a candidate for [KnownChanges.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/KnownChanges.md) ?  I found myself in this thread after investigating why we are now alphabetical vs semantically ordered.

sharwell:
@felickz This is [already included](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/KnownChanges.md#sa1208) on the KnownChanges page. Did you want to see more? If you have suggestions for making this clearer let me know. :smile:

felickz:
Apologies, this issue was reported on SA1210 but i also see 1208 tightly coupled.

The "g" in "global::" is getting involved in the alphabetical ordering whereas before it looks to have not been.

Before: 

    using System;

    using Microsoft.Owin.Infrastructure;

    using global::Owin.Security.Providers

After:

    using System;

    using global::Owin.Security.Providers;

    using Microsoft.Owin.Infrastructure;

sharwell:
@felickz Thanks for clarifying that case. :smile: I filed #2071 to update the known changes documentation to address it specifically.

PR #898, opened 2015-05-29
Minor performance improvement for SA1211

Maxwe11:
In case if aliases should be reordered, current implementation search in list of
processed aliases where current alias should be placed. Search begins from the
beginning of list. We can improve this search making lookup from the end.


vweijsters:
I think the discussion of #828 applies here as well. In my opinion we should just report a generic message and leave the ordering to the user or the codefix in #650, thereby simplifying the code here a lot.

Maxwe11:
@sharwell thoughts?

sharwell:
Seems like we could make a base class for all the `using` declaration analyzers, which just asks the question:

> Should the current using declaration appear after the next using declaration? If so, a diagnostic gets reported for the current using declaration.

Such a base class could easily handle the preprocessor conditions described in #866 by simply not calling the comparison method if a preprocessor line appears between the two.

PR #902, opened 2015-05-30
Update to StyleCop.Analyzers 1.0.0 Alpha 9

sharwell:
SA1508 is temporarily disabled due to a bug (it's throwing an exception which is reported as AD0001).


vweijsters:
:+1: Looks good

PR #905, opened 2015-05-31
SA1024 codefix

Maxwe11:


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1024CodeFixProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+                if (!diagnostic.Id.Equals(SA1024ColonsMustBeSpacedCorrectly.DiagnosticId))
+                {
+                    continue;
+                }
+
+                SyntaxToken token = root.FindToken(diagnostic.Location.SourceSpan.Start);
+                if (token.IsMissing)
+                {
+                    continue;
+                }
```
vweijsters: This should never occur, if this ever happens, there is a bug in Roslyn. I think this statement should be removed. It would also remove the need to get the syntax root here, improving the performance of this code. The penalty for obtain the syntax root can then be taken when the user activates the code fix.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1024CodeFixProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (!hasPrecedingSpace)
+            {
+                // only the first token on the line has leading trivia, and those are ignored
+                SyntaxToken precedingToken = token.GetPreviousToken();
+                SyntaxTriviaList triviaList = precedingToken.TrailingTrivia;
+                if (triviaList.Count > 0 && !triviaList.Last().IsKind(SyntaxKind.MultiLineCommentTrivia))
+                {
+                    hasPrecedingSpace = true;
+                }
+            }
```
vweijsters: The code above makes a big assumption that trivia between tokens on the same line are always placed as trailing trivia of the first token. An alternative would be to merge the trailing and leading trivia of the two tokens and do analysis on that
```C#
    var combinedTrivia = precedingToken.TrailingTrivia.AddRange(token.LeadingTrivia);
```

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1024CodeFixProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            if (missingFollowingSpace && requireBefore && !hasPrecedingSpace)
+            {
+                SyntaxTrivia whitespace = SyntaxFactory.Space;
+                SyntaxToken corrected = token.WithTrailingTrivia(token.TrailingTrivia.Insert(0, whitespace))
+                                             .WithLeadingTrivia(token.LeadingTrivia.Add(whitespace));
+                Document updatedDocument = document.WithSyntaxRoot(root.ReplaceToken(token, corrected));
+                return Task.FromResult(updatedDocument);
+            }
+            else if (missingFollowingSpace)
```
vweijsters: The `else if` construction is not needed, as the successful `if` above will exit the method.

vweijsters:
:+1: Look good. I've made few remarks, but nothing major.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1024CodeFixProvider.cs
Maxwe11: Should we also fix this in diagnostic?

Maxwe11:
@vweijsters updated. Will squash if it's :ok_hand:  for you.

PR #908, opened 2015-05-31
SA1026 unit tests

### LINKED ISSUE #234: Diagnostic unit tests: SA1026 (CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation)
Implement unit tests for #32.

Maxwe11:
Also all occurrences of `Task.FromResult(true)` replaced with cached instance

Closes #234

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeFixProvider.cs
@@ -2,10 +2,12 @@
 {
     using System.Collections.Immutable;
     using System.Composition;
+    using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CodeActions;
     using Microsoft.CodeAnalysis.CodeFixes;
+    using StyleCop.Analyzers.Helpers;
 
     /// <summary>
     /// Implements a code fix for <see cref="SA1026CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation"/>.
@@ -31,10 +33,8 @@
         }
 
         /// <inheritdoc/>
-        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
         {
-            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
-
             foreach (var diagnostic in context.Diagnostics)
             {
                 if (!diagnostic.Id.Equals(SA1026CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation.DiagnosticId))
@@ -42,22 +42,23 @@
                     continue;
                 }
 
-                SyntaxToken token = root.FindToken(diagnostic.Location.SourceSpan.Start);
-                if (token.IsMissing)
-                {
-                    continue;
-                }
-
-                context.RegisterCodeFix(CodeAction.Create(SpacingResources.SA1026CodeFix, t => GetTransformedDocumentAsync(context.Document, root, token)), diagnostic);
+                context.RegisterCodeFix(CodeAction.Create(
+                    SpacingResources.SA1026CodeFix,
+                    token => GetTransformedDocumentAsync(context.Document, diagnostic, token)),
+                    diagnostic);
             }
+
+            return Task.FromResult(true);
         }
 
-        private static Task<Document> GetTransformedDocumentAsync(Document document, SyntaxNode root, SyntaxToken token)
+        private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
         {
-            SyntaxToken corrected = token.WithoutTrailingWhitespace().WithoutFormatting();
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            SyntaxToken token = root.FindToken(diagnostic.Location.SourceSpan.Start);
+            SyntaxToken corrected = token.WithoutTrailingWhitespace(removeEndOfLineTrivia: true).WithoutFormatting();
             Document updatedDocument = document.WithSyntaxRoot(root.ReplaceToken(token, corrected));
 
-            return Task.FromResult(updatedDocument);
+            return updatedDocument;
         }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation.cs
@@ -32,7 +32,7 @@
         private static readonly string HelpLink = "http://www.stylecop.com/docs/SA1026.html";
 
         private static readonly DiagnosticDescriptor Descriptor =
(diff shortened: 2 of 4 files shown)


vweijsters:
:+1: Looks good

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TaskEx.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Threading.Tasks;
+
+    internal static class TaskEx
```
sharwell: :exclamation: Let's call this [`SpecializedTasks`](https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/Utilities/SpecializedTasks.cs) instead.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TaskEx.cs
```diff
@@ -0,0 +1,9 @@
...
+    {
+        internal static Task CompletedTask { get; } = Task.FromResult(true);
```
sharwell: :bulb: Rather than return `Task.FromResult(true)`, return `Task.FromResult(default(VoidResult));`, by including [this `VoidResult`](https://github.com/rackerlabs/dotnet-threading/blob/master/Rackspace.Threading/VoidResult.cs) as a nested class.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeFixProvider.cs
```diff
@@ -42,22 +42,22 @@ public override FixAllProvider GetFixAllProvider()
...
 
-                SyntaxToken token = root.FindToken(diagnostic.Location.SourceSpan.Start);
-                if (token.IsMissing)
-                {
-                    continue;
-                }
-
-                context.RegisterCodeFix(CodeAction.Create(SpacingResources.SA1026CodeFix, t => GetTransformedDocumentAsync(context.Document, root, token)), diagnostic);
+                context.RegisterCodeFix(CodeAction.Create(SpacingResources.SA1026CodeFix,
+                                        token => GetTransformedDocumentAsync(context.Document, diagnostic.Location.SourceSpan.Start, token)),
```
sharwell: :bulb: Avoid aligning arguments like this. If you want to align arguments, move the first argument to the next line as well, like this:

```csharp
context.RegisterCodeFix(CodeAction.Create(
    SpacingResources.SA1026CodeFix,
    token => ...,
    ...);
```

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation.cs
```diff
@@ -49,54 +49,31 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
         }
 
-        private void HandleSyntaxTree(SyntaxTreeAnalysisContext context)
+        private static void HandleSyntaxTree(SyntaxTreeAnalysisContext context)
         {
             SyntaxNode root = context.Tree.GetCompilationUnitRoot(context.CancellationToken);
             foreach (var token in root.DescendantTokens())
             {
-                switch (token.Kind())
+                if (token.Kind() != SyntaxKind.NewKeyword)
```
sharwell: :bulb: You can also use `!token.IsKind(SyntaxKind.NewKeyword)` if you only need to check one kind.

sharwell:
:bulb: Note that I very much prefer to see changes like the use of a shared `CompletedTask` be a separate pull request. I try not to make this a hard-and-fast rule, but it definitely slows down this PR's review process.

Maxwe11:
@sharwell splitted #911

sharwell:
Thanks for taking the time to split these. Like I indicated, it wasn't required but it really does help.

:question: I see you changed the implementation of SA1026 as part of this. Were these part of a simplification, or did you find bugs/limitations that needed to be fixed?

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation.cs
```diff
@@ -49,54 +49,31 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                {
+                    continue;
+                }
 
-            if (!token.TrailingTrivia.First().IsKind(SyntaxKind.WhitespaceTrivia))
-            {
-                return;
+                SyntaxTriviaList trailingTriviaList = token.TrailingTrivia;
+                if (trailingTriviaList.Count > 0 &&
+                    (trailingTriviaList[0].IsKind(SyntaxKind.WhitespaceTrivia) || trailingTriviaList[0].IsKind(SyntaxKind.EndOfLineTrivia)))
```
Maxwe11: >I see you changed the implementation of SA1026 as part of this. Were these part of a simplification, or did you find bugs/limitations that needed to be fixed?

@sharwell most is simplification except addition in this line: `trailingTriviaList[0].IsKind(SyntaxKind.EndOfLineTrivia)`
for cases like this
```new
[] { 1, 2, 3 };```

File: StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeFixProvider.cs [hunk 6]
```diff
@@ -42,22 +42,23 @@ public override FixAllProvider GetFixAllProvider()
...
+            return Task.FromResult(true);
         }
 
-        private static Task<Document> GetTransformedDocumentAsync(Document document, SyntaxNode root, SyntaxToken token)
+        private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
         {
-            SyntaxToken corrected = token.WithoutTrailingWhitespace().WithoutFormatting();
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            SyntaxToken token = root.FindToken(diagnostic.Location.SourceSpan.Start);
+            SyntaxToken corrected = token.WithoutTrailingWhitespace(removeEndOfLineTrivia: true).WithoutFormatting();
```
Maxwe11: @sharwell and in codefix we also consider new line

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1026CodeMustNotContainSpaceAfterNewKeywordInImplicitlyTypedArrayAllocation.cs
sharwell: To clarify, unlike many diagnostics, SA1026 does not make an exception for the case where `new` is the last token on a line. The previous implementation included that exception, which you have now removed.

Is this correct?

Maxwe11: The rule doesn't says what is a whitespace exactly. 
I think `\r\n` should be considering as a whitespace, or I'm wrong?

sharwell: Based on the comment in the code fix, my statement does appear to be correct.

sharwell: Based on the wording in [SA1022](http://www.stylecop.com/docs/SA1022.html) (and others), I'm under the impression that whitespace is just space and tab. However, for this rule I don't see any problem with implementing it to also not require any new lines between the tokens.

Issue #813, opened 2015-05-11
Add "StyleCopAnalyzers Analyzers" project


sharwell:
Over time we've seen several issues arise which are specific to this particular project. It would be helpful to have a project where we can create analyzers to catch some of these situations early. This concept was initially discussed in #480.

PR #911, opened 2015-06-01
Cached CompletedTask

Maxwe11:
excluded from #908

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SpecializedTasks.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SpecializedTasks.cs
@@ -1,0 +1,16 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Threading.Tasks;
+
+    internal static class SpecializedTasks
+    {
+        internal static Task CompletedTask { get; } = Task.FromResult(default(VoidResult));
+
+        private sealed class VoidResult
+        {
+            private VoidResult()
+            {
+            }
+        }
+    }
+}
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
@@ -53,7 +53,7 @@
                 context.RegisterCodeFix(CodeAction.Create(description, cancellationToken => this.GetTransformedDocumentAsync(context.Document, diagnostic, cancellationToken)), diagnostic);
             }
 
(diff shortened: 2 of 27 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SpecializedTasks.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Threading.Tasks;
+
+    internal static class SpecializedTasks
+    {
+        internal static Task CompletedTask { get; } = Task.FromResult(default(VoidResult));
```
sharwell: The big downside of this compared to `Task.CompletedTask` is we don't create the task with `InternalTaskOptions.DoNotDispose`. If someone calls `SpecializedTasks.CompletedTask.Dispose()`, this property will become unusable until the process is restarted.

Maxwe11: `Task.CompletedTask` not supported in portable subset.
We can add one more rule for internal analyzer #813 :smile:

sharwell: > `Task.CompletedTask` not supported in portable subset.

True, but also doesn't address the concern. Has the Roslyn team indicated that they don't dispose of `Task` instances by design?

Maxwe11: Good catch. Didn't find any documentation about that except source code:
https://github.com/dotnet/roslyn/blob/918dd32ccdfd901bfd71d88bdb9f7051adfb3d51/src/Features/Core/CodeFixes/CodeFixService.cs#L219

https://github.com/dotnet/roslyn/blob/918dd32ccdfd901bfd71d88bdb9f7051adfb3d51/src/Features/Core/CodeFixes/CodeFixService.cs#L393

https://github.com/dotnet/roslyn/blob/3faf03a3e52c1b46bbc3967060cf6406d4ecc9af/src/Workspaces/Core/Portable/CodeFixes/FixAllOccurrences/BatchFixAllProvider.cs#L111

in all case they use own cached task if user's implementation of `RegisterCodeFixesAsync` returns `null`.

sharwell: :thought_balloon: That's a really interesting find.

sharwell:
This is blocked on dotnet/roslyn#3218.

sharwell:
Merging this based on the comments on the issue linked above along with a review of the implementation. The paths taken for normal async/await operations do not throw if the `Task` is both completed and disposed.

PR #918, opened 2015-06-06
Implemented SA1008 tests + codefix. Reworked SA1008 to pass the tests

### LINKED ISSUE #193: Diagnostic unit tests: SA1008 (OpeningParenthesisMustBeSpacedCorrectly)
Implement unit tests for #14.

### LINKED ISSUE #212: Code fix: OpeningParenthesisMustBeSpacedCorrectly
Implement a code fix for #14.

### LINKED ISSUE #683: SA1008 (OpeningParenthesisMustBeSpacedCorrectly) incorrectly reported for cast in foreach
The following code incorrectly results in a SA1008 warning, claiming that no space should exist before the `(` character of the cast.

```csharp
foreach (object element in (IEnumerable)collection)
{
}
```

### LINKED ISSUE #876: SA1008 (OpeningParenthesisMustBeSpacedCorrectly) is reported in for loop
The following code will report SA1008
```csharp
for (var i = leadingTrivia.Count - 1; (i >= 0) && leadingTrivia[i].IsKind(SyntaxKind.WhitespaceTrivia); i--)
```

vweijsters:
Fixes #193
Fixes #212
Fixes #683 
Fixes #876

Have a good look at this one please. I had to rework the analyzer quite a bit to make it work properly.
I verified the new analyzer code against the StyleCopAnalyzers code base and the Roslyn code base and I didn't find any weird behavior from the analyzer.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
@@ -221,5 +221,12 @@
                 return new DiagnosticResult(supportedDiagnostics.Single(i => i.Id == diagnosticId));
             }
         }
+
+        protected DiagnosticResult CSharpDiagnostic(DiagnosticDescriptor descriptor)
+        {
+            var analyzers = this.GetCSharpDiagnosticAnalyzers();
+            var supportedDiagnostics = analyzers.SelectMany(analyzer => analyzer.SupportedDiagnostics);
+            return new DiagnosticResult(supportedDiagnostics.Single(dd => dd.Equals(descriptor)));
+        }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/StyleCop.Analyzers.Test.csproj
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/StyleCop.Analyzers.Test.csproj
@@ -30,7 +30,7 @@
     <DocumentationFile>bin\Debug\StyleCop.Analyzers.Test.xml</DocumentationFile>
     <NoWarn>1591</NoWarn>
     <CodeAnalysisRuleSet>..\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <AllowUnsafeBlocks>false</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -43,7 +43,7 @@
     <DocumentationFile>bin\Release\StyleCop.Analyzers.Test.xml</DocumentationFile>
     <NoWarn>1591</NoWarn>
     <CodeAnalysisRuleSet>..\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <AllowUnsafeBlocks>false</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CodeAnalysis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/StyleCop.Analyzers.Test.csproj
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/StyleCop.Analyzers.Test.csproj
@@ -30,6 +30,7 @@
     <DocumentationFile>bin\Debug\StyleCop.Analyzers.Test.xml</DocumentationFile>
     <NoWarn>1591</NoWarn>
     <CodeAnalysisRuleSet>..\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -42,6 +43,7 @@
     <DocumentationFile>bin\Release\StyleCop.Analyzers.Test.xml</DocumentationFile>
     <NoWarn>1591</NoWarn>
     <CodeAnalysisRuleSet>..\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CodeAnalysis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
@@ -272,6 +274,7 @@
     <Compile Include="SpacingRules\SA1005UnitTests.cs" />
     <Compile Include="SpacingRules\SA1006UnitTests.cs" />
     <Compile Include="SpacingRules\SA1007UnitTests.cs" />
+    <Compile Include="SpacingRules\SA1008UnitTests.cs" />
     <Compile Include="SpacingRules\SA1021UnitTests.cs" />
     <Compile Include="SpacingRules\SA1022UnitTests.cs" />
     <Compile Include="SpacingRules\SA1024UnitTests.cs" />
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
@@ -27,17 +27,49 @@
         /// The ID for diagnostics produced by the <see cref="SA1008OpeningParenthesisMustBeSpacedCorrectly"/> analyzer.
         /// </summary>
         public const string DiagnosticId = "SA1008";
+
+        internal const string LocationKey = "location";
+        internal const string ActionKey = "action";
+        internal const string LocationPreceding = "preceding";
+        internal const string LocationFollowing = "following";
+        internal const string ActionInsert = "insert";
+        internal const string ActionRemove = "remove";
+
         private const string Title = "Opening parenthesis must be spaced correctly";
-        private const string MessageFormat = "Opening parenthesis must{0} be {1} by a space.";
         private const string Category = "StyleCop.CSharp.SpacingRules";
         private const string Description = "An opening parenthesis within a C# statement is not spaced correctly.";
         private const string HelpLink = "http://www.stylecop.com/docs/SA1008.html";
 
-        private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
+        private const string MessageNotPreceded = "Opening parenthesis must not be preceded by a space.";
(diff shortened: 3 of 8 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/StyleCop.Analyzers.Test.csproj
```diff
@@ -30,6 +30,7 @@
     <DocumentationFile>bin\Debug\StyleCop.Analyzers.Test.xml</DocumentationFile>
     <NoWarn>1591</NoWarn>
     <CodeAnalysisRuleSet>..\StyleCop.Analyzers.ruleset</CodeAnalysisRuleSet>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
sharwell: :exclamation: This shouldn't be added

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -49,158 +51,161 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                {
+                    switch (leadingTriviaList[i].Kind())
+                    {
+                        case SyntaxKind.WhitespaceTrivia:
+                            break;
+
+                        case SyntaxKind.EndOfLineTrivia:
+                            isFirstOnLine = true;
+                            done = true;
+                            return;
```
sharwell: :question: Why set the local variables only to return?

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -49,158 +51,161 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
-                if (!allowLeadingSpace && precededBySpace)
-                {
-                    // Opening parenthesis must{ not} be {preceded} by a space.
-                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, token.GetLocation(), " not", "preceded"));
-                }
-                else if (!allowLeadingNoSpace && !precededBySpace)
+                var hasLeadingComment = (leadingTriviaList.Count > 0) && leadingTriviaList.Last().IsKind(SyntaxKind.MultiLineCommentTrivia);
+                var hasLeadingSpace = (leadingTriviaList.Count > 0) && leadingTriviaList.Last().IsKind(SyntaxKind.WhitespaceTrivia);
+
+                if (!isFirstOnLine && !hasLeadingComment && (haveLeadingSpace ^ hasLeadingSpace))
```
sharwell: :bulb: Use `!=` instead of `^`.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -49,158 +51,161 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
                 {
-                    // Opening parenthesis must{} be {preceded} by a space.
-                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, token.GetLocation(), string.Empty, "preceded"));
+                    var properties = new Dictionary<string, string>
+                    {
+                        { "location", "preceding" },
```
sharwell: :bulb: Define `internal const` strings for these keys and values, and use them in the code fix. Also, consider using this:

```cs
var properties = ImmutableDictionary.Create<string, string>()
    .Add("location", "preceding")
    .Add("action", haveLeadingSpace ? "insert" : "remove");
```

Since `properties` will now be an `ImmutableDictionary<string, string>`, you can also remove the call to `ToImmutableDictionary()`.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -49,158 +51,161 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                        { "action", haveLeadingSpace ? "insert" : "remove" }
+                    };
+
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, token.GetLocation(), properties.ToImmutableDictionary(), haveLeadingSpace ? string.Empty : " not", "preceded"));
```
sharwell: :bulb: Consider using multiple `DiagnosticDescriptor` instances (similar to SA1612). The current format can't be easily localized, but if we use multiple descriptors they can each be easily translated to other languages.

StyleCop.Analyzers/StyleCop.Analyzers.Test/StyleCop.Analyzers.Test.csproj
vweijsters: This is necessary for the tests in `TestFixedStatementAsync`. If we never allow this for our test project, we will have a hard time validating any rule that touches unsafe constructions.

Besides that, I don't really see an issue with it for a test project. What is your objection exactly?

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
vweijsters: The `return` is the mistake here, it should have been a `break`. When an open parenthesis token is the first character on a line, it still needs to be validated for illegal white space following the token.

sharwell: Can you add a test which catches this case?

vweijsters: Will do

StyleCop.Analyzers/StyleCop.Analyzers.Test/StyleCop.Analyzers.Test.csproj
sharwell: The tests use unsafe constructs, but within this project they exist only as string literals that are not compiled.

> What is your objection exactly?

Specifically that we aren't using it.

vweijsters: I see your (very valid) point now. I was mixing up the project scopes in which the tests were actually executing

vweijsters:
Updated after CR

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -25,17 +27,49 @@ public class SA1008OpeningParenthesisMustBeSpacedCorrectly : DiagnosticAnalyzer
...
+            ImmutableArray.Create<DiagnosticDescriptor>()
+                .Add(DescriptorNotPreceded)
+                .Add(DescriptorPreceded)
+                .Add(DescriptorNotFollowed);
+
+        /// <summary>
+        /// Gets the diagnostic descriptor for an opening parenthesis that must not be preceded by whitespace.
+        /// </summary>
+        /// <value>The diagnostic descriptor for an opening parenthesis that must not be preceded by whitespace.</value>
+        public static DiagnosticDescriptor DescriptorNotPreceded =>
```
sharwell: :bulb: Use `{ get; } =` instead of `=>`. The way it's written here will cause a new `DiagnosticDescriptor` instance to be created every time this property is accessed.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -25,17 +27,49 @@ public class SA1008OpeningParenthesisMustBeSpacedCorrectly : DiagnosticAnalyzer
...
 
-        private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.DisabledNoTests, Description, HelpLink);
+        private const string MessageNotPreceded = "Opening parenthesis must not be preceded by a space.";
+        private const string MessagePreceded = "Opening parenthesis must be preceded by a space.";
+        private const string MessageNotFollowed = "Opening parenthesis must not be followed by a space.";
 
         private static readonly ImmutableArray<DiagnosticDescriptor> SupportedDiagnosticsValue =
-            ImmutableArray.Create(Descriptor);
+            ImmutableArray.Create<DiagnosticDescriptor>()
```
sharwell: :bulb: This is inefficient, and also unnecessary. You only need to return one descriptor per diagnostic ID. Also, the following factory method would be preferred because it only creates one array:

```cs
ImmutableArray.Create(descriptor1, descriptor2, descriptor3)
```

vweijsters:
Updated again. 

I had to introduce a static constructor because of the conversion of the properties, there is some weird initialization problem that results in null values in the immutable array otherwise.

I'm also still exporting all descriptors, as the test code modification that I made to check for the correct descriptor requires it.

sharwell:
> I had to introduce a static constructor because of the conversion of the properties, there is some weird initialization problem that results in null values in the immutable array otherwise.

Initializers are always run top to bottom. When the immutable array was created, the properties returned their default value null because their initializers were below the call to `ImmutableArray.Create`.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -25,17 +26,51 @@ public class SA1008OpeningParenthesisMustBeSpacedCorrectly : DiagnosticAnalyzer
...
+
+        private static readonly ImmutableArray<DiagnosticDescriptor> SupportedDiagnosticsValue;
+
+        static SA1008OpeningParenthesisMustBeSpacedCorrectly()
+        {
+            DescriptorNotPreceded = new DiagnosticDescriptor(DiagnosticId, Title, MessageNotPreceded, Category, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
+            DescriptorPreceded = new DiagnosticDescriptor(DiagnosticId, Title, MessagePreceded, Category, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
+            DescriptorNotFollowed = new DiagnosticDescriptor(DiagnosticId, Title, MessageNotFollowed, Category, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
+
+            SupportedDiagnosticsValue = ImmutableArray.Create(DescriptorNotPreceded, DescriptorPreceded, DescriptorNotFollowed);
```
sharwell: :bulb: This only needs to return one of the three. They all have the same diagnostic ID so it does not matter which of the three gets returned.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -221,5 +221,12 @@ protected DiagnosticResult CSharpDiagnostic(string diagnosticId = null)
                 return new DiagnosticResult(supportedDiagnostics.Single(i => i.Id == diagnosticId));
             }
         }
+
+        protected DiagnosticResult CSharpDiagnostic(DiagnosticDescriptor descriptor)
+        {
+            var analyzers = this.GetCSharpDiagnosticAnalyzers();
```
sharwell: I updated this method during the merge to simply do this:

```cs
return this.CSharpDiagnostic(descriptor.Id).WithMessageFormat(descriptor.MessageFormat);
```

PR #927, opened 2015-06-17
SA1206 rule implementation

### LINKED ISSUE #66: SA1206: DeclarationKeywordsMustFollowOrder
http://www.stylecop.com/docs/SA1206.html

baks:
Fixes #66 

This is implementation of SA1206 rule

Implementation return diagnostics as orginal StyleCop.
But there are a slight differences :
* location of diagnostic is set to location of modifier, instead of start of line containing declaration
* if there are more than one access modifier diagnostics are reported seperately for those access modifiers

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1206DeclarationKeywordsMustFollowOrder.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1206DeclarationKeywordsMustFollowOrder.cs
@@ -2,6 +2,8 @@
 {
     using System.Collections.Immutable;
     using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
 
     /// <summary>
@@ -31,16 +33,39 @@
         /// </summary>
         public const string DiagnosticId = "SA1206";
         private const string Title = "Declaration keywords must follow order";
-        private const string MessageFormat = "TODO: Message format";
+        private const string MessageFormat = "The {0} keyword must come before the '{1}' keyword in the element declaration.";
         private const string Category = "StyleCop.CSharp.OrderingRules";
         private const string Description = "The keywords within the declaration of an element do not follow a standard ordering scheme.";
         private const string HelpLink = "http://www.stylecop.com/docs/SA1206.html";
 
         private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.DisabledNoTests, Description, HelpLink);
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
 
         private static readonly ImmutableArray<DiagnosticDescriptor> SupportedDiagnosticsValue =
             ImmutableArray.Create(Descriptor);
+
+        /// <summary>
+        /// Represents modifier type for implementing SA1206 rule
+        /// </summary>
+        private enum ModifierType
+        {
+            /// <summary>
+            /// Represents default value
+            /// </summary>
+            None,
+            /// <summary>
+            /// Represents any of access modifiers i.e public, protected, internal, private
+            /// </summary>
+            Access,
+            /// <summary>
+            /// Represents static modifier
+            /// </summary>
+            Static,
+            /// <summary>
+            /// Represents other modifiers i.e partial, virtual, abstract, override, extern, unsafe, new, async, const, sealed, readonly, volatile, fixed
+            /// </summary>
+            Other
+        }
 
         /// <inheritdoc/>
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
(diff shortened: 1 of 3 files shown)


sharwell:
> location of diagnostic is set to location of modifier, instead of start of line containing declaration

sharwell:
:question: Did you follow the same policy we decided on for `using` declarations? Specifically:

> Report a warning on the current modifier, if the current modifier should be located after the next modifier.

baks:
>:question: Can you point me to an example in the tests of what you mean by this?

For example in `SA1206UnitTests.cs` on line `62` diagnostics are reported for `protected` and `internal` access modifier.

sharwell:
For the following declaration:

```csharp
virtual new internal async protected void FifthMethod() { }
```

If you follow the rule I mentioned, the warnings would appear on:

* `new`, because it should appear after `internal`
* `async`, because it should appear after `protected`

:thought_balloon: I only bring this up because I like the idea of having a consistent *simple* rule to determine which items to report warnings on when checking the order (or in this case, partial order) of a list of items. We can then use the same rule on many different types of analyzers, including `using` declarations and this one.

baks:
@sharwell there would be one more warning if analyzer follow the rule that you've mentioned:
* `virtual` because it should appear after `internal`/`protected`

Now the warnings appear on: (because I didn't follow the rule that you've described)
* `internal` because it should come before `other` keywords
* `protected` because it should come before `other` keywords

So, when I was implementing that rule I followed StyleCop diagnostics messages, but you've also a good point.  I've feeling that it's better to report diagnostics on access modifiers because it teaches that the access modifiers should go first (but this can be only my feeling :wink:). But for the consistency it's better to use the same rule.

:question: So, we wait for somebody else to look at this, or you confirm that I should change how SA1206 works?

sharwell:
`virtual` appears before `new` in that example, which are two items in the same category so (for this analyzer) they are unordered with respect to each other. The rule only looks at an item and the immediately following item since that can be readily implemented in O(n) time and O(1) space.

sharwell:
> :question: So, we wait for somebody else to look at this, or you confirm that I should change how SA1206 works?

Let me look back over the current state of the `using` declaration analyzers. I do think we should come up with a single simple rule that applies to all ordering analyzers - either what I described or the opposite (report on current if current should appear before previous).

baks:
>virtual appears before new in that example, which are two items in the same category so (for this analyzer) they are unordered with respect to each other. The rule only looks at an item and the immediately following item since that can be readily implemented in O(n) time and O(1) space.

This cause two additionals run of analyzer to get all the diagnostics. 
If we fix the warnings that we get from rule you've described we get:
`virtual internal new protected async`
So, there we would get diagnostics for `virtual` and for `new`. Then we fix it and get
`internal virtual protected new async` and now we would get diagnostic for `virtual` and finally after fixing it we get correct order:
`internal protected virtual new async`

Ordering modifiers of a declaration looks quite easy compared to ordering `using` directives and while I think more about the StyleCop way I like it more.

sharwell:
> This cause two additional runs of analyzer to get all the diagnostics.

The code fix would be implemented to move as few items as necessary so that SA1206 is not reported in the result. One pass of the code fix on the above would give the following:

```csharp
internal protected virtual new async void FifthMethod() { }
```

baks:
The diagnostics should be also for developers, so the developer could learn how to correctly order modifiers. And if analyzer will be reporting incomplete diagnostics it take more time for unfamiliar developers with this rule.
I think in case of modifiers there is a reason to report all diagnostics at once. And in actual implementation I don't see that it adds complexity to implemenation.

That's my point, but if I'm wrong let me know.

sharwell:
I'm going to look at this tomorrow with the intent to merge as-is. We can create a separate issue to review the ordering rules across the various analyzers.

sharwell:
:bulb: You included a commit to merge master into your working branch. These commits are rarely required, and when included it causes the pull request to use two merges instead of just one. If AppVeyor isn't reporting a failure, you probably don't need the merge. :+1:

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1206DeclarationKeywordsMustFollowOrder.cs
```diff
@@ -31,17 +33,40 @@ public class SA1206DeclarationKeywordsMustFollowOrder : DiagnosticAnalyzer
         /// </summary>
         public const string DiagnosticId = "SA1206";
         private const string Title = "Declaration keywords must follow order";
-        private const string MessageFormat = "TODO: Message format";
+        private const string MessageFormat = "The {0} keyword must come before the '{1}' keyword in the element declaration.";
```
sharwell: :bulb: In general, the `'` characters are used to mark sections of the message which are replaceable. In this case, I would have expected the following:

```
The '{0}' keyword must come before the '{1}' keyword in the element declaration.
```

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1206DeclarationKeywordsMustFollowOrder.cs
```diff
@@ -54,7 +79,177 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+            }
+            else if (first == ModifierType.Other && (second == ModifierType.Static || second == ModifierType.Access))
+            {
+                result = greaterThan;
+            }
+
+            return result;
+        }
+
+        private static string GetModifierTypeText(ModifierType modifierType)
```
sharwell: :bulb: In practice, I found this had two problems:

1. It makes the messages not localizable, because other languages don't have the same syntax.
2. It makes it more challenging for the reader to know what to move. If it said `internal` must appear before `static`, there would be no question.

I'll file a bug to correct this.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1206DeclarationKeywordsMustFollowOrder.cs
```diff
@@ -54,7 +79,177 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
+                    break;
+                case ModifierType.Other:
+                    result = "other";
+                    break;
+            }
+
+            return result;
         }
+
+        private static bool AccessOrStaticModifierNotFollowingOtherModifier(ModifierType current, ModifierType previous) => (current == ModifierType.Access || current == ModifierType.Static) && previous != ModifierType.Other;
```
sharwell: If the expression is more than one token, this syntax will probably make readability *worse*. I can't tell where the signature ends and the body begins when I'm glancing at this.

PR #930, opened 2015-06-25
Fix SA1119 for rc3

### LINKED ISSUE #929: Update SA1119 (StatementMustNotUseUnnecessaryParenthesis) for new API restrictions
SA1119 relies on a hidden diagnostic which is not included in `SupportedDiagnostics` to gray out unnecessary parentheses. Starting with the RC3 release of Roslyn, the pattern used here is no longer supported. This analyzer needs to be updated to the following pattern to make it compatible with (presumably) upcoming releases of Visual Studio 2015:

1. Include the hidden diagnostic in the `SupportedDiagnostics`
2. Enable the hidden diagnostic by default
3. Update `Initialize` to only register a compilation start action. In the callback, check to see if SA1119 is not suppressed before registering other callbacks. Now that the parentheses diagnostic is always enabled, this analyzer will always be active during builds. We can eliminate essentially all overhead from this by conditionally registering its callbacks in the compilation start handler.

pdelvo:
Fixes #929. I verified that this works in rc3 as well as rc2. I would like to implement tests that verify that the parenthesis diagnostic is not reported if the main diagnostic is suppressed, but I need #522 for that.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
@@ -64,5 +64,24 @@ namespace StyleCop.Analyzers
                 },
                 syntaxKinds);
         }
+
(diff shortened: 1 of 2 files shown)


sharwell:
:+1:

PR #934, opened 2015-06-29
Improve SA1113 implementation

Maxwe11:
- avoid code duplication (removed `CheckIfCommasAreAtTheSameLineAsThePreviousParameter` overloads)
- replace `List<T>` with `SyntaxNodeOrTokenList` which is value type and probably should result in less heap allocations

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1113CommaMustBeOnSameLineAsPreviousParameter.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1113CommaMustBeOnSameLineAsPreviousParameter.cs
@@ -1,12 +1,11 @@
 ﻿namespace StyleCop.Analyzers.ReadabilityRules
 {
-    using System.Collections.Generic;
     using System.Collections.Immutable;
-    using System.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.Helpers;
 
     /// <summary>
     /// A comma between two parameters in a call to a C# method or indexer, or in the declaration of a method or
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1113CommaMustBeOnSameLineAsPreviousParameter.cs
```diff
@@ -88,361 +88,239 @@ private void HandleArrayDeclaration(SyntaxNodeAnalysisContext context)
...
-                .ChildTokens()
-                .Where(t => t.IsKind(SyntaxKind.CommaToken))
-                .ToList();
-
-            var argumentListSyntax = invocationEpression.ArgumentList;
-            CheckIfCommasAreAtTheSameLineAsThePreviousParameter(context, commas, argumentListSyntax);
         }
 
-        private void HandleInvoationExpression(SyntaxNodeAnalysisContext context)
+        private static void HandleInvoationExpression(SyntaxNodeAnalysisContext context)
```
vweijsters: There's a typo in the method name

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1113CommaMustBeOnSameLineAsPreviousParameter.cs
```diff
@@ -88,361 +88,239 @@ private void HandleArrayDeclaration(SyntaxNodeAnalysisContext context)
...
-        private static void CheckIfCommasAreAtTheSameLineAsThePreviousParameter(SyntaxNodeAnalysisContext context, List<SyntaxToken> commas,
-            AttributeListSyntax attributeListSyntax)
+        private static void CheckIfCommasAreAtTheSameLineAsThePreviousParameter(SyntaxNodeAnalysisContext context, SyntaxNodeOrTokenList nodeOrTokenList)
         {
-            for (int index = 0; index < commas.Count; index++)
+            // If flag is even we expecting parameter syntax node, otherwise we expecting comma token.
+            int index = 0;
+            SyntaxNode previousNode = null;
+
+            foreach (var nodeOrToken in nodeOrTokenList)
```
vweijsters: As an index is needed, it would probably be better to use a `for` loop here, instead of `foreach`. Especially since `SyntaxNodeOrTokenList` supports an indexer

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1113CommaMustBeOnSameLineAsPreviousParameter.cs
```diff
@@ -88,361 +88,239 @@ private void HandleArrayDeclaration(SyntaxNodeAnalysisContext context)
...
+                    if (nodeOrToken.IsNode)
+                    {
+                        return;
+                    }
+
+                    // previousNode not set
+                    if (previousNode == null)
+                    {
+                        return;
+                    }
```
vweijsters: This check is superfluous. It will never trigger, unless there is a bug in Roslyn.

vweijsters:
There is additional room for optimization:

```c#
    if (argumentList != null && !argumentList.IsMissing)
    {
        var arguments = argumentList.Arguments;
        if (arguments.Count > 1)
        {
            CheckIfCommasAreAtTheSameLineAsThePreviousParameter(context, arguments.GetWithSeparators());
        }
    }
```

and 

```c#
    if (parameterList != null && !parameterList.IsMissing)
    {
        var parameters = parameterList.Parameters;
        if (parameters.Count > 1)
        {
            CheckIfCommasAreAtTheSameLineAsThePreviousParameter(context, parameters.GetWithSeparators());
        }
    }
```

are each occurring multiple times and can therefore be put in a method.

Maxwe11:
updated

vweijsters:
:+1: Looks good to me

Maxwe11:
squashed

PR #935, opened 2015-06-29
Improved implementation of VerifyFixAsync that will ignore diagnostics…

vweijsters:
… for which code fixes are not available.


vweijsters:
Running this will result in 16 failed tests. See also #936

sharwell:
The behavior of this change does not match the behavior of the IDE for the same code fixes. While the code fix creates separate `XmlTextSyntax` nodes for the newline and the following summary text, the IDE merges these two into a single `XmlTextSyntax` with two tokens (`XmlTextLiteralNewLineToken` and `XmlTextLiteralToken`) before invoking the analyzers again.

vweijsters:
I implemented a workaround for #936. It is my opinion that there was nothing wrong with the code to begin with. The fact that the IDE merges two tokens is an implementation detail of Roslyn and our code should not depend on that. 

I added the workaround to removing a blocking factor for this PR. Given the complexity of the code affected by #936, it seemed a better solution to me to add a workaround for now.

sharwell:
> I added the workaround to removing a blocking factor for this PR.

At first glance, the workaround is exactly what I had in mind. :smile: 

> It is my opinion that there was nothing wrong with the code to begin with.

:thought_balloon: In an ideal world, the analyzers would treat equivalent representations of the source code in an equivalent manner. I'm comfortable knowing that we have regression tests in place to alert us to any change in the representation of the comments provided by Roslyn that would impact the behavior of the analyzers.

sharwell:
That was a big improvement. Commit e0304de took [8:17 minutes](https://ci.appveyor.com/project/sharwell/stylecopanalyzers/build/1.0.838) on AppVeyor, but the updated commit bb9cee1 to [only 6:50 minutes](https://ci.appveyor.com/project/sharwell/stylecopanalyzers/build/1.0.853). It's still slower than prior to this pull request but not as much slower.

Hopefully performance improvements to xunit can improve this more in the future.

PR #938, opened 2015-07-01
Fix SA1508

Maxwe11:
SA1508 throws ArgumentOutOfRange


pdelvo:
Could you include a test for this?

vweijsters:
Please add a test case that will reproduce the issue.

Maxwe11:
Will try, but I'm not sure what's going wrong. It was happen in moment of typing multiline comment

    public void Bar(string s, int i, int i2) /* */
    {
        var ints = new[] { 1, 2, 3 };
    }

PR #939, opened 2015-07-01
Introduce GetReplacementToken helper for Dictionary

Maxwe11:
This helper method allow us avoid closure allocations when used with SyntaxNode.ReplaceTokens and etc.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DictionaryHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/DictionaryHelper.cs
@@ -1,0 +1,23 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Collections.Generic;
+    using Microsoft.CodeAnalysis;
+
+    internal static class DictionaryHelper
+    {
+        /// <summary>
+        /// This helper method allow us avoid closure allocations when used with SyntaxNode.ReplaceTokens and etc.
+        /// </summary>
+        /// <param name="replacements">Dictionary with key value pairs {token to replace, replacement token}.</param>
+        /// <param name="original">The token to replace. This argument used for getting replacement token from dictionary.</param>
+        /// <param name="couldBeRewritten">The token to replace which could be rewritten. We ignore this argument.</param>
+        /// <returns>Returns replacement token from dicionary.</returns>
+        internal static SyntaxToken GetReplacementToken(
+            this Dictionary<SyntaxToken, SyntaxToken> replacements,
+            SyntaxToken original,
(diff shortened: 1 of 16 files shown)


sharwell:
This still requires the use of an uncached delegate with a closure. At first read, I believe it performs exactly the same number of allocations as the original code.

Also, I *really* don't like adding extension methods to well-known built-in types. It doesn't mean I wouldn't ever do it, but I'm certainly looking for strong empirical evidence that it's providing improved performance.

Maxwe11:
>This still requires the use of an uncached delegate with a closure. At first read, I believe it performs exactly the same number of allocations as the original code.

I've checked IL, there is no closure :smile:

Maxwe11:
Original code requires 3 allocation:
1. dictionary with replacements
2. delegate allocation
3. closure

With this change we have only:
1. dictionary with replacements
2. delegate allocation

sharwell:
I don't understand why we should need to deal with this ourselves, so I filed dotnet/roslyn#3774. :smile:

sharwell:
I would prefer this pull request be rewritten using [this approach](https://github.com/dotnet/roslyn/blob/0382e3e3fc543fc483090bff3ab1eaae39dfb4d9/src/Features/Core/CodeFixes/Suppression/AbstractSuppressionCodeFixProvider.PragmaWarningCodeAction.cs#L61) instead of the current one. In addition to avoiding the extension method on `Dictionary<SyntaxToken, SyntaxToken>`, it provides additional memory savings my using simple arrays instead of a full hash map.

Maxwe11:
That approach still require 3 allocations. Moreover ReplaceTokens use HashSet under hood. Seems the best what we can do - CustomRewriter.We already have a couple:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/a45d579c4188e957234512336d679dde2e4984dc/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1502CodeFixProvider.cs#L194

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/a45d579c4188e957234512336d679dde2e4984dc/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CodeFixProvider.cs#L232

sharwell:
Or `DocumentEditor`: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/f977394726a25a3f8ff4b2b6e5934aebba2a74b9/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1107FixAllProvider.cs#L65

Maxwe11:
closing as `wontfix`

PR #941, opened 2015-07-01
Sa1205 fix

### LINKED ISSUE #937: SA1205 (PartialElementsMustDeclareAccess) code fix inserts modifier at the wrong location
The code fix is placingthe modifier before xml documentation.

Im fixing that

pdelvo:
Fixes #937.

SA1400 already had a method which did the right thing so I just put it into a helper class. I also added a regression test.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TypeDeclarationSyntaxHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TypeDeclarationSyntaxHelper.cs
@@ -1,0 +1,71 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Linq;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+
+    internal class TypeDeclarationSyntaxHelper
+    {
+        /// <summary>
+        /// Adds a modifier token for <paramref name="modifierKeyword"/> to the beginning of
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TypeDeclarationSyntaxHelper.cs
```diff
@@ -0,0 +1,71 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Linq;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+
+    internal class TypeDeclarationSyntaxHelper
```
sharwell: :bulb: `static class`

:question: SA1400 uses this for updating modifiers of both types and members. Should it have a different name?

pdelvo: You're right. Do you have a suggestion?

sharwell: `DeclarationModifiersHelper`?

pdelvo: fixed

pdelvo:
I implemented the PR feedback

Issue #923, opened 2015-06-13
SA1115 (ParameterMustFollowComma) incorrectly reported after multiline parameter


sharwell:
SA1115 should not be reported for the following statement. However, I am currently seeing it reported 4 times.

```csharp
return Nfa.Choice(
    Nfa.Sequence(
        Nfa.Match(ANTLRParser.TOKEN_REF),
        Nfa.Optional(Nfa.Rule(Bindings.ElementOptions)),
        Nfa.Optional(Nfa.Match(ANTLRParser.ARG_ACTION))),
    Nfa.Match(ANTLRParser.RULE_REF),
    Nfa.Sequence(
        Nfa.Match(ANTLRParser.CHAR_LITERAL),
        Nfa.Optional(Nfa.Rule(Bindings.ElementOptions))),
    Nfa.Sequence(
        Nfa.Match(ANTLRParser.STRING_LITERAL),
        Nfa.Optional(Nfa.Rule(Bindings.ElementOptions))),
    Nfa.Sequence(
        Nfa.Match(ANTLRParser.DOLLAR),
        Nfa.Rule(Bindings.Label)),
    Nfa.Match(ANTLRParser.ACTION));
```

pdelvo:
I think this issue can be closed

PR #942, opened 2015-07-03
Fix for #923

oatkins:
Added new tests for issue #923 and a proposed fix.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1115ParameterMustFollowComma.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1115ParameterMustFollowComma.cs
@@ -132,17 +132,18 @@
                 return;
             }
 
-            var previousLine = attributeList.Attributes[0].GetLocation().GetLineSpan().StartLinePosition.Line;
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1115ParameterMustFollowComma.cs
```diff
@@ -156,17 +157,18 @@ private void HandleAttribute(SyntaxNodeAnalysisContext context)
...
+            var previousLine = attribute.ArgumentList.Arguments[0].GetLocation().GetLineSpan().EndLinePosition.Line;
             for (int i = 1; i < attribute.ArgumentList.Arguments.Count; i++)
             {
                 var currentArgument = attribute.ArgumentList.Arguments[i];
-                var currentLine = currentArgument.GetLocation().GetLineSpan().StartLinePosition.Line;
+                var lineSpan = currentArgument.GetLocation().GetLineSpan();
+                var currentLine = lineSpan.StartLinePosition.Line;
                 if (currentLine - previousLine > 1)
                 {
                     context.ReportDiagnostic(Diagnostic.Create(Descriptor, currentArgument.GetLocation()));
```
Maxwe11: Could you cache `currentArgument.GetLocation()` into local variable above and use it here?

PR #945, opened 2015-07-07
Add a equivalence key to all code actions

pdelvo:
A roslyn analyzer will complain in rc3 if the equivalence key is not set. I set it to the concatenation of all diagnostic ids the code fix can fix. I don't think we have a code fix where we want to do it differently.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
@@ -50,7 +50,7 @@
                 }
 
                 string description = DocumentationResources.SA1609SA1610CodeFix;
-                context.RegisterCodeFix(CodeAction.Create(description, cancellationToken => this.GetTransformedDocumentAsync(context.Document, diagnostic, cancellationToken)), diagnostic);
+                context.RegisterCodeFix(CodeAction.Create(description, cancellationToken => this.GetTransformedDocumentAsync(context.Document, diagnostic, cancellationToken), equivalenceKey: nameof(SA1609SA1610CodeFixProvider)), diagnostic);
             }
 
             return SpecializedTasks.CompletedTask;
--- a/StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SX1309CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SX1309CodeFixProvider.cs
@@ -55,7 +55,7 @@
                 if (!string.IsNullOrEmpty(token.ValueText))
                 {
                     string newName = '_' + token.ValueText;
-                    context.RegisterCodeFix(CodeAction.Create(string.Format(NamingResources.SX1309CodeFix, newName), cancellationToken => RenameHelper.RenameSymbolAsync(document, root, token, newName, cancellationToken)), diagnostic);
+                    context.RegisterCodeFix(CodeAction.Create(string.Format(NamingResources.SX1309CodeFix, newName), cancellationToken => RenameHelper.RenameSymbolAsync(document, root, token, newName, cancellationToken), equivalenceKey: nameof(SA1309CodeFixProvider)), diagnostic);
                 }
             }
         }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1103CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1103CodeFixProvider.cs
@@ -44,10 +44,10 @@
 
                 if (queryExpression.DescendantTrivia().All(AcceptableSingleLineTrivia))
                 {
-                    context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixSingleLine, token => GetTransformedDocumentFromSingleLineAsync(context.Document, diagnostic, token)), diagnostic);
+                    context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixSingleLine, token => GetTransformedDocumentFromSingleLineAsync(context.Document, diagnostic, token), equivalenceKey: nameof(SA1103CodeFixProvider)), diagnostic);
                 }
 
-                context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixMultipleLines, token => GetTransformedDocumentForMultipleLinesAsync(context.Document, diagnostic, token)), diagnostic);
+                context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixMultipleLines, token => GetTransformedDocumentForMultipleLinesAsync(context.Document, diagnostic, token), equivalenceKey: nameof(SA1103CodeFixProvider)), diagnostic);
             }
         }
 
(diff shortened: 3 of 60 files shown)


sharwell:
:question: What do you mean "will complain"?

pdelvo:
Roslyn has a new rule in its analyzers that checks if the CodeAction.Create invokation is providing an equivalence key. https://github.com/dotnet/roslyn/blob/cbd066333305ef9baed06035db81717ac8f32acb/src/Diagnostics/CodeAnalysis/Core/FixAnalyzers/FixerWithFixAllAnalyzer.cs#L312

sharwell:
It looks like the equivalence key only needs to be specified in cases where we associate more than one code fix with a single diagnostic. Do we ever do this?

pdelvo:
Not yet. So should we just disable the warning?

sharwell:
I'm thinking yes, and possibly file a bug. From the look of dotnet/roslyn#3093, this diagnostic shouldn't even be reported for our code (because we never produce multiple code fixes for a single diagnostic).

sharwell:
I'm going to close this for now. When the warnings show up after the update to RC3 (or later), we can look back at it then and try to find a less "ugly" alternative to this.

vweijsters:
The SA1103CodeFixProvider does actually register two possible fixes for the same diagnostics in some cases.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609SA1610CodeFixProvider.cs
```diff
@@ -50,7 +50,8 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
                 }
 
                 string description = DocumentationResources.SA1609SA1610CodeFix;
-                context.RegisterCodeFix(CodeAction.Create(description, cancellationToken => this.GetTransformedDocumentAsync(context.Document, diagnostic, cancellationToken)), diagnostic);
+                context.RegisterCodeFix(CodeAction.Create(description, cancellationToken => this.GetTransformedDocumentAsync(context.Document, diagnostic, cancellationToken),
+                    equivalenceKey: SA1619GenericTypeParametersMustBeDocumentedPartialClass.DiagnosticId + SA1610PropertyDocumentationMustHaveValueText.DiagnosticId), diagnostic);
```
sharwell: :question: Can we get away with just using `string.Empty` for all cases where we only register one fix for each diagnostic? The advantage of this as a convention is it will be immediately obvious looking at the call to `CodeAction.Create` whether we intend to provide only one fix or multiple potential fixes.

:bulb: If not `string.Empty`, at least simplify the pattern to just `nameof(SA1609SA1610CodeFixProvider)` as the equivalence key.

sharwell:
Reopening, but we need to find a solution with less ugly.

pdelvo:
I changed all the key to usethe name of the code fix instead

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1011CodeFixProvider.cs
```diff
@@ -51,7 +51,7 @@ public override FixAllProvider GetFixAllProvider()
                     continue;
                 }
 
-                context.RegisterCodeFix(CodeAction.Create(SpacingResources.SA1011CodeFix, t => GetTransformedDocumentAsync(context.Document, root, token)), diagnostic);
+                context.RegisterCodeFix(CodeAction.Create(SpacingResources.SA1011CodeFix, t => GetTransformedDocumentAsync(context.Document, root, token), equivalenceKey: nameof(SA1011ClosingSquareBracketsMustBeSpacedCorrectly)), diagnostic);
```
sharwell: :bulb: `nameof(SA1011CodeFixProvider)`

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1103CodeFixProvider.cs
```diff
@@ -44,10 +44,10 @@ public override FixAllProvider GetFixAllProvider()
 
                 if (queryExpression.DescendantTrivia().All(AcceptableSingleLineTrivia))
                 {
-                    context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixSingleLine, token => GetTransformedDocumentFromSingleLineAsync(context.Document, diagnostic, token)), diagnostic);
+                    context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixSingleLine, token => GetTransformedDocumentFromSingleLineAsync(context.Document, diagnostic, token), equivalenceKey: nameof(SA1103CodeFixProvider)), diagnostic);
                 }
 
-                context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixMultipleLines, token => GetTransformedDocumentForMultipleLinesAsync(context.Document, diagnostic, token)), diagnostic);
+                context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixMultipleLines, token => GetTransformedDocumentForMultipleLinesAsync(context.Document, diagnostic, token), equivalenceKey: nameof(SA1103CodeFixProvider)), diagnostic);
```
sharwell: :bulb: `+ "Multiple"`

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1103CodeFixProvider.cs
```diff
@@ -44,10 +44,10 @@ public override FixAllProvider GetFixAllProvider()
 
                 if (queryExpression.DescendantTrivia().All(AcceptableSingleLineTrivia))
                 {
-                    context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixSingleLine, token => GetTransformedDocumentFromSingleLineAsync(context.Document, diagnostic, token)), diagnostic);
+                    context.RegisterCodeFix(CodeAction.Create(ReadabilityResources.SA1103CodeFixSingleLine, token => GetTransformedDocumentFromSingleLineAsync(context.Document, diagnostic, token), equivalenceKey: nameof(SA1103CodeFixProvider)), diagnostic);
```
sharwell: :bulb: `+ "Single"`

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SX1309CodeFixProvider.cs
```diff
@@ -55,7 +55,7 @@ public override FixAllProvider GetFixAllProvider()
                 if (!string.IsNullOrEmpty(token.ValueText))
                 {
                     string newName = '_' + token.ValueText;
-                    context.RegisterCodeFix(CodeAction.Create(string.Format(NamingResources.SX1309CodeFix, newName), cancellationToken => RenameHelper.RenameSymbolAsync(document, root, token, newName, cancellationToken)), diagnostic);
+                    context.RegisterCodeFix(CodeAction.Create(string.Format(NamingResources.SX1309CodeFix, newName), cancellationToken => RenameHelper.RenameSymbolAsync(document, root, token, newName, cancellationToken), equivalenceKey: nameof(SA1309CodeFixProvider)), diagnostic);
```
sharwell: :bulb: `nameof(SX1309CodeFixProvider)`

sharwell:
In addition to the mentioned items, during the merge I addressed several diagnostics which were omitted from this pull request.

* SA1022CodeFixProvider
* SA1024CodeFixProvider
* SA1026CodeFixProvider
* SA1027CodeFixProvider
* SA1102CodeFixProvider
* SA1207CodeFixProvider
* OpenCloseSpacingCodeFixProvider

PR #947, opened 2015-07-10
Implement a fix all provider for removing region directives

pdelvo:
While testing I discovered that the BatchFixer screws up when multiple regions are close together.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1123UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1123UnitTests.cs
@@ -41,6 +41,7 @@
 }";
 
             await this.VerifyCSharpFixAsync(testCode, fixedCode).ConfigureAwait(false);
+            await this.VerifyCSharpFixAllFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
         }
 
         [Fact]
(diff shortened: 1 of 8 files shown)


sharwell:
I think we should hold off on this one until after Visual Studio 2015 RTM. There were some improvements to the batch fixer since the RC, in particular it was updated so if two code fixes conflict then one is applied and the other will not be. This means it might not fix every last instance, but it will certainly be working towards fixing all instances.

Considering we aren't currently testing the fix all providers, I see two passes of a well-tested code fix as preferable to an untested single-pass code fix.

sharwell:
Marking as blocked on Visual Studio 2015 RTM.

sharwell:
@pdelvo Is this still needed?

pdelvo:
The batch fixer does not unbalance the directives any more. But when im thinking of "fix all occurences" I would think that it would fix all occurences. And the batch fixer does not do that and thats annoying. This fix all provider fixes all occurences.

sharwell:
I like your document-based fix all provider. I think it could be used to improve the efficiency of several fixers (eventually).

Let's get the test capabilities for fix all providers merged and then update this pull request to include tests before we merge it.

pdelvo:
I fixed the merge conflict

FrankBakkerNl:
Looking for how to implement a FixAllProvider for my own project I found this solution here which solves is very nicely. I was wondering why in GetFixAsync you are using Task.FromResult(newSolution)

If I read this correct it means the new solution will be created before the task returned by GetFixAsync completes. I do not think this is how this API is intended to be used, in stead you should create a CodeAction which will perform the fix on demand when it gets invoked.

If you extract the for loop into another async method you do not need to await the actual execution of the fix inside GetFixAsync. This way the fix will not be executed before the codeaction is executed, it also allows you to propagate the Cancellation token to any async method you use inside the fix.

Based on the work here, I created a (non generic) FixAllProvider for my own project. check it out here 
https://github.com/FrankBakkerNl/DefaultIfNullAnalyser/blob/master/DefaultIfNullAnalyzer/DefaultIfNullAnalyzer/DefaultIfNullFixAllProvider.cs.

sharwell:
@FrankBakkerNl Thanks for catching this. I opened up issue #1411 to correct it.

PR #951, opened 2015-07-12
Add a test that verifies that SA1119's secondary diagnostic is not fired when the main diagnostic is disabled

pdelvo:


vweijsters:
:+1: Looks good to me.

Would be nice to have an extension method that will disable one or more diagnostics for a given project within a solution. Something like `void SuppressDiagnostics(this Solution solution, ProjectId projectId, params string[] diagnosticIdentifiers)`

PR #954, opened 2015-07-13
Implement a code fix for SA1402

### LINKED ISSUE #949: Code fix: SA1402 (FileMayOnlyContainASingleClass)
Implement a code fix for #91.

sharwell:
Fixes #949


pdelvo:
I just tried it. It works, but it should probably correct the whitespaces before the new class. Otherwise I get SA1505 on the opening curly bracket of the namespace

sharwell:
@pdelvo It also doesn't optimize using directives. I would prefer leaving the correction of other violations for a separate change. I'm concerned it adds quite a bit of complexity to the implementation and I'm not sure the current output will be problematic for users. We can always fix this later. :smile:

pdelvo:
This is definitly better then not having a code fix

pdelvo:
LGTM :shipit:

PR #961, opened 2015-07-14
Initial StyleCopTester implementation

### LINKED ISSUE #952: Command line tool to run StyleCopAnalyzers on a solution
It would be nice to have a command line utility that runs the StyleCopAnalyzers analyzers on a solution (code fixes considered later).

```
StyleCopTester [options] <Solution>
```

By default, all analyzers which are enabled by default run. The following options are avaliable for `[options]`:

| Option | Meaning |
| --- | --- |
| `/all` | Run all StyleCopAnalyzers analyzers, including ones that are disabled by default |
| `/id:<id>` | Enable analyzer with diagnostic ID `<id>` (when this is specified, only this analyzer is enabled) |

:memo: A primary purpose of this tool is testing performance with both the sampling and instrumenting profilers. The Test Explorer in Visual Studio 2015 does not currently provide enough ability to configure profiling methods for unit tests, and running inside the IDE introduces significant variance due to the editor environment, hence the need to create this as a command line tool.

pdelvo:
This implementation is currently not really usefull. When certain analyzers are included in the list of analyzers that are used it will not find any diagnostics. This bug is fixed in roslyn versions post rc2. Also currently there isnt really much validation of the command line parameters.

Do we want tests for this?

Fixes #952

### DIFF
--- a/StyleCop.Analyzers/StyleCopTester/Program.cs
+++ b/StyleCop.Analyzers/StyleCopTester/Program.cs
@@ -1,0 +1,129 @@
+﻿namespace StyleCopTester
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Reflection;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Microsoft.CodeAnalysis.MSBuild;
+
+    /// <summary>
+    /// StyleCopTester is a tool that will analyze a solution, find diagnostics in it and will print out the number of
+    /// diagnostics it could find. This is useful to easily test performance without having the overhead of visual
+    /// studio running.
+    /// </summary>
+    internal class Program
+    {
+        private static void Main(string[] args)
+        {
+            // A valid call must have at least two parameters. A solution file and /all or one /id:SAXXXX
+            if (args.Length <= 1)
+            {
+                PrintHelp();
+            }
+            else
+            {
+                Stopwatch stopwatch = Stopwatch.StartNew();
+                var analyzers = GetAllAnalyzers();
+
+                analyzers = FilterAnalyzers(analyzers, args).ToImmutableArray();
+
+                if (analyzers.Length == 0)
+                {
+                    PrintHelp();
+                    return;
+                }
+
+                MSBuildWorkspace workspace = MSBuildWorkspace.Create();
+                Solution solution = workspace.OpenSolutionAsync(args[0]).Result;
+
+                Console.WriteLine($"Loaded solution in {stopwatch.ElapsedMilliseconds}ms");
+
+                stopwatch.Restart();
+
+                var diagnostics = GetAnalyzerDiagnosticsAsync(solution, args[0], analyzers).Result;
+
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -0,0 +1,129 @@
...
+
+                Console.WriteLine($"Found {diagnostics.Count} diagnostics in {stopwatch.ElapsedMilliseconds}ms");
+            }
+        }
+
+        private static IEnumerable<DiagnosticAnalyzer> FilterAnalyzers(IEnumerable<DiagnosticAnalyzer> analyzers, string[] args)
+        {
+            bool useAll = args.Contains("/all");
+
+            HashSet<string> ids = new HashSet<string>(args.Where(y => y.StartsWith("/id:")).Select(y => y.Substring(4)));
```
Maxwe11: Why we need these if useAll is true?

sharwell: We don't. But it also doesn't really matter if we do calculate them.

StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            Compilation compilation = await project.GetCompilationAsync().ConfigureAwait(false);
+            CompilationWithAnalyzers compilationWithAnalyzers = compilation.WithAnalyzers(analyzers);
+
+            return await compilationWithAnalyzers.GetAnalyzerDiagnosticsAsync().ConfigureAwait(false);
+        }
+
+        private static void PrintHelp()
+        {
+            Console.WriteLine("Usage: StyleCopTester [options] <Solution>");
```
vweijsters: This should be `Console.WriteLine("Usage: StyleCopTester <Solution> [options]")`

pdelvo: I should make the command line parsing a bit more robust and support both

StyleCop.Analyzers/StyleCopTester/Program.cs
Maxwe11: More obviously skip specified diagnostics if both `/all` and `/id` options presented.

    StyleCopTester /all /id:SA1010 /id:SA1011

StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// <summary>
+    /// StyleCopTester is a tool that will analyze a solution, find diagnostics in it and will print out the number of
+    /// diagnostics it could find. This is useful to easily test performance without having the overhead of visual
+    /// studio running.
+    /// </summary>
+    internal class Program
+    {
+        private static void Main(string[] args)
+        {
+            // A valid call must have at least two parameters. A solution file and /all or one /id:SAXXXX
```
sharwell: :question: If only one parameter is given (the solution file), I expected it to enable all diagnostics which are enabled by default (just run the default configuration). I don't mind creating a separate issue for this, but I wanted to check and see if you implemented it this way already.

sharwell:
This will be an awesome addition to the project. :sunglasses:

StyleCop.Analyzers/StyleCopTester/Program.cs
pdelvo: I don't do this right now. I would have to do this manually. Im going to do that

PR #962, opened 2015-07-14
Implemented performance improvement for TriviaHelper.GetContainingTriviaList 

### LINKED ISSUE #957: Improve performance of TriviaHelper.GetContainingTriviaList
TriviaHelper.GetContainingTriviaList is currently responsible for 15.4% of the inclusive samples in my cpu sampling. I think this is way to much. The most expensive call in it is actually TriviaList.AddRange with 13.3%. The callers of this method don't really need a TriviaList so it might be a good idea to switch to a different list.

vweijsters:
Fixes #957 

Tested it using the StyleCopTester pull request.
In release mode, running SA1512 over the Microsoft/CodeContracts solution resulted in:

Scenario | Result (time spent in TriviaHelper.GetContainingTriviaList)
----- | -----
Original | 12764 ms (average)
New | 1962 ms (average)

Big thanks to @pdelvo for helping out (= doing a lot of the work) with this one.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
@@ -139,7 +139,7 @@
             SyntaxTriviaList part1;
             SyntaxTriviaList part2;
 
-            triviaIndex = token.TrailingTrivia.IndexOf(trivia);
+            triviaIndex = BinarySearch(token.TrailingTrivia, trivia);
             if (triviaIndex != -1)
             {
                 var nextToken = token.GetNextToken(includeZeroWidth: true);
@@ -150,13 +150,41 @@
             else
             {
                 var prevToken = token.GetPreviousToken();
-                triviaIndex = prevToken.TrailingTrivia.Count + token.LeadingTrivia.IndexOf(trivia);
+                triviaIndex = prevToken.TrailingTrivia.Count + BinarySearch(token.LeadingTrivia, trivia);
 
                 part1 = prevToken.TrailingTrivia;
                 part2 = token.LeadingTrivia;
             }
 
             return new DualTriviaListHelper(part1, part2);
+        }
+
+        private static int BinarySearch(SyntaxTriviaList leadingTrivia, SyntaxTrivia trivia)
+        {
+            int low = 0;
+            int high = leadingTrivia.Count - 1;
+            while (low <= high)
+            {
+                int index = low + ((high - low) >> 1);
+                int order = leadingTrivia[index].Span.CompareTo(trivia.Span);
+
+                if (order == 0)
+                {
+                    return index;
+                }
+
+                if (order < 0)
+                {
+                    low = index + 1;
+                }
+                else
+                {
+                    high = index - 1;
+                }
+            }
+
+            // Entry was not found
+            throw new KeyNotFoundException("trivia was not in trivia list");
         }
 
         /// <summary>
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
@@ -184,7 +184,7 @@
             }
 
             // Entry was not found
-            throw new KeyNotFoundException("trivia was not in trivia list");
+            return -1;
         }
 
         /// <summary>


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
```diff
@@ -14,20 +17,20 @@ internal static class TriviaHelper
...
         {
             for (var index = 0; index < triviaList.Count; index++)
             {
                 var currentTrivia = triviaList[index];
                 switch (currentTrivia.Kind())
                 {
-                case SyntaxKind.EndOfLineTrivia:
-                case SyntaxKind.WhitespaceTrivia:
-                    break;
+                    case SyntaxKind.EndOfLineTrivia:
```
pdelvo: You should probably revert thise indentation changes

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1512SingleLineCommentsMustNotBeFollowedByBlankLine.cs
```diff
@@ -170,9 +171,9 @@ private static bool IsOnOwnLine(SyntaxTriviaList triviaList, int triviaIndex)
             return false;
         }
 
-        private static bool IsPartOfFileHeader(SyntaxTriviaList triviaList, int triviaIndex)
+        private static bool IsPartOfFileHeader(IReadOnlyList<SyntaxTrivia> triviaList, int triviaIndex)
         {
-            if (triviaList.FullSpan.Start > 0)
+            if (triviaList[0].FullSpan.Start > 0)
```
pdelvo: Can you be sure that triviaList is not empty?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
vweijsters: Will do. The auto formatter of Visual Studio got me again :disappointed:

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1512SingleLineCommentsMustNotBeFollowedByBlankLine.cs
vweijsters: No, probably not.

pdelvo:
Thats a really significant improvement. Using the StyleCopTester with roslyn 1.0.0 running all our analyzers on Microsoft/CodeContracts gave me these results (green arrows are improvements)

comparison

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
```diff
@@ -130,19 +133,30 @@ internal static SyntaxTriviaList WithoutLeadingWhitespace(this SyntaxTriviaList
...
                 var nextToken = token.GetNextToken(includeZeroWidth: true);
-                return token.TrailingTrivia.AddRange(nextToken.LeadingTrivia);
+
+                part1 = token.TrailingTrivia;
+                part2 = nextToken.LeadingTrivia;
+            }
+            else
+            {
+                var prevToken = token.GetPreviousToken();
+                triviaIndex = prevToken.TrailingTrivia.Count + token.LeadingTrivia.IndexOf(trivia);
```
pdelvo: triviaIndex is not used. This is actually a (relativly) big performance hit. Thats how I noticed it.

sharwell: :question: Is this something that still needs to be fixed before the change is merged?

pdelvo: You could remove this when merging. This shows up prominently in performance reports

vweijsters: :exclamation: You cannot remote that. `triviaIndex` is an out parameter and several callers of the method require that value.

vweijsters:
I also removed all instances of SyntaxTriviaList.AddRange. 
This will address some of the issues in #964.

I added it in the PR because it needed the helper class introduced here.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
pdelvo: You're right. I overlooked that

PR #965, opened 2015-07-15
Remove unnecessary Location allocations

pdelvo:
The most allocated object in our source is SourceLocation. I looked through our source and we allocate locations when we actually just want the line span. I removed these allocations.


vweijsters:
:+1: Looks good to me

PR #967, opened 2015-07-15
Don't use semantics for SA1121 if they are not necessary

pdelvo:
This diagnostic was one of our slowest. Now you have to search for it in a performance report.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SyntaxTriviaHelpers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SyntaxTriviaHelpers.cs
@@ -1,0 +1,54 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Linq;
+    using System.Runtime.CompilerServices;
+    using Microsoft.CodeAnalysis;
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SyntaxTriviaHelpers.cs
```diff
@@ -0,0 +1,54 @@
+﻿namespace StyleCop.Analyzers.Helpers
+{
+    using System.Linq;
+    using System.Runtime.CompilerServices;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Foo = Microsoft.CodeAnalysis.CSharp;
```
sharwell: :expressionless:

pdelvo: >.<

sharwell:
:bulb: **SyntaxTriviaHelpers.cs** should actually be named **SyntaxTreeHelpers.cs**.

PR #970, opened 2015-07-15
Improve SA1516 performance and reduce allocations

pdelvo:
With this change the analyzer does no longer create two unnecessary immutable lists.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516ElementsMustBeSeparatedByBlankLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516ElementsMustBeSeparatedByBlankLine.cs
@@ -216,8 +216,7 @@
             var parent = firstNode.Parent;
 
             var allTrivia = parent.DescendantTrivia(TextSpan.FromBounds(firstNode.Span.End, secondNode.Span.Start),
-                descendIntoTrivia: true,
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516ElementsMustBeSeparatedByBlankLine.cs
```diff
@@ -216,8 +217,7 @@ private static void ReportIfThereIsNoBlankLine(SyntaxNodeAnalysisContext context
 
             var allTrivia = parent.DescendantTrivia(TextSpan.FromBounds(firstNode.Span.End, secondNode.Span.Start),
                 descendIntoTrivia: true,
-                descendIntoChildren: n => true)
-                .ToImmutableList();
+                descendIntoChildren: n => true);
```
sharwell: :question: (I know you weren't working on this but...) Isn't this the same as passing `descendIntoChildren: null`, or simply omitting the argument? If so, the `null` value will avoid some virtual calls when traversing the tree.

pdelvo: Youre right. I changed that.

vweijsters:
:+1: looks good to me

PR #972, opened 2015-07-15
Fixes SA1505 issue with comments

### LINKED ISSUE #971: SA1505 curly bracket must not be followed by a blank line wrongly reported when next line is a comment
The following code wrongly reports SA1505:

    public object SomeMethod()
    {
        // Some comment.
        return null;
    }

The code 'fix' amusingly *inserts* a blank line before the comment.

vweijsters:
Fixes #971 

The issue was introduced in #962 and the test set didn't detect the issue.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1505OpeningCurlyBracketsMustNotBeFollowedByBlankLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1505OpeningCurlyBracketsMustNotBeFollowedByBlankLine.cs
@@ -127,7 +127,7 @@
                 startIndex++;
             }
 
-            startIndex = startIndex % separatingTrivia.Count;
+            startIndex = (startIndex == separatingTrivia.Count) ? 0 : startIndex + 1;
 
             var done = false;
             var eolCount = 0;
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1505OpeningCurlyBracketsMustNotBeFollowedByBlankLine.cs
```diff
@@ -127,7 +127,7 @@ private void AnalyzeOpenBrace(SyntaxNodeAnalysisContext context, SyntaxToken ope
                 startIndex++;
             }
 
-            startIndex = startIndex % separatingTrivia.Count;
+            startIndex = (startIndex == separatingTrivia.Count) ? 0 : startIndex + 1;
```
sharwell: :memo: I tried to ask about this line after merging #962, but the diff went away (look at the files tab in that pull request) :confused:

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1505OpeningCurlyBracketsMustNotBeFollowedByBlankLine.cs
```diff
@@ -127,7 +127,7 @@ private void AnalyzeOpenBrace(SyntaxNodeAnalysisContext context, SyntaxToken ope
                 startIndex++;
             }
 
-            startIndex = startIndex % separatingTrivia.Count;
```
sharwell: :memo: I tried to ask about this line after merging #962, but the diff went away (look at the files tab in that pull request) :confused:

PR #978, opened 2015-07-16
Improve efficiency of SA1120 (CommentsMustContainText)

### LINKED ISSUE #963: Performance opportunity in SA1120 analyzer
In my tests [this line](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1120CommentsMustContainText.cs#L84) was using up 3.7% of all samples (inclusive). I think that's too much for what its actually doing. It should be relatively easy to improve this.

sharwell:
Fixes #963

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1120CommentsMustContainText.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1120CommentsMustContainText.cs
@@ -80,10 +80,7 @@
 
         private static void HandleSingleLineComment(SyntaxTreeAnalysisContext context, SyntaxTrivia singleLineComment)
         {
-            // Remove the leading // from the comment
-            var commentText = singleLineComment.ToString().Substring(2);
             int index = 0;
-
             var list = TriviaHelper.GetContainingTriviaList(singleLineComment, out index);
             var firstNonWhiteSpace = TriviaHelper.IndexOfFirstNonWhitespaceTrivia(list);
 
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1120CommentsMustContainText.cs
```diff
@@ -91,19 +88,40 @@ private static void HandleSingleLineComment(SyntaxTreeAnalysisContext context, S
...
             // blank line in the code which is commented out.
-            bool isFirstOrLast = index == firstNonWhiteSpace;
-            if (!isFirstOrLast)
+            bool isFirst = index == firstNonWhiteSpace;
+            if (!isFirst)
             {
                 // This is -2 because we need to go back past the end of line trivia as well.
                 var lastNonWhiteSpace = TriviaHelper.IndexOfTrailingWhitespace(list) - 2;
-                isFirstOrLast = index == lastNonWhiteSpace;
+                if (index == lastNonWhiteSpace)
```
sharwell: :exclamation: This should be `!=`.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1120CommentsMustContainText.cs
```diff
@@ -91,19 +88,40 @@ private static void HandleSingleLineComment(SyntaxTreeAnalysisContext context, S
...
+        private static bool IsNullOrWhiteSpace(string value, int startIndex)
+        {
+            if (value == null)
+            {
+                return true;
+            }
+
+            for (int i = startIndex; i < value.Length; i++)
+            {
+                if (!char.IsWhiteSpace(value[startIndex]))
```
sharwell: :exclamation: This should be `value[i]`.

Maxwe11:
:+1:  LGTM

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1120CommentsMustContainText.cs
pdelvo: Was a unit test failing because of that? If not we should probably add one

Issue #973, opened 2015-07-16
SA1008 (Opening parenthesis must not be preceded by a space) triggered after named parameter


Noryoko:
example:

string.Concat(arg0: "test", arg1: ("A" + "B") + "C");

sharwell:
Hi @Noryoko! Thank you for taking the time to try out this project and report the issue. I've labeled this as **up for grabs** for now; if you'd like to try getting your hands dirty (to fix it), just let us know. Otherwise one of the regulars will pick it up. :+1:

Noryoko:
I can work on this over the weekend. Seems like a good starter.

sharwell:
:+1: I sent you an invitation to join the [DotNetAnalyzers](https://github.com/DotNetAnalyzers) organization, which will allow be to assign the issue to you. Feel free to ask questions here or in Gitter if you get stuck.

Noryoko:
Thanks! This should make for an interesting weekend.

PR #979, opened 2015-07-16
added test case for SA1008

Noryoko:
It seems #973 this was fixed by #918. I've added a test case.


StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1008UnitTests.cs
```diff
@@ -1638,4 +1674,4 @@ protected override CodeFixProvider GetCSharpCodeFixProvider()
             return new SA1008CodeFixProvider();
         }
     }
-}
+}
```
Maxwe11: bring back new line please

PR #983, opened 2015-07-16
Add unit tests for SA1004

### LINKED ISSUE #229: Diagnostic unit tests: SA1004 (DocumentationLinesMustBeginWithSingleSpace)
Implement unit tests for #10.

sharwell:
This diagnostic is also enabled by default now.

Fixes #229

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1004UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1004UnitTests.cs
@@ -1,0 +1,112 @@
+﻿namespace StyleCop.Analyzers.Test.SpacingRules
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.SpacingRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1004DocumentationLinesMustBeginWithSingleSpace"/>.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


pdelvo:
Should we also report warnings for multiline xml comments?

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1004UnitTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    ///</summary>
+    ///   <param name=""x"">The document root.</param>
+    ///    <param name=""y"">The XML header token.</param>
+    private void Method1(int x, int y)
+    {
+    }
+}
+";
+
+            // Currently the extra indentation for <param> elements is not checked.
```
pdelvo: We should probably add a skipped test for that.

sharwell: Do we *want* to check for extra indentation? It caused a problem in the past (#715). If you think we should check extra indentation for some set of rules, I'm thinking we should create a new issue to define exactly what those rules are.

sharwell:
> Should we also report warnings for multiline xml comments?

Probably, but I'm thinking that should be a separate issue. It wouldn't be the first time we implemented support for single-line documentation comments but left out the multi-line version.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1004UnitTests.cs
pdelvo: Sounds good

pdelvo:
Im okay with not supporting multiline documentation for now. I have not seen anyone using them

PR #986, opened 2015-07-17
Fix NRE in SA1505 when open brace is last in file

### LINKED ISSUE #981: SA1505 analyzer crashes when the open brace token is the last token in the file
It is throwing a NRE because the next token has SyntaxKind.None. Discovered while typing

sharwell:
Fixes #981


pdelvo:
LGTM. We should probably look for uses of GetNextToken and make sure that we don't have this problem in other places.

Issue #974, opened 2015-07-16
SA1009 (ClosingParenthesisMustBeSpacedCorrectly) fails with any method declaration


roadz99:
Sample code fails unless you add a trailing space after the closing parenthesis

    using System;

    public class Foo
    {
        public void Method()
        {
        }
    }

sharwell:
I marked this **up for grabs**. Did you want to work on it? :smile:

roadz99:
There is no test class for this rule so I'm a little scared to change it!

sharwell:
You could fix #194. :smile:

Maxwe11:
I also saw this earlier but was confused because [this diagnostic disabled](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs#L37).

sharwell:
@Maxwe11 the disabled diagnostics are enabled in debug builds, and you can also enable them manually (via ruleset files and/or the Code Analysis UI).

roadz99:
I'll take this one too. Looks like a simple fix but I'll try to add a good number of unit tests too.

Issue #194, opened 2014-11-20
Diagnostic unit tests: SA1009 (ClosingParenthesisMustBeSpacedCorrectly)


sharwell:
Implement unit tests for #15.

roadz99:
I'll make a start on this one.

PR #988, opened 2015-07-17
SA1009 - ClosingParenthesisMustBeSpacedCorrectly bug fix and unit tests

roadz99:
Added a number of unit tests to address issue #194. I had a little trouble trying to get 100% code coverage because I couldn't work out what some of the code was trying to do. If anybody can suggest additional scenarios I'd be happy to add more tests.

Fixed issue #974.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
@@ -1,0 +1,195 @@
+﻿namespace StyleCop.Analyzers.Test.SpacingRules
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.SpacingRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1009ClosingParenthesisMustBeSpacedCorrectly"/>
+    /// </summary>
+    public class SA1009UnitTests : CodeFixVerifier
+    {
+        private const string SimpleMethodOutline = @"using System;
+
+public class Foo
+{
+    public void Method()
+    {
+    }
+}";
+
+        [Fact]
+        public async Task TestEmptySourceAsync()
+        {
+            var testCode = string.Empty;
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestMethodWithNoParametersAsync()
+        {
+            const string testCode = @"using System;
+
+public class Foo
+{
+    public void Method()
+    {
+    }
+}";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestMethodWithWhitespaceBeforeClosingParenthesisAsync()
+        {
+            const string testCode = @"using System;
+
+public class Foo
+{
+    public void Method( )
+    {
+    }
+}";
+
+            DiagnosticResult expected = this.CSharpDiagnostic().WithArguments(" not", "preceded").WithLocation(5, 25);
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestMethodWith2CorrectlySpaceParametersAsync()
+        {
+            const string testCode = @"using System;
+
+public class Foo
+{
+    public void Method(int param1, int param2)
+    {
+    }
+}";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestMethodWith2ParametersndSpacebeforeClosingParenthesisAsync()
+        {
+            const string testCode = @"using System;
+
+public class Foo
+{
+    public void Method(int param1, int param2 )
+    {
+    }
+}";
+            DiagnosticResult expected = this.CSharpDiagnostic().WithArguments(" not", "preceded").WithLocation(5, 25);
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
```diff
@@ -0,0 +1,278 @@
...
+    using StyleCop.Analyzers.SpacingRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1009ClosingParenthesisMustBeSpacedCorrectly"/>
+    /// </summary>
+    public class SA1009UnitTests : CodeFixVerifier
+    {
+        private const string SimpleMethodOutline = @"using System;
```
pdelvo: I think this constant can be removed

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
```diff
@@ -0,0 +1,278 @@
...
+    public void Method(int param1, int param2)
+    {
+    }
+}";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestMethodWith2ParametersndSpaceBeforeClosingParenthesisAsync()
```
pdelvo: There is a spelling mistake in the method name

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
```diff
@@ -0,0 +1,278 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using StyleCop.Analyzers.SpacingRules;
+    using TestHelper;
+    using Xunit;
+
+    /// <summary>
+    /// Unit tests for <see cref="SA1009ClosingParenthesisMustBeSpacedCorrectly"/>
+    /// </summary>
+    public class SA1009UnitTests : CodeFixVerifier
```
sharwell: :bulb: For diagnostics that don't have a code fix, it's best to use `DiagnosticVerifier` as the base class.

pdelvo:
I think test cases are missing for:
- Foo().Test
- Foo()?.Test
- ((1+1))
- new Action(() => { })()
- Foo()[0]
- true ? Foo() : Bar()
I might have missed some but these are the ones I could think of.

This is a good start!

roadz99:
Which one of the following is correct? The first line looks correct to me but the rule fails the first one and allows the one with the space.

(i)++
(i) ++

sharwell:
@roadz99 The first one is correct according to *this* rule, but it might fail SA1119 (StatementMustNotUseUnnecessaryParenthesis) (which you don't need to worry about here).

roadz99:
Ok. I'll fix that bug in the analyser.

roadz99:
I realised that the increment/decrement operator tests were duplicating SA1020IncrementDecrementSymbolsMustBeSpacedCorrectly so I've taken them out.

sharwell:
@roadz99 It depends on which warning was getting produced by the code example you gave. If SA1020 is reported, then it's an issue with that diagnostic. But if SA1009 is reported, then it would be this one. Either way it seems like a reasonable test case to include for both diagnostics.

roadz99:
@sharwell thanks for the explanation. I'll leave the tests in and fix the anlyzer.

roadz99:
@pdelvo many thanks. That helped a lot.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
```diff
@@ -0,0 +1,557 @@
...
+        {
+            var invalidStatement = @"var x = (true == true)? GetA() : GetB();";
+
+            DiagnosticResult expected = this.CSharpDiagnostic().WithArguments(string.Empty, "followed").WithLocation(7, 34);
+
+            await this.TestWhitespaceInStatementOrDeclAsync(invalidStatement, expected).ConfigureAwait(false);
+        }
+
+        /// <inheritdoc/>
+        protected override IEnumerable<DiagnosticAnalyzer> GetCSharpDiagnosticAnalyzers()
```
pdelvo: NIT: This should probably be the last member in this class

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
```diff
@@ -0,0 +1,557 @@
...
+        {
+            var invalidStatement = @"System.EventHandler handler = (s, e)=> { };";
+
+            DiagnosticResult expected = this.CSharpDiagnostic().WithArguments(string.Empty, "followed").WithLocation(7, 48);
+
+            await this.TestWhitespaceInStatementOrDeclAsync(invalidStatement, expected).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestSpaceAfterParenthisisInAddOperationAsync()
```
pdelvo: NIT: You could use a theory to write these tests.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
sharwell: By convention, yes, but *technically* we violate the ordering rules everywhere else. StyleCop says private members should appear after protected members.

pdelvo:
In SA1009ClosingParenthesisMustBeSpacedCorrectly.cs could you change "DisabledNoTests" to "EnabledByDefault" in line 37?

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
roadz99: Please could you clarify what you mean by using a theory? Also, NIT? Sorry - I'm new here :-)

pdelvo: You can instead of using Fact do something like this:
```csharp
[Theory]
[InlineData("+")]
[InlineData("-")]
[InlineData("*")]
[InlineData("/")]
TestSpaceAfterParenthisisInOperationAsync(string operator) { ... }
```
xunit will call this method for each of these operators and you have to write this method only once.

NIT means nitpicking. So not very important

sharwell:
Good job both of you. :+1:

PR #991, opened 2015-07-17
Update to the latest pre-release version of xunit

sharwell:
This should improve testing performance a bit.


pdelvo:
LGTM. I tested this yesterday and it works for me. But I couldn't really notice big performance boost unfortunately.

sharwell:
I'm seeing ~30% faster to discover tests after I rebuild.

PR #994, opened 2015-07-18
Cache empty arrays

Maxwe11:
Caching non empty arrays in progress :smile:

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
@@ -76,7 +76,7 @@
             {
                 if (!XmlCommentHelper.HasDocumentation(declaration))
                 {
-                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation()));
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation(), ArrayEx.Empty<object>()));
                 }
             }
         }
@@ -95,7 +95,7 @@
             {
                 if (!XmlCommentHelper.HasDocumentation(declaration))
                 {
-                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation()));
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation(), ArrayEx.Empty<object>()));
                 }
             }
         }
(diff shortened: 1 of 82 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
```diff
@@ -76,7 +76,7 @@ private void HandleTypeDeclaration(SyntaxNodeAnalysisContext context)
             {
                 if (!XmlCommentHelper.HasDocumentation(declaration))
                 {
-                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation()));
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation(), ArrayEx.Empty<object>()));
```
sharwell: :exclamation: This is handled by Roslyn in cases where mscorlib contains `Array.Empty<T>`.  I don't want to make the code more complex to address this. If we need it, we can target .NET 4.6 or another framework which contains it.

sharwell:
I didn't see any instances other than the fixes to `params` arguments. As such, based on my previous comment I'm closing as **wontfix**.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
Maxwe11: Any links with info about this?

sharwell: * dotnet/roslyn#1103
* dotnet/roslyn#1117

Maxwe11: But I don't know what ver of mscorlib used there
http://tryroslyn.azurewebsites.net/#K4Zwlgdg5gBAygTxAFwKYFsDcAoADsAIwBswBjGUogQxBBgGEYBvbGNmfYsmANwHswAExgARABQBKZq3ayAspJyyAvjLZqOhEuX5CYC3FQBOVdHT4EAVqlLIA2gF0YfKSxXZlQAA

sharwell: It's one that doesn't have `Array.Empty<T>()` (it doesn't compile even if you try to reference it directly).

sharwell: :memo: Also note that all of these cases only occur in code that has violations. Clean code, which is the most important case to optimize for, does not benefit from these changes at all.

PR #999, opened 2015-07-19
Improve IndentationHelper

Maxwe11:
- removed string concatenation
- removed char array allocation
- removed delegate and closure allocation

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/IndentationHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/IndentationHelper.cs
@@ -1,6 +1,5 @@
 ﻿namespace StyleCop.Analyzers.Helpers
 {
-    using System.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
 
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/IndentationHelper.cs
```diff
@@ -92,19 +91,23 @@ public static SyntaxTrivia GenerateWhitespaceTrivia(IndentationOptions indentati
...
             {
-                indentationString = string.Concat(leadingTrivia[i].ToFullString(), indentationString);
+                if (!trivia.IsKind(SyntaxKind.WhitespaceTrivia))
+                {
+                    break;
+                }
+
+                foreach (char c in trivia.ToFullString())
+                {
+                    indentationCount += c == '\t' ? indentationOptions.TabSize : 1;
```
sharwell: :bulb: The behavior of this is wrong for text like `" \t \t \t \t"` (written as a string literal just so you can see the special characters clearly). Seems like if we're already rewriting the algorithm we could find the issues and fix them too. What do you think?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/IndentationHelper.cs
```diff
@@ -92,19 +91,23 @@ public static SyntaxTrivia GenerateWhitespaceTrivia(IndentationOptions indentati
...
+                }
             }
 
-            var indentationCount = indentationString.ToCharArray().Sum(c => IndentationAmount(c, indentationOptions));
             return indentationCount / indentationOptions.IndentationSize;
```
sharwell: :question: Shouldn't this be:

```csharp
return (indentationCount + indentationOptions.IndentationSize - 1) / indentationOptions.IndentationSize;
```

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/IndentationHelper.cs
Maxwe11: >The behavior of this is wrong for text like " \t \t \t \t" 

What's wrong with this?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/IndentationHelper.cs
Maxwe11: `IndentationSize` should be already included as trivia itself in `leadingTrivia` list

sharwell: Suppose you have a line indented by 3 spaces, but the indentation size is 4. The current implementation will return 0 (i.e. truncation), but my implementation would return 1 (i.e. a ceiling function).

Maxwe11: That's a bug. Bugs should be fixed separately with appropriate test. Fill in issue :sunglasses:

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/IndentationHelper.cs
sharwell: It should return 4, but it currently returns 5. It's counted like you wrote `"    \t\t\t\t"`.

Maxwe11: It works as existing code.

sharwell:
I filed issues for the remaining work.

Issue #1013, opened 2015-07-20
Code fix: SA1207 (ProtectedMustComeBeforeInternal)


sharwell:
Implement a code fix for #67.

EdwinEngelen:
grabbed

sharwell:
Fixed in #1030

PR #1030, opened 2015-07-23
SA1207 code fix added. Closes issue #1013.

EdwinEngelen:
:memo: `SA1207CodeFixProvider.ComputeReplacementToken` has some open questions.


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            var newDeclarationNode = originalDeclarationNode.ReplaceTokens(childTokens, ComputeReplacementToken);
+
+            var newSyntaxRoot = syntaxRoot.ReplaceNode(originalDeclarationNode, newDeclarationNode);
+            return document.WithSyntaxRoot(newSyntaxRoot);
+        }
+
+        // QUESTION: Why two tokens as argument?
+        // QUESTION: Which token should we use?
```
sharwell: * Sometimes we use a `Dictionary<SyntaxToken, SyntaxToken>` to map nodes we analyze to their replacements. In this case, we want to use `originalToken` because that's the key we added to the dictionary.
* Sometimes we want to create a new token right within this method, as a function of the current input to the method. In this case, we want to use `rewrittenToken` because that's the token in the most recent state of the tree.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+                context.RegisterCodeFix(CodeAction.Create(OrderingResources.SA1207CodeFix, token => GetTransformedDocumentAsync(context.Document, diagnostic, token)), diagnostic);
+            }
+
+            return SpecializedTasks.CompletedTask;
+        }
+
+        private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        {
+            var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
```
Maxwe11: :bulb: Consider replace these two line with `DocumentEditor`.
`DocumentEditor` request syntax root and semantic model at one hit because it knows that `GetSemanticModelAsync` requests syntax root inside and cache it so it requests syntax root synchronously.
Example of DocumentEditor usage you can find in the [repo](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/search?utf8=%E2%9C%93&q=DocumentEditor)

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+                return document;
+            }
+
+            var newDeclarationNode = originalDeclarationNode.ReplaceTokens(childTokens, ComputeReplacementToken);
+
+            var newSyntaxRoot = syntaxRoot.ReplaceNode(originalDeclarationNode, newDeclarationNode);
+            return document.WithSyntaxRoot(newSyntaxRoot);
+        }
+
+        // QUESTION: Why two tokens as argument?
```
sharwell: During the replacement process, the syntax tree goes through several iterations - each time a change is applied some tokens only exist in the tree from that point forward. `originalToken` refers to the token at a particular position when the replacement process started. `rewrittenToken` refers to the current token at the same position.

:memo: If no changes were made to `originalToken` or any of its parents (yet), then `rewrittenToken` is the same as `originalToken`.

Here's an example where you could see a difference:

1. The replacement process replaces the `{` token of a `ClassDeclarationSyntax`.
2. The replacement process gets to the `}` token of the same `ClassDeclarationSyntax`. At this point:
  * `originalToken.Parent` refers to the original `ClassDeclarationSyntax`
  * `rewrittenToken.Parent` refers to the updated `ClassDeclarationSyntax` which has a new `{` token

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+        // QUESTION: Which token should we use?
+        // QUESTION: We don't need to verify that the original order of protected and internal is incorrect, do we?
```
sharwell: Code fixes run on the same input as the diagnostic. If the diagnostic only gets reported if the order is incorrect, then the code fix only runs when the order is incorrect.

EdwinEngelen: Clear. We don't need to verify that the original order is incorrect.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+        private static SyntaxToken ComputeReplacementToken(SyntaxToken originalToken, SyntaxToken rewrittenToken)
+        {
+            if (rewrittenToken.IsKind(SyntaxKind.InternalKeyword))
+            {
+                return SyntaxFactory.Token(SyntaxKind.ProtectedKeyword);
```
sharwell: :exclamation: This will not preserve trivia (whitespace, comments, preprocessor directives, etc) from the original token. You should do the following:

```csharp
return SyntaxFactory.Token(SyntaxKind.ProtectedKeyword).WithTriviaFrom(rewrittenToken);
```

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+            else if (rewrittenToken.IsKind(SyntaxKind.ProtectedKeyword))
+            {
+                return SyntaxFactory.Token(SyntaxKind.InternalKeyword);
```
sharwell: :exclamation: This will not preserve trivia (see previous instance).

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
EdwinEngelen: I've changed the code to use 'originalToken' (as requested). Should I use the trivia from 'rewrittenToken'?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1207CodeFixProvider.cs
sharwell: :exclamation: The semantic model isn't used by this code fix. This line should just be removed. `DocumentEditor` would be unnecessary overhead.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1207UnitTests.cs
```diff
@@ -54,11 +60,11 @@ private static IEnumerable<string> ValidAccessModifiers
             }
         }
 
-        private static IEnumerable<string> InvalidAccessModifiers
+        private static IEnumerable<Tuple<string, string>> InvalidAndFixedAccessModifiers
         {
             get
             {
-                yield return "internal protected";
+                yield return Tuple.Create("internal protected", "protected internal");
```
sharwell: :exclamation: Add the following tests:

| Before | After |
| --- | --- |
| `internal static protected` | `protected static internal` |
| `internal protected abstract` | `protected internal abstract` |
| `/*comment*/internal /* 2 */  protected /*3*/` | `/*comment*/protected /* 2 */  internal /*3*/` |

EdwinEngelen: Should the leading spaces be included?

Issue #205, opened 2014-11-20
Diagnostic unit tests: SA1019 (MemberAccessSymbolsMustBeSpacedCorrectly)


sharwell:
Implement unit tests for #25.

nukefusion:
Just had a stab at this and issued a pull request.

PR #1031, opened 2015-07-23
Diagnostic unit tests: SA1019 (MemberAccessSymbolsMustBeSpacedCorrectly)

nukefusion:
Implementing unit tests for #205

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1019UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1019UnitTests.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


sharwell:
Welcome, new contributor! :tada:

sharwell:
:bulb: It's a solid start. You should add some tests for cases where a comment exists on either side of the operator.

nukefusion:
Thanks for the feedback. What's the exact expectation in those scenarios? The current Analyzer implementation will flag the following as SA1019:
```
foo. // This is foo
    Bar();
```
but not:
```
foo.// This is foo
    Bar();
```
whereas the existing (older) StyleCop implementation will flag both.
StyleCop docs state:
>  A member access symbol should not have whitespace on either side, unless it is the first or last character on the line.

I would interpret that such that the Analyzer implementation is more correct, although adding a comment right after a member access symbol is pretty ugly IMO :astonished: 
What's your take on this?

sharwell:
I was actually thinking about this:

```
foo/**/./**/Bar();
```

However there's one case that you may have seen before and just weren't thinking about it. Lines 27-53 of the [code in this blog post](http://blog.280z28.org/archives/2008/10/20/) have several cases where a comment *precedes* a member access operator.

nukefusion:
I don't think I've *ever* considered adding block comments to either side of the operator like that :)
I've added in some additional test cases for that scenario and for cases where comments are on either the following or preceding line (as referenced in the linked blog post).

Issue #808, opened 2015-05-10
SA1111 (ClosingParenthesisMustBeOnLineOfLastParameter) triggers on brackets used for attributes


vweijsters:
The following code should not but does trigger SA1111:

```C#
    [
    Export
    ]
    public class Foo
    {
    }
```

Stylecop does not trigger for this, inline with the documentation.

PR #1032, opened 2015-07-23
Fix for SA1111 incorrectly reported for Attributes with no arguments

nukefusion:
Fix for #808.  SA1111 being incorrectly reported for Attributes with no arguments when the closing bracket is on the next line.


sharwell:
:bulb: The email address used for your commits isn't associated with your GitHub profile, so GitHub doesn't know to credit you with the contribution once it's merged. You can edit your GitHub profile to add new email addresses you want it to recognize.

sharwell:
:memo: The bug was filed by @vweijsters who won't be able to review it immediately. If no one else picks it up in the meantime, I'll take a look at after I take care of some of the other reviews to make sure it solves the problem and that it's the solution we want in the end. :smile:

nukefusion:
Thanks for the heads-up re e-mail address, didn't notice beforehand. :)

PR #1047, opened 2015-07-26
Move TestEmptySourceAsync into DiagnosticVerifier

sharwell:


Maxwe11:
Any plans make `DiagnosticVerifier` and etc. reuseable and publish nuget package?

sharwell:
@Maxwe11 Not right now. Right now we don't have to worry about things like backward API compatibility or VB code, and can delay implementation of new features until we need them for testing our own things. It's probably best to stay on that path for a while longer until the test framework stabilizes.

Noryoko:
:+1:

PR #1055, opened 2015-07-26
Implement new diagnostic SA1652 (EnableXmlDocumentationOutput)

### LINKED ISSUE #1053: New rule proposal: Documentation output should be enabled
**Category:** Documentation
**Title:** Documentation output should be enabled (could change)
**Description:** When a project is configured to not produce an XML documentation file with the output (this is the default), the semantics of documentation comments are not checked. This quickly results in several other problems, including the following:

* #1012 
* #1050 
* dotnet/roslyn#2773

sharwell:
Fixes #1053.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
@@ -68,6 +68,11 @@
 
         private void HandleTypeDeclaration(SyntaxNodeAnalysisContext context)
         {
+            if (context.GetDocumentationMode() != DocumentationMode.Diagnose)
+            {
+                return;
+            }
+
             BaseTypeDeclarationSyntax declaration = context.Node as BaseTypeDeclarationSyntax;
 
             bool isNestedInClassOrStruct = this.IsNestedType(declaration);
@@ -83,6 +88,11 @@
 
         private void HandleMethodDeclaration(SyntaxNodeAnalysisContext context)
         {
+            if (context.GetDocumentationMode() != DocumentationMode.Diagnose)
+            {
+                return;
+            }
+
             MethodDeclarationSyntax declaration = context.Node as MethodDeclarationSyntax;
             SyntaxKind defaultVisibility = SyntaxKind.PrivateKeyword;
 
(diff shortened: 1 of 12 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1652EnableXmlDocumentationOutput.cs
```diff
@@ -0,0 +1,53 @@
...
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxTreeActionHonorExclusions(HandleSyntaxTree);
+        }
+
+        private static void HandleSyntaxTree(SyntaxTreeAnalysisContext context)
+        {
+            if (context.Tree.Options.DocumentationMode == DocumentationMode.None)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, context.Tree.GetLocation(new TextSpan(0, 0))));
```
pdelvo: Do we really want to report a diagnostic on every SyntaxTree?

sharwell: Maybe, maybe not. Since the documentation mode is a property of each `SyntaxTree`, I figured this makes sense at least until someone has a major problem with it.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
```diff
@@ -218,6 +263,11 @@ private void HandleEventDeclaration(SyntaxNodeAnalysisContext context)
 
         private void HandleEventFieldDeclaration(SyntaxNodeAnalysisContext context)
         {
+            if (context.GetDocumentationMode() != DocumentationMode.Diagnose)
+            {
+                return;
+            }
+
```
JohanLarsson: No way to avoid the repetition of these?

sharwell: Not without additional refactoring. Each of these methods is its own entry and exit point for analysis of a particular type of syntax node.

Noryoko:
:+1:

PR #1059, opened 2015-07-26
SA1009 updated to single code fix

Noryoko:


sharwell:
:+1: That was disturbingly easy.

PR #1061, opened 2015-07-27
Enable SA1019 (MemberAccessSymbolsMustBeSpacedCorrectly) by default

sharwell:
Also remove a test which is no longer necessary.


pdelvo:
:+1:

PR #1065, opened 2015-07-27
Fix handling of verbatim string arguments

### LINKED ISSUE #1064: SA1009: mis-fires for @"" strings
The last line in this code snippet produces a message from SA1009. It claims I have whitespace preceding the closing parenthesis. Which is true, but it's allowable as it is not *immediately* preceding the parenthesis.

            this.connection.Execute(@"
                PRAGMA foreign_keys = ""1"";
                PRAGMA recursive_triggers = ""1"";
                PRAGMA synchronous = ""1"";
            ");

sharwell:
* [x] Add regression test for #1064 
* [x] Fix #1064

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
@@ -418,6 +418,29 @@ public class Foo
             await this.TestWhitespaceInStatementOrDeclAsync(invalidStatement, validStatement, expected).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// This is a regression test for DotNetAnalyzers/StyleCopAnalyzers#1064 "SA1009: mis-fires for verbatim
+        /// strings"
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestVerbatimStringArgumentAsync()
+        {
+            string testCode = @"using System;
+public class Foo
+{
+    public void Method()
+    {
+        Console.WriteLine(@""
+    Written content...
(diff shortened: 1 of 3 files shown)


sharwell:
@AArnott Can you review this? :smile:

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
```diff
@@ -418,6 +418,29 @@ public Foo(int i) : base()
...
+        public async Task TestVerbatimStringArgumentAsync()
+        {
+            string testCode = @"using System;
+public class Foo
+{
+    public void Method()
+    {
+        Console.WriteLine(@""
+    Written content...
+"");
```
AArnott: While this probably repros the problem, would it be interesting to have a test case with spaces on the last line of the quote (indenting the `");` as I had in my original bug)?

sharwell: It would still just be a verbatim string. I looked into it a bit and it doesn't seem like a second test with indentation would improve the overall quality of the regression test.

sharwell: :memo: I made sure to push 446605e before fixing the issue, so you can see in the AppVeyor report that it did reproduce the issue before it was corrected.

Noryoko:
:+1:

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1009UnitTests.cs
AArnott: Great. Thanks

PR #1085, opened 2015-07-28
Fix #1082. SA1011 fires for increment and decrement operators

### LINKED ISSUE #1082: SA1011 (Closing square brackets must be followed by a space) fires for increment and decrement operators)
Example:

    var i = new int[1];
    i[0]++;

I'll grab this if you agree it needs fixing

roadz99:
When implementing the diagnostic/fix for the following case...

```
i[0] ++;
```

...I started breaking a lot of tests. It looks like this rule is careful not to report things that other rules report. I decided it's safer to concentrate my change on the increment and decrement operators.


roadz99:
That's weird, GitHub is showing my fork as up to date. I'll try to figure this out tomorrow.

sharwell:
@roadz99 looks like it was a glitch. I kicked off a new build.

PR #1087, opened 2015-07-29
Fix analysis of comments following varied trivia

### LINKED ISSUE #1083: SA1515 fires after some directives when it shouldn't
The following code raises SA1515 (Single-line comment must be preceded by blank line) errors when it shouldn't.  It raises them for the "// Inside region", "// After region", and "// After pragma" comments even though there are blank lines before them.  It's interesting that it does _not_ have this problem for the "// After if" and "// After endif" comments.

```C#
#region Test

// Inside region
using System;

#endregion

// After region
#pragma warning restore 1234

// After pragma
using System.Reflection;

#if DEBUG

// After if
using System.Resources;

#endif

// After endif
using System.Runtime.InteropServices;

sharwell:
* [x] Add regression test for #1083
* [x] Fix #1083


Noryoko:
LGTM :+1:

PR #1095, opened 2015-07-31
Fix implicit interface implementation handling for events

### LINKED ISSUE #1091: SA1648 (inheritdoc must be used with inheriting class) wrongly raised for event
An event which implements an interface member wrongly reports SA1648 when it has <inheritdoc /> in its XML documentation.

    public interface IX
    {
        /// <summary>
        /// Raised when an E happens.
        /// </summary>
        event EventHandler E;
    }

    public class B
    {
        /// <summary>
        /// Raised when a Q happens.
        /// </summary>
        public virtual event EventHandler Q;
    }

    public class Z : B, IX
    {
        /// <inheritdoc />
        public event EventHandler E;

        /// <inheritdoc />
        public override event EventHandler Q
        {
            add { }
            remove { }
        }
    }

SA1648 is **wrongly** reported for `E`. It is correctly not reported for `Q`.

sharwell:
* [x] Add regression tests for #1091 
* [x] Fixes #1091


Noryoko:
:+1:

PR #1116, opened 2015-08-03
Updated documentation of AnalyzerExtensions

sharwell:


Noryoko:
:+1:

otac0n:
Just FYI: you should not be doing this pattern:

    <see cref="Diagnostic"/>s

You should prefer this:

    <see cref="Diagnostic">Diagnostics</see>

sharwell:
@otac0n My general preference is to use sentence structures that avoid plural forms of type names. The problem with the form you suggested is preferred is it immediately becomes incorrect if you rename the type. In the form I used, the pluralization might not be correct but the refactoring would at least update the text to match the new name.

otac0n:
"Refactorability over readability" is untenable in documentation.

sharwell:
> "Refactorability over readability" is untenable in documentation.

Either without the other is sub-optimal. I added you to our Gitter room - I'll send you some links to material I've been working on. :smile:

otac0n:
Looks good.

PR #1118, opened 2015-08-03
Fix static elements should not be compared to instance elements

### LINKED ISSUE #1117: SA1202 (Public methods must come before private) conflicts with SA1204 (Static methods must appear before instance methods)
The following code produces either SA1202 or SA1204 depending on the order of the two methods.

    public void A()
    {
    }

    private static void B()
    {
    }

This appears to be an error with SA1204, since its precedence should be lower than that of SA1202.

Noryoko:
- [x] Add regression test for #1117
- [x] Fixes #1117

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
@@ -160,12 +160,13 @@
                     && currentAccessLevel < previousAccessLevel)
                 {
                     context.ReportDiagnostic(
-                    Diagnostic.Create(Descriptor,
-                    NamedTypeHelpers.GetNameOrIdentifierLocation(member),
-                    UpperAccessLevelNames[currentAccessLevel],
-                    currentMemberStatic ? " static" : string.Empty,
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
```diff
@@ -126,79 +126,60 @@ private static void HandleCompilationUnit(SyntaxNodeAnalysisContext context)
...
-                        NamedTypeHelpers.GetNameOrIdentifierLocation(members[i + 1]),
-                        UpperAccessLevelNames[nextAccessLevel],
-                        memberType,
-                        LowerAccessLevelNames[currentAccessLevel]));
+                    Diagnostic.Create(Descriptor,
+                    NamedTypeHelpers.GetNameOrIdentifierLocation(member),
+                    UpperAccessLevelNames[currentAccessLevel],
+                    currentMemberStatic ? " static" : string.Empty,
+                    MemberNames[currentSyntaxKind],
+                    LowerAccessLevelNames[previousAccessLevel]));
```
Maxwe11: each level should be indented

    context.ReportDiagnostic(
        Diagnostic.Create(
            Descriptor,
            NamedTypeHelpers.GetNameOrIdentifierLocation(member),
    ...

PR #1120, opened 2015-08-04
Fix SA1204 incorrectly reported for consts

### LINKED ISSUE #1119: SA1204 (Static fields must appear before instance fields) conflicts with SA1203 (Constants must appear before fields)
The following two fields give either SA1204 or SA1203 depending on their order:

```csharp
private const char ReportDelimiterCharacter = ';';

private static readonly string A = "Haddock";
```

The above order is, I think, the correct one. It gives SA1204, which therefore seems to be in error.

Noryoko:
- [x] Add regression tests for #1119
- [x] Fixes #1119

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1204StaticElementsMustAppearBeforeInstanceElements.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1204StaticElementsMustAppearBeforeInstanceElements.cs
@@ -86,7 +86,6 @@
         {
             var previousSyntaxKind = SyntaxKind.None;
             var previousMemberStatic = true;
-            var previousMemberConst = true;
             foreach (var member in members)
             {
                 var currentSyntaxKind = member.Kind();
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1204StaticElementsMustAppearBeforeInstanceElements.cs
```diff
@@ -84,23 +84,29 @@ private static void HandleTypeDelcaration(SyntaxNodeAnalysisContext context)
...
+            var previousSyntaxKind = SyntaxKind.None;
+            var previousMemberStatic = true;
+            var previousMemberConst = true;
+            foreach (var member in members)
             {
-                var currentSyntaxKind = members[i].Kind();
+                var currentSyntaxKind = member.Kind();
                 currentSyntaxKind = currentSyntaxKind == SyntaxKind.EventFieldDeclaration ? SyntaxKind.EventDeclaration : currentSyntaxKind;
+                var modifiers = member.GetModifiers();
+                var currentMemberStatic = modifiers.Any(SyntaxKind.StaticKeyword);
```
sharwell: :bulb: `const` members are static. You can simplify the logic by using:

```csharp
bool currentMemberStatic = modifiers.Any(SyntaxKind.StaticKeyword) || modifiers.Any(SyntaxKind.ConstKeyword);
```

sharwell: :x: The previous comment is not quite correct. If we don't distinguish between static and instance, an SA1204 warning would be reported when a constant follows an instance field, but only SA1203 should be reported in that case.

PR #1122, opened 2015-08-05
Implement a custom BatchFixAllProvider

### LINKED ISSUE #886: SA1028 code fix fails to fix all AssemblyInfo.cs
Repro: Create a new (console) project, all the analyzer reference. SA1028 is reported for a few places in the AssemblyInfo.cs file. The violations can be auto fixed one at a time, but when trying to fix all nothing will happen. No preview window will open.

sharwell:
* This code is a copy of [`BatchFixAllProvider`](https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/CodeFixes/FixAllOccurrences/BatchFixAllProvider.cs) that has been modified to work in a separate assembly by replacing uses of internal APIs with public APIs. In addition, the generated code detection mechanism was completely removed; this detection is already performed by analyzers prior to reporting diagnostics so duplicating the check is a waste of resources.
* Update all code fix providers which use the batch fixer. These code fix providers now use `CustomBatchFixAllProvider` which allows fixes to be applied to files which the built-in batch fixer believes are generated.
* Update the batch fixer to use a deterministic ordering for applying code fixes. For unit tests which include merge conflicts, this ensures the conflict always occurs in a particular order with reproducible results.

Fixes #886


sharwell:
Blocked on #1126.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/CustomBatchFixAllProvider.cs
```diff
@@ -0,0 +1,558 @@
...
+                var fixesBag = new List<CodeAction>[documents.Length];
+                var options = new ParallelOptions() { CancellationToken = fixAllContext.CancellationToken };
+                Parallel.ForEach(documents, options, (document, state, index) =>
+                {
+                    fixAllContext.CancellationToken.ThrowIfCancellationRequested();
+                    fixesBag[index] = new List<CodeAction>();
+                    this.AddDocumentFixesAsync(document, documentsAndDiagnosticsToFixMap[document], fixesBag[index].Add, fixAllContext).Wait(fixAllContext.CancellationToken);
+                });
+
+                if (fixesBag.Any(fixes => fixes?.Count > 0))
```
Maxwe11: >`fixes?.Count`

`fixes` could be `null`?

sharwell: Actually I believe the `Parallel.ForEach` call would throw an exception here if cancellation was requested (the only way `fixesBag` could still contain a null entry here).

Maxwe11:
`CustomBatchFixAllProvider` itself covered by tests? :smile:

sharwell:
> `CustomBatchFixAllProvider` itself covered by tests? :smile: 

It's covered because of #1126. :+1:

sharwell:
@Maxwe11 Specifically, 61.2% of the blocks in this are covered by the current tests. Most of the uncovered code is due to the fact that we only test `FixAllScope.Document`.

It might be possible to bring over the tests for this from Roslyn. I'll open a new issue for this.

pdelvo:
Looks good to me :+1:

PR #1125, opened 2015-08-05
StyleCopTester improvements

sharwell:
* Do not require `/all` or `/id:SAXXXX`
* Provide a default debug configuration that works (clone and go)
* Include a summary of located diagnostics by ID

### DIFF
--- a/StyleCop.Analyzers/StyleCopTester/Program.cs
+++ b/StyleCop.Analyzers/StyleCopTester/Program.cs
@@ -20,8 +20,8 @@
     {
         private static void Main(string[] args)
         {
-            // A valid call must have at least two parameters. A solution file and /all or one /id:SAXXXX
-            if (args.Length <= 1)
+            // A valid call must have at least one parameter (a solution file). Optionally it can include /all or /id:SAXXXX.
+            if (args.Length < 1)
             {
                 PrintHelp();
             }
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -39,15 +39,21 @@ private static void Main(string[] args)
                 }
 
                 MSBuildWorkspace workspace = MSBuildWorkspace.Create();
-                Solution solution = workspace.OpenSolutionAsync(args[0]).Result;
+                string solutionPath = args.SingleOrDefault(i => !i.StartsWith("/"));
+                Solution solution = workspace.OpenSolutionAsync(solutionPath).Result;
```
Maxwe11: :bulb:  `GetAwaiter().GetResult()`
It would be easier when it throws

StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -59,7 +65,20 @@ private static IEnumerable<DiagnosticAnalyzer> FilterAnalyzers(IEnumerable<Diagn
 
             foreach (var analyzer in analyzers)
             {
-                if (useAll || analyzer.SupportedDiagnostics.Any(y => ids.Contains(y.Id)))
+                if (useAll)
+                {
+                    yield return analyzer;
+                }
```
Maxwe11: this branch should be moved out of `foreach`

sharwell: It wouldn't really make a difference. This is not performance-critical code (it's a one-time initialization in a test application). Also, it would still need to use a `foreach` loop because of the `yield return` statements.

StyleCop.Analyzers/StyleCopTester/Program.cs
sharwell: I'll leave this for another time.

PR #1126, opened 2015-08-05
Test all fix all providers

pdelvo:
With this PR all code fix tests will also try to verify that the fix all provider works correctly. This is WIP because of some test failures.


pdelvo:
@sharwell I merged your PR

sharwell:
The remaining test which fails *only* fails when you run the entire test suite. If you run the test either by itself or limited to just the tests in `SA1503UnitTests`, it passes.

sharwell:
I'm going to disable the failing test if it continues to fail after #1122.

PR #1128, opened 2015-08-06
update ordering diagnostics to proper StyleCop order

### LINKED ISSUE #1123: SA1204 incorrectly reported when private static method follows non-private instance method
The following code should not report SA1204, but actually reports it multiple times:

    public class TestClass
    {
        public static void A1() { }
        public void A2() { }
        internal static void A3() { }
        internal void A4() { }
        protected static void A5() { }
        protected void A6() { }
        private static void A7() { }
        static void A7a() { }
        private void A8() { }
        void A8a() { }
    }

I think that the following changes would fix the code:

- constants and static fields should be treated as the same from the standpoint of this rule (since SA1203 handles that distinction),

- in the `HandleMemberList()` method, each time the access level changes *or* each time the "syntax kind" changes, `previousMemberStatic` should be reset to `true`.

I'm close to a fix for this.

Noryoko:
Originally found in 1204. The also applies to 1202, 1203, 1214, and 1215. 1215 was updated to match the diagnostic location style of the other diagnostic rules.

fixes #1123

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1204UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1204UnitTests.cs
@@ -22,16 +22,66 @@
         public async Task TestValidOrderingAsync()
         {
             var testCode = @"public static class TestClass1 { }
-public class TestClass2 {
+
+public class TestClass2
+{
     public const int TestField1 = 1;
-    public static readonly int TestField2 = 2;
-    public static int TestField3;
-    public readonly int TestField4;
-    public int TestField5;
+    public static readonly int TestField2 = 1;
+    public static int TestField3 = 1;
+    public readonly int TestField4 = 1;
+    public int TestField5 = 1;
+    internal const int TestField6 = 1;
+    internal static readonly int TestField7 = 1;
+    internal static int TestField8 = 1;
+    internal readonly int TestField9 = 1;
+    internal int TestField10 = 1;
+    protected internal const int TestField11 = 1;
+    protected internal static readonly int TestField12 = 1;
+    protected internal static int TestField13 = 1;
+    protected internal readonly int TestField14 = 1;
+    protected internal int TestField15 = 1;
+    protected const int TestField16 = 1;
+    protected static readonly int TestField17 = 1;
+    protected static int TestField18 = 1;
+    protected readonly int TestField19 = 1;
+    protected int TestField20 = 1;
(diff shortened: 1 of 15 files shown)


sharwell:
@oatkins Can you review the tests included in this for accuracy?

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1204UnitTests.cs
```diff
@@ -22,16 +22,66 @@ public class SA1204UnitTests : DiagnosticVerifier
...
+    protected static int TestField18 = 1;
+    protected readonly int TestField19 = 1;
+    protected int TestField20 = 1;
+    private const int TestField21 = 1;
+    private static readonly int TestField22 = 1;
+    private static int TestField23 = 1;
+    private readonly int TestField24 = 1;
+    private int TestField25 = 1;
+
+    public TestClass2()
```
oatkins: There is another case not covered here.

image

A static class must be before a non-static one. Interestingly, this rule also applies to classes that are not nested. I suspect (though I didn't check) that the current implementation only works on children of a type.

oatkins:
The tests look good.

It looks as though we have a large block of correctly ordered code that is used in the tests for multiple rules. Perhaps this could be refactored into a common location? It also seems to be missing a nested static class, for completeness.

I'm going to merge this branch into the version I'll use on my machine today, so I should find any remaining gremlins pretty quickly.

Noryoko:
Good catch. I'll get those added.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/AccessLevel.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace StyleCop.Analyzers.OrderingRules
```
sharwell: :bulb: Consider moving this to `StyleCop.Analyzers.Helpers`.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/AccessLevelHelper.cs
```diff
@@ -0,0 +1,81 @@
+﻿namespace StyleCop.Analyzers.OrderingRules
```
sharwell: :bulb: Consider moving this to `StyleCop.Analyzers.Helpers`.

sharwell:
:+1:

PR #1135, opened 2015-08-06
Enable tests for code fix behavior with alternate indentation

sharwell:
I was able to address the original reliability problems by disabling test parallelization. It did not appear to have an effect on the overall build and test time.


Noryoko:
:+1:

PR #1146, opened 2015-08-07
Add statistics to StyleCopTester

pdelvo:


pdelvo:
I fixed the build break

StyleCop.Analyzers/StyleCopTester/Statistic.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal struct Statistic
+    {
+        public Statistic(int numberOf1, int numberOf2, int numberOf3)
+        {
+            this.NumberofNodes = numberOf1;
+            this.NumberOfTokens = numberOf2;
+            this.NumberOfTrivia = numberOf3;
+        }
+
+        public int NumberofNodes { get; set; }
```
sharwell: These properties don't need setters.

StyleCop.Analyzers/StyleCopTester/Statistic.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace StyleCopTester
+{
+    internal struct Statistic
+    {
+        public Statistic(int numberOf1, int numberOf2, int numberOf3)
```
sharwell: :bulb: These parameter names could be... improved :expressionless:

pdelvo:
I implemented the CR feedback

PR #1152, opened 2015-08-08
Fix comments following documentation incorrectly reported

### LINKED ISSUE #1132: SA1515 (Single line comment must be preceded by blank line) and SA1506 (A documentation header must not be followed by a blank line) require wrong spacing
In legacy StyleCop, the following is correct:

    /// <summary>Some summary.</summary>
    // Some comment.
    public void A()
    {
    }

Introducing a space between the two comment lines would cause a violation of SA1506.

In this project, the above code gives SA1515. Introducing a space between the two comment lines, which is wrong by legacy StyleCop, removes the analyzer warning.

See also #1130 and #1131.

Incidentally, I don't really *like* this style, but I have *many* violations because of the new analyzer behaviour. The primary reason is that we have many comments like

    /// <summary>Doc comments here...</summary>
    // ReSharper disable once XXXX
    public void SomeMethodWithReSharperSuppression()
    {
    }

where the spacing has been chosen to avoid SC violations. It's a pain that these now give analyzer warnings.

Noryoko:
- [x] Add regression tests for #1132
- [x] Fixes #1132

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1506UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1506UnitTests.cs
@@ -548,6 +548,48 @@
             await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Verifies that documentation followed by comments are handled properly.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


sharwell:
:bulb: Can you include the following in your test as well?

```csharp
/** <summary>more documentation.</summary> */
/* another comment */
public void TestMethod2() { }
```

sharwell:
I'll leave that test for a separate issue actually.

PR #1153, opened 2015-08-08
StyleCopTester should only analyze C# projects in a solution

sharwell:
Was receiving a large number of AD0001 reports due to attempting to analyze VB projects in **Roslyn.sln**.


Noryoko:
:+1:

PR #1157, opened 2015-08-08
Reduce the number of memory allocations in SA1404 (CodeAnalysisSuppressionMustHaveJustification)

sharwell:
* Avoid semantic analysis when simple name analysis will work. This step reuses cached information from SA1121 (UseBuiltInTypeAlias).
* Avoid calling `GetTypeByMetadataName` for every attribute in the compilation. Instead, this method is called once per compilation and cached.

When analyzing StyleCopAnalyzers, SA1404 was the most expensive analyzer in terms of both number of memory allocations and total bytes allocated. The following summary shows the improvement:

| Commit | Inclusive Allocations | Inclusive Bytes |
| --- | --- | --- |
| 6e92da965a17c7885d736bc5f33a56db89041f9b | 545065 | 18977452 |
| 4ae9b5e7f559dd01977d42de8f047bd87368366e | 113 | 10304 |

**Edit:** Analysis now run with the instrumenting profiler to verify accuracy.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1404UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1404UnitTests.cs
@@ -27,6 +27,22 @@
         }
 
         [Fact]
+        public async Task TestSuppressionWithStringLiteralAndUsingAliasDirectiveAsync()
+        {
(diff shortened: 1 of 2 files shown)


pdelvo:
Thats a huge improvement! I just looked at our unit tests for SA1404 and it does not have any test that uses aliases. We probably want some.

sharwell:
Added some tests

pdelvo:
:+1:

PR #1163, opened 2015-08-08
Fixes that the SA1412 fix all provider fixes violations with differen…

### LINKED ISSUE #1159: Equivalence key for SA1412 (StoreFilesAsUtf8) should depend on source encoding
Some encodings are not as easy to convert and require manual checking of special characters afterwards. This strategy will bring such files to the user's attention.

pdelvo:
…t encodings

Fixes #1159.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/MaintainabilityResources.resx
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/MaintainabilityResources.resx
@@ -133,6 +133,6 @@
     <value>Remove parentheses</value>
   </data>
   <data name="SA1412CodeFix" xml:space="preserve">
-    <value>Fix encoding</value>
+    <value>Change encoding from '{0}' to UTF-8 with byte order mark</value>
   </data>
 </root>
--- a/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1412FixAllProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1412FixAllProvider.cs
@@ -36,7 +36,20 @@
                 return null;
             }
 
-            return CodeAction.Create(MaintainabilityResources.SA1412CodeFix, token => Task.FromResult(newSolution));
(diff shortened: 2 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/MaintainabilityResources.resx
```diff
@@ -133,6 +133,6 @@
     <value>Remove parentheses</value>
   </data>
   <data name="SA1412CodeFix" xml:space="preserve">
-    <value>Fix encoding</value>
+    <value>Change encoding from '{0}' to UTF-8 with byte order mark</value>
```
Noryoko: Should format stings be in resource files?

pdelvo: Is there a reason not to?

Noryoko: Never mind. Looks consistent with our other code fix providers. :+1:

Noryoko:
:+1:

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1412FixAllProvider.cs
```diff
@@ -36,34 +36,34 @@ internal sealed class SA1412FixAllProvider : FixAllProvider
...
-            var diagnostics = await fixAllContext.GetDocumentDiagnosticsAsync(document).ConfigureAwait(false);
+            Solution solution = project.Solution;
+
+            var documentIds = project.DocumentIds;
 
-            if (diagnostics.Length == 0)
+            foreach (var documentId in documentIds)
             {
-                return solution;
+                solution = await FixDocumentAsync(fixAllContext, solution.GetDocument(documentId)).ConfigureAwait(false);
```
Maxwe11: Could we process all documents in parallel via `Task.WhenAll`?

sharwell: It's possible, but *believe me when I say* the problems with this code fix are not the performance of this block.

PR #1167, opened 2015-08-09
Codefix sa1203 take2

### LINKED ISSUE #1104: Code fix: SA1203 (ConstantsMustAppearBeforeFields)
Implement a code fix for #63.

### LINKED ISSUE #1106: Code fix: SA1214 (StaticReadonlyElementsMustAppearBeforeStaticNonReadonlyElements)
Implement a code fix for #74.

### LINKED ISSUE #1113: Code fix: SA1215 (InstanceReadonlyElementsMustAppearBeforeInstanceNonReadonlyElements)
Implement a code fix for #75.

hvanbakel:
New branch, with the proper commits. Maybe the flag enum solution can work for other tests as well. I haven't implemented this in the analyzer, it's now just used for the code fix.

Fixes #1104 
Fixes #1106
Fixes #1113


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203CodeFixProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+            return SpecializedTasks.CompletedTask;
+        }
+
+        private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        {
+            var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+
+            while (true)
+            {
+                var typeDeclarationNode =
```
Maxwe11: 1. this could be null, we need to check it before using.
2. this should be moved out of loop because `diagnostic.Location.SourceSpan` not changed inside loop.

@sharwell PTAL is it right behavior for codefix: fix all occurrences at one activation.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1203UnitTests.cs
```diff
@@ -145,11 +169,126 @@ public class Test
...
             yield return new SA1203ConstantsMustAppearBeforeFields();
         }
+
+        protected override CodeFixProvider GetCSharpCodeFixProvider()
+        {
+            return new SA1203CodeFixProvider();
+        }
     }
-}
+}
```
Maxwe11: bring back new line

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+    using System.Linq;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using StyleCop.Analyzers.Helpers;
+
+    /// <summary>
+    /// Helper class for dealing with member priority.
+    /// </summary>
+    public class MemberOrderHelper
```
Maxwe11: :question:  Where it could be used except SA1203

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203CodeFixProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+                    syntaxRoot.FindNode(diagnostic.Location.SourceSpan).FirstAncestorOrSelf<TypeDeclarationSyntax>();
+
+                var fieldReplaced = false;
+
+                var allFields = GetOrderHelperForFields(typeDeclarationNode);
```
Maxwe11: for same declaration we will allocate new list with same fields

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
hvanbakel: On all the other places where member order is used? You can now compare 2 members and get a result which indicates which one should come first?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203CodeFixProvider.cs
hvanbakel: 1. that's theoretically impossible but I'll add the check (you're referring to a field which is not in a class?
2. this cannot be moved outside of the loop because while the sourcespan doesn't cahnge, `syntaxRoot` does, after a move of a field this is a new tree so we need to get the typedeclaration from the current tree

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203CodeFixProvider.cs
hvanbakel: No, new tree, so new instances of fields

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+                if (Has(syntax, SyntaxKind.ReadOnlyKeyword))
+                {
+                    flags |= ModifierFlags.Readonly;
+                }
+            }
+
+            return flags == 0 ? ModifierFlags.None : flags;
+        }
+
+        private static bool Has(SyntaxTokenList list, SyntaxKind modifier)
```
sharwell: :exclamation: This method is a duplicate of the `SyntaxTokenList.Any(SyntaxKind)` method.

```
SyntaxTokenList list = ...;
SyntaxKind kind = ...;
bool has = list.Any(kind);
```

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+            /// Readonly modifier
+            /// </summary>
+            Readonly = 1 << 3,
+            /// <summary>
+            /// No modifiers
+            /// </summary>
+            None = 1 << 4
+        }
+
+        private enum AccesibilityFlags
```
sharwell: :exclamation: Just use the `Microsoft.CodeAnalysis.Accessibility` enumeration.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+            /// </summary>
+            Static = 1 << 2,
+            /// <summary>
+            /// Readonly modifier
+            /// </summary>
+            Readonly = 1 << 3,
+            /// <summary>
+            /// No modifiers
+            /// </summary>
+            None = 1 << 4
```
sharwell: :bulb: `None` should have the value 0.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// </value>
+        public MemberDeclarationSyntax Member { get; }
+
+        /// <summary>
+        /// The priority for this member.
+        /// </summary>
+        /// <value>
+        /// The priority for this member.
+        /// </value>
+        public int Priority => (int)this.modifierFlags + (int)this.accessibiltyFlags;
```
sharwell: :exclamation: This means `public static int y` has the same priority as `protected internal const int z`?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+        public int Priority => (int)this.modifierFlags + ((int)this.accessibilty * 100/*the * 100 ensures the accesibility is more important than the modifier*/);
+
+        /// <summary>
+        /// Checks if the current field should be declared before the other.
+        /// </summary>
+        /// <param name="other">The field to compare against.</param>
+        /// <returns>True if the field should be delcared before the other.</returns>
+        public bool ShouldBeBefore(MemberOrderHelper other) => this.Priority > other.Priority;
+
+        private static Accessibility GetAccessibilityFlags(SyntaxTokenList syntax)
```
Noryoko: @sharwell Not sure I agree with using `Microsoft.CodeAnalysis.Accessibility`. Below are it's values which do not match the order of StyleCop.

```
NotApplicable = 0
Private = 1
ProtectedAndInternal = 2
Protected = 3
Internal = 4
ProtectedOrInternal = 5
Public = 6
```

Ending up making `AccessLevel` and `AccessLevelHelper` in `StyleCop.Analyzers.Helpers` for determining access levels. `AccessLevelHelper.GetAccessLevel` could be used in place of this method.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+                if (syntax.Any(SyntaxKind.ReadOnlyKeyword))
+                {
+                    flags |= ModifierFlags.Readonly;
+                }
+            }
+
+            return flags == 0 ? ModifierFlags.None : flags;
+        }
+    }
+}
```
Noryoko: Add a new line please.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203CodeFixProvider.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            var trackedRoot = root.TrackNodes(field, firstNonConst);
+            var fieldToMove = trackedRoot.GetCurrentNode(field);
+            var firstNonConstTracked = trackedRoot.GetCurrentNode(firstNonConst);
+            root = trackedRoot.InsertNodesBefore(firstNonConstTracked, new[] { fieldToMove });
+            var fieldToMoveTracked = root.GetCurrentNodes(field).Last();
+            return root.RemoveNode(fieldToMoveTracked, SyntaxRemoveOptions.KeepNoTrivia);
+        }
+    }
+}
```
Noryoko: Add a new line please.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                return Accessibility.ProtectedAndInternal;
+            }
+
+            if (syntax.Any(SyntaxKind.ProtectedKeyword))
+            {
+                return Accessibility.Protected;
+            }
+
+            // Do not assign private accessiblity here, since we ant to have private come after non qualified.
```
Noryoko: In the case of consts non qualified members are considered private.

sharwell: This is the case for all members of a `struct` or `class`. For types within a namespace, the accessibility is `internal`.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
hvanbakel: Added it back, makes more sense to have it separate especially since the order is not as you would want it to be.

hvanbakel:
Fixed #1106 as well, using the same helper class

sharwell:
@hvanbakel I haven't reviewed the code, but I wanted to talk a bit about the way it behaves. In my mind, the behavior to fix a single violation on member `MemberName` would be to move `MemberName` *up* until it reaches a position where the violation is no longer reported. Does this describe the way you have currently implemented it?

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1203UnitTests.cs
```diff
@@ -145,11 +169,126 @@ public class Test
...
+    public const string Before2 = ""test"";
+
+    public const string After2 = ""test"";
+
+    public int between;
+    private const string Before1 = ""test"";
+
+    private const string After1 = ""test"";
+
+    private int field1;
```
Noryoko: While I like the final order, I'm thinking we should only move the fields that have diagnostics reported. I propose the resulting order would be this.

```
private const string Before1 = "test";
public const string Before2 = "test";
public const string After2 = "test"; // moved up until no longer in violation
private const string After1 = "test"; // moved up until no longer in violation
private int field1;
public int between;
```

@sharwell @hvanbakel What do you think of this? I believe that fixing the other violations should be handled when fixing SA1202 violations and are out of scope for SA1203.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// </value>
+        public MemberDeclarationSyntax Member { get; }
+
+        /// <summary>
+        /// The priority for this member.
+        /// </summary>
+        /// <value>
+        /// The priority for this member.
+        /// </value>
+        public int Priority => (int)this.modifierFlags + ((int)this.accessibilty * 100/*the * 100 ensures the accesibility is more important than the modifier*/);
```
Maxwe11: formatting :-1:

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,169 @@
...
+        public MemberOrderHelper(MemberDeclarationSyntax member)
+        {
+            this.Member = member;
+            var modifiers = member.GetModifiers();
+
+            this.modifierFlags = GetModifierFlags(modifiers);
+            this.accessibilty = GetAccessibilityFlags(modifiers);
+        }
+
+        private enum AccesibilityFlags
```
Noryoko: :blub: Use `StyleCop.Analyzers.Helpers.AccessLevel`.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// The priority for this member only from modifiers.
+        /// </summary>
+        /// <value>
+        /// The priority for this member.
+        /// </value>
+        public int ModifierPriority => (int)this.modifierFlags;
+
+        private static AccesibilityFlags GetAccessibilityFlags(SyntaxTokenList syntax)
+        {
+            if (syntax.Any(SyntaxKind.PublicKeyword))
```
Noryoko: :blub: Use `StyleCop.Analyzers.Helpers.AccessLevelHelper.GetAccessLevel(member). In the case of `AccessLevel.NotSpecified` return `AccessLevel.Private` instead. The code below contains a bug that will not properly handle `protected internal`.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203SA1214CodeFixProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (typeDeclarationNode == null)
+            {
+                return document;
+            }
+
+            var allFields = GetOrderHelperForFields(typeDeclarationNode);
+            var fieldToMove = new MemberOrderHelper(fieldDeclaration);
+            for (var i = 0; i < allFields.Count; i++)
+            {
+                if (allFields[i].ModifierPriority < fieldToMove.ModifierPriority)
```
Noryoko: :bulb: Get the index of `fieldDeclaration` in `typeDeclarationNode.Members` and iterate backwards over members. Instead of enumerating the entire collection in `GetOrderHelperForFields`.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
hvanbakel: But this has not specified defined as being above public?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203SA1214CodeFixProvider.cs
hvanbakel: I implemented everything except the 'iterate backwards' as that would put the field above the first field which it should be above. However, that does not mean it will get rid of the violation if there's another field above it.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,135 @@
...
+        private static AccessLevel GetAccessibilityFlags(SyntaxTokenList syntax)
+        {
+            if (syntax.Any(SyntaxKind.PublicKeyword))
+            {
+                return AccessLevel.Public;
+            }
+
+            if (syntax.Any(SyntaxKind.InternalKeyword))
+            {
+                return AccessLevel.Internal;
```
Noryoko: :exclamation: `protected internal` will always return here.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203SA1214CodeFixProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            return document.WithSyntaxRoot(syntaxRoot);
+        }
+
+        private static SyntaxNode MoveField(FieldDeclarationSyntax fieldDeclaration, TypeDeclarationSyntax typeDeclarationNode, SyntaxNode syntaxRoot)
+        {
+            var fieldToMove = new MemberOrderHelper(fieldDeclaration);
+            for (var i = 0; i < typeDeclarationNode.Members.Count; i++)
+            {
+                var member = typeDeclarationNode.Members[i];
+                if (member.Kind() != SyntaxKind.FieldDeclaration)
```
Noryoko: :bulb: `!member.IsKind(SyntaxKind.FieldDeclaration)`

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203SA1214CodeFixProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            syntaxRoot = MoveField(fieldDeclaration, typeDeclarationNode, syntaxRoot);
+
+            return document.WithSyntaxRoot(syntaxRoot);
+        }
+
+        private static SyntaxNode MoveField(FieldDeclarationSyntax fieldDeclaration, TypeDeclarationSyntax typeDeclarationNode, SyntaxNode syntaxRoot)
+        {
+            var fieldToMove = new MemberOrderHelper(fieldDeclaration);
+            for (var i = 0; i < typeDeclarationNode.Members.Count; i++)
```
Noryoko: :bulb: Could be simplified to `foreach (var member in typeDeclarationNode.Members)`. I only see the `i` used at the beginning to get the member.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1203UnitTests.cs
```diff
@@ -145,11 +169,137 @@ public class Test
...
+
+            var fixedTestCode = @"public class Foo
+{
+    private const string Before1 = ""test"";
+
+    public const string Before2 = ""test"";
+
+    private const string After1 = ""test"";
+
+    public const string After2 = ""test"";
```
Noryoko: :question: Shouldn't this be moved to before `After1`?

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -96,6 +96,15 @@ public class Foo
...
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            var fixTestCode = @"
+public class Foo
+{
+    private static readonly int j = 0;
+    private static int i = 0;
+}";
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
```
Noryoko: Verify `fixTestCode`.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -126,6 +135,14 @@ public class Foo
             };
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            var fixTestCode = @"
+public class Foo
+{
+    private static readonly int j = 0;
+    private static int i = 0;}";
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
```
Noryoko: Verify `fixTestCode`.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -170,6 +187,15 @@ public struct Foo
...
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            var fixTestCode = @"
+public struct Foo
+{
+    private static readonly int j = 0;
+    private static int i = 0;
+}";
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
```
Noryoko: Verify `fixTestCode`.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -207,6 +233,31 @@ public class FooInner
...
+
+    public class FooInner 
+    {
+        private static readonly int e = 1;
+        private int aa = 0;
+        public static readonly int t = 2;
+        private static int z = 999;
+    }
+}";
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
```
Noryoko: Verify `fixTestCode`.

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -207,6 +233,31 @@ public class FooInner
...
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            var fixTestCode = @"
+public class Foo
+{
+
+    public static readonly int  u = 5;
+
+    public static readonly int j = 0;
```
Noryoko: :question: Why was this moved when no diagnostic was reported by SA1214?

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1214UnitTests.cs
```diff
@@ -207,6 +233,31 @@ public class FooInner
...
+    public string s = ""qwe"";
+    private static readonly int i = 0;
+
+    public void Ff() {}
+
+    public static string s2 = ""qwe"";
+
+    public class FooInner 
+    {
+        private static readonly int e = 1;
```
Noryoko: :question:  Shouldn't this be after `public static readonly int t = 2;`?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203SA1214CodeFixProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using StyleCop.Analyzers.Helpers;
+
+    /// <summary>
+    /// Implements code fixes for <see cref="SA1203ConstantsMustAppearBeforeFields"/>.
```
Noryoko: Update the summary please.

hvanbakel:
Fix added for #1113

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
pdelvo: Does this need to be a class or would a struct work too?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
pdelvo: @Noryoko: protected internal has its own branch so it wouldnt return here? Or have I missed something?

Noryoko: Comment was for a previous commit. When internal alone was checked first.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/MemberOrderHelper.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            if (syntax.Any(SyntaxKind.ProtectedKeyword))
+            {
+                return AccessLevel.Protected;
+            }
+
+            return AccessLevel.Private;
```
pdelvo: This is not always correct. For types defined outisde of other types the default visibility is internal. We actually do have this method already [here](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/33c13dca8ac7e520b73e5a2eda03bbaac6588f42/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs#L25).

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1203SA1214SA1215CodeFixProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        private static SyntaxNode MoveField(SyntaxNode root, MemberDeclarationSyntax field, MemberDeclarationSyntax firstNonConst)
+        {
+            var trackedRoot = root.TrackNodes(field, firstNonConst);
+            var fieldToMove = trackedRoot.GetCurrentNode(field);
+            var firstNonConstTracked = trackedRoot.GetCurrentNode(firstNonConst);
+            if (!fieldToMove.HasLeadingTrivia)
+            {
+                fieldToMove = fieldToMove.WithLeadingTrivia(firstNonConstTracked.GetLeadingTrivia());
```
pdelvo: Wouldnt this also copy over documentation and comments?

pdelvo: I tried this and it transforms this
```csharp
class Foo
{
    /// <summary>
    /// Bar
    /// </summary>
    string bar;
const string foo = "a";
}
```
to this
```csharp
class Foo
{
    /// <summary>
    /// Bar
    /// </summary>
    const string foo = "a";
    /// <summary>
    /// Bar
    /// </summary>
    string bar;
}
```

I would just not worry about trivia here.

pdelvo:
:thought_balloon: We might want to fix the spacing between items if we move a field to be the first member. But i think this can be done in a different PR.

When the issue with the copied trivia is resolved I think this can be merged :+1:

hvanbakel:
Latest comments by @pdelvo are now fixed

pdelvo:
:+1:

sharwell:
Rebased and merged as commit 907c45fcebf6791ef436d3935280e6660f9b0639.

PR #1174, opened 2015-08-09
Fix order of using directives should reset at preprocessor directives

### LINKED ISSUE #1161: Order of using directives should reset at preprocessor directives
The following block results in warnings regarding the order of preprocessor directives:

```csharp
using System;
using Microsoft.VisualStudio;
using MyList = System.Collections.Generic.List<int>;

#if FORVS10 || FORVS12 || FORVS13 || FORVS15
using Microsoft.CodeAnalysis;
#else
using Microsoft.CodeAnalysis;
#endif
```

However, each time a preprocessor line is reached, the relative order of items should be reset. In other words, two `using` directives should not be compared to each other if a preprocessor line exists between them.

## Actual Result

Warnings SA1210 and SA1209 are reported.

## Expected result

No warnings.

## Impact

This affects the usability of SA1208, SA1209, SA1210, SA1211, SA1216, and SA1217.

Noryoko:
- [x] Add regression tests for #1161
- [x] Fixes #1161

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/UsingDirectiveSyntaxHelpers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/UsingDirectiveSyntaxHelpers.cs
@@ -34,23 +34,8 @@
 
             foreach (var trivia in usingDirective.GetLeadingTrivia())
             {
-                switch (trivia.Kind())
+                if (trivia.IsDirective)
                 {
-                case SyntaxKind.IfDirectiveTrivia:
-                case SyntaxKind.ElifDirectiveTrivia:
-                case SyntaxKind.ElseDirectiveTrivia:
-                case SyntaxKind.EndIfDirectiveTrivia:
-                case SyntaxKind.RegionDirectiveTrivia:
-                case SyntaxKind.EndRegionDirectiveTrivia:
-                case SyntaxKind.DefineDirectiveTrivia:
-                case SyntaxKind.UndefDirectiveTrivia:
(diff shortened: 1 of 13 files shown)


Noryoko:
@sharwell Please review the tests for accuracy.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/UsingDirectiveSyntaxHelpers.cs
```diff
@@ -20,6 +20,44 @@ internal static class UsingDirectiveSyntaxHelpers
...
+        internal static bool IsPrecededByPreprocessorDirective(this UsingDirectiveSyntax usingDirective)
+        {
+            if (!usingDirective.HasLeadingTrivia)
+            {
+                return false;
+            }
+
+            foreach (var trivia in usingDirective.GetLeadingTrivia())
+            {
+                switch (trivia.Kind())
```
sharwell: :bulb:

```csharp
if (trivia.IsDirective)
  return true;
```

sharwell:
:question: Can you add tests for SA1208, SA1211, SA1216, and SA1217 as well?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/UsingDirectiveSyntaxHelpers.cs
Noryoko: That certainly simplifies things.

Noryoko:
Will do.

PR #1180, opened 2015-08-10
Fix SA1202 wrongly treats explicit interface implementations as private

### LINKED ISSUE #1179: SA1202 wrongly treats explicit interface implementations as private
Legacy StyleCop correctly treats members that explicitly implement an interface member as public. The new analyzers, however, treat these as private.

    internal interface IA
    {
        string T { get; }

        int U { get; }
    }

    class Test : IA
    {
        public int Q { get; set; }

        string IA.T
        {
            get { return null; }
        }

        // SA1202 (All public properties must come before all private properties) wrongly reported here.
        public int W { get; set; }

        protected string S { get; set; }

        // SA1202 should be reported here (according to legacy StyleCop), but is not.
        int IA.U
        {
            get { return 0; }
        }
    }

Interestingly, explicit implementations are treated as public by legacy StyleCop even when they implement an internal interface. (I used an internal interface in my example; the results are the same if you make it public.)

Sorry I missed this subtlety before.

Noryoko:
- [x] Add regression tests for SA1202
- [x] Fixes #1179

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1202UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1202UnitTests.cs
@@ -542,6 +542,72 @@ class MyClass
         }
 
         /// <summary>
+        /// Verifies that the analyzer will properly handle explicit interface implementations.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


Noryoko:
@oatkins Please review the unit test for accuracy.

sharwell:
Assigning to @oatkins for test review :smile:

oatkins:
:+1:

Issue #1172, opened 2015-08-09
SA1501 code fix does not indent while clause of do...while loop


sharwell:
## Input

```csharp
    do { Operation(); } while (false);
```

## Expected result

```csharp
    do
    {
        Operation();
    }
    while (false);
```

## Actual result

```csharp
    do
    {
        Operation();
    }
while (false);
```

hvanbakel:
Grabbed

PR #1182, opened 2015-08-11
Fix for 1172

hvanbakel:
PR for #1172

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
@@ -451,13 +451,42 @@ public class Foo
 }
 ";
 
-            DiagnosticResult[] expectedDiagnostics =
-            {
-                this.CSharpDiagnostic().WithLocation(7, 12)
-            };
+            DiagnosticResult expectedDiagnostics = this.CSharpDiagnostic().WithLocation(7, 12);
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expectedDiagnostics, CancellationToken.None).ConfigureAwait(false);
             await this.VerifyCSharpDiagnosticAsync(fixedTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Test for issue 1172, https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1172
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestRegressionIssue1172WithHalfClassAsync()
+        {
+            var testCode = @"using System;
+
+class TypeName
+{
+    void MethodName()
+    {
+        do { Bar(); } while (false);
+";
+
+            var fixedTestCode = @"using System;
+
+class TypeName
+{
+    void MethodName()
+    {
+        do
+        {
+            Bar();
+        }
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1501CodeFixProvider.cs
```diff
@@ -75,8 +75,19 @@ private static SyntaxNode ReformatBlockAndParent(Document document, SyntaxNode s
                 newParentLastToken = newParentLastToken.WithTrailingTrivia(newTrailingTrivia);
             }
 
+            var parentNextToken = block.CloseBraceToken.GetNextToken();
+
+            var nextTokenLine = parentNextToken.GetLineSpan().StartLinePosition.Line;
```
sharwell: :bulb: This is available via an extension method in `LocationHelpers`:

```csharp
parentNextToken.GetLine();
```

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1501CodeFixProvider.cs
```diff
@@ -75,8 +75,19 @@ private static SyntaxNode ReformatBlockAndParent(Document document, SyntaxNode s
...
+            var blockCloseLine = block.CloseBraceToken.GetLineSpan().EndLinePosition.Line;
```
sharwell: :bulb:

```csharp
block.CloseBraceToken.GetEndLine();
```

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1501CodeFixProvider.cs
```diff
@@ -75,8 +75,19 @@ private static SyntaxNode ReformatBlockAndParent(Document document, SyntaxNode s
                 newParentLastToken = newParentLastToken.WithTrailingTrivia(newTrailingTrivia);
             }
 
+            var parentNextToken = block.CloseBraceToken.GetNextToken();
```
sharwell: :question: What happens if `CloseBraceToken` is the last token in the file? For example, what happens if the following code is the entire file?

```csharp
class TypeName
{
  void MethodName()
  {
    if (false) { return; }
```

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -418,6 +418,49 @@ public void Bar()
...
+        do
+        {
+            Bar();
+        }
+        while (false);
+    }
+}
+";
+
+            DiagnosticResult[] expectedDiagnostics =
```
sharwell: :bulb: You only need an array for tests producing multiple diagnostics.

```csharp
DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(7, 12);
```

sharwell:
I left a question and a few comments in the diff.

In addition, can you amend this to use your name instead of "connected-web-development" as the author of the commit(s)? Thanks!

sharwell:
@hvanbakel I marked this **do not merge** just so it's easy for me to see that I finished reviewing it and it's waiting for feedback. Once you have a look at the feedback let me know and I'll remove that label. :smile:

hvanbakel:
Looking at it right now

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1501CodeFixProvider.cs
hvanbakel: Added a test for it, works fine

sharwell:
Rebased and merged manually in 1498eaf672f2cf01d1620d0e2c6538f53f22b81f.

Thank you so much! :tada:

PR #1188, opened 2015-08-12
Fix SA1009 incorrectly reporting trailing comments

### LINKED ISSUE #1187: Code fix for SA1009 removes trivia
When the code fix for SA1009 is executed on the following call, the comment is removed.

```csharp
x.MethodName(
    arg1
    , _ => Assert.True(true) // comment
    , _ => Assert.True(true);
```

### Actual result

```csharp
testCallBack.AssertOnCallbackBeforeAndAfter(
    TimeOutMs
    , ev => Assert.True(true) , ev => Assert.AreEqual(isSuccess, ev.Success));
```

### Expected result

I'm actually not sure here. Perhaps no code fix offered? I'm open to suggestions.

Noryoko:
Diagnostic results were matched against classic StyleCop.

fixes #1187

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs
@@ -162,6 +162,18 @@
                 break;
             }
 
+            foreach (var trivia in token.TrailingTrivia)
+            {
+                if ((trivia.IsKind(SyntaxKind.SingleLineCommentTrivia) || trivia.IsKind(SyntaxKind.MultiLineCommentTrivia))
+                    && trivia.GetLineSpan().StartLinePosition.Line == token.GetLine())
+                {
+                    lastInLine = false;
+                    precedesStickyCharacter = false;
+
+                    break;
+                }
+            }
+
             if (precededBySpace)
             {
                 // Closing parenthesis must{ not} be {preceded} by a space.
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -162,6 +162,18 @@ private static void HandleCloseParenToken(SyntaxTreeAnalysisContext context, Syn
                 break;
             }
 
+            foreach (var trivia in token.TrailingTrivia)
+            {
+                if ((trivia.IsKind(SyntaxKind.SingleLineCommentTrivia) || trivia.IsKind(SyntaxKind.MultiLineCommentTrivia))
```
sharwell: :question: Does this work? If so it would improve performance by not needing to call `GetLineSpan()`.

```csharp
foreach (var trivia in token.TrailingTrivia)
{
  if (trivia.IsKind(SyntaxKind.EndOfLine))
  {
    break;
  }
  else if (trivia.IsKind(SyntaxKind.SingleLineCommentTrivia)
    || trivia.IsKind(SyntaxKind.MultiLineCommentTrivia))
  {
    lastInLine = false;
    precedesStickyCharacter = false;
    break;
  }
}
```

Noryoko: :+1:  Always a fan of performance gains.

PR #1189, opened 2015-08-12
Update behavior of SA1501 when SA1503 is disabled

### LINKED ISSUE #1175: Proposal: Special case behavior for SA1501 when SA1503 is suppressed
## Background

Currently, code like the following violates SA1503:

```csharp
if (x) return;
```

To fix this violation, the user adds braces. The following "simplest" addition of braces violates SA1501:

```csharp
if (x) { return; }
```

To resolve both violations, the user has to write the statement as follows:

```csharp
if (x)
{
    return;
}
```

## Problem

Teams that allow simple statements to be written without braces may disable SA1503. When SA1503 is disabled, the original snippet above does not report any violation at all.

## Proposal

When SA1503 is disabled, I propose modifying the behavior of SA1501 to disallow the following statement:

```csharp
if (x) return;
```

To resolve the issue, the code fix should move the return statement to the next line:

```csharp
if (x)
    return;
```

:memo: The documentation for SA1501 would need to be updated. The title and message format already make sense with this proposal, but the Cause section of the documentation specifically mentions braces as the item of interest.

sharwell:
This commit updates SA1501 to include special handling for the case where
SA1503 is disabled. In this scenario, statements are analyzed instead of
curly braces.

Fixes #1175

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
@@ -539,7 +539,7 @@ class TypeName
             this.suppressSA1503 = true;
 
             var testCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
@@ -548,7 +548,7 @@ public class Foo
     }
 }";
             var fixedCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
@@ -589,7 +589,7 @@ public class Foo
             this.suppressSA1503 = true;
 
             var testCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
@@ -616,7 +616,7 @@ public class Foo
             this.suppressSA1503 = true;
 
             var testCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
@@ -643,7 +643,7 @@ public class Foo
             this.suppressSA1503 = true;
 
             var testCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
@@ -652,7 +652,7 @@ public class Foo
 }";
 
             var fixedTestCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
@@ -677,7 +677,7 @@ else
             this.suppressSA1503 = true;
 
             var testCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
  public void Bar(int i)
  {
@@ -686,7 +686,7 @@ public class Foo
 }";
 
             var fixedTestCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
  public void Bar(int i)
  {
@@ -708,7 +708,7 @@ public class Foo
             this.suppressSA1503 = true;
 
             var testCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
@@ -719,7 +719,7 @@ public class Foo
 }";
 
             var fixedTestCode = @"using System.Diagnostics;
-public class Foo
+public class TypeName
 {
     public void Bar(int i)
     {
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -490,15 +512,339 @@ void MethodName()
...
+}";
+
+            var fixedTestCode = @"using System.Diagnostics;
+public class Foo
+{
+    public void Bar(int i)
+    {
+        if (i == 0)
+            Debug.Assert(true);
+else
```
sharwell: :memo: Not sure yet what caused this.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -490,15 +512,339 @@ void MethodName()
...
+}";
+
+            var batchFixedTestCode = @"using System.Diagnostics;
+public class Foo
+{
+    public void Bar(int i)
+    {
+        if (i == 0)
+            if (i == 0)
+    Debug.Assert(true);
```
sharwell: :memo: Not sure yet what causes this result from the batch fixer.

sharwell:
There are a couple remaining buggy behaviors in the code fix for this, but it does seem to work decently well overall. I'm a bit short on time so I configured the tests to detect the bugs as the expected behavior and left comments for review. Let me know whether you'd like to see them fixed before merging or if bug reports would be acceptable.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
Noryoko: :question: What happens in this case?
```
if (i == 0)
    Debug.Assert(true); else Debug.Assert(false);
```

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
Noryoko: :bulb: Before the second `if` is rewritten it has no leading trivia. `IndentationHelper.GetIndentationSteps` would then produce `0` and only one indent is done.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
sharwell: In the updated implementation, I believe this does not report SA1501 at all.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
sharwell: :arrow_right: I fixed this by only reporting (and thus fixing) the first instance of SA1501 on a line (still for the specific case where SA1503 is disabled).

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -490,15 +512,414 @@ void MethodName()
...
+public class TypeName
+{
+    public void Bar(int i)
+    {
+        if (i == 0)
+            if (i == 0) Debug.Assert(true);
+    }
+}";
+
+            var batchFixedTestCode = @"using System.Diagnostics;
```
Noryoko: Since `batchFixedTestCode` now matches `fixedTestCode` should it be removed?

Noryoko:
:question: Can you add tests for `else` and `else if` statements?

Noryoko:
```
if (i == 0)
    Debug.Assert(true);
else Debug.Assert(false);
```

```
if (i == 0) Debug.Assert(true); else if (i == 0) Debug.Assert(false);
```

```
if (i == 0)
    Debug.Assert(true);
else if (i == 0) Debug.Assert(false);
```

```
if (i == 0) Debug.Assert(true); else if (i == 0) Debug.Assert(false); else Debug.Assert(false);
```

```
if (i == 0) if (i == 0) Debug.Assert(false); else Debug.Assert(false); else Debug.Assert(true);
```

```
if (i == 0)
    if (i == 0) Debug.Assert(false); else Debug.Assert(false); else Debug.Assert(true);
```

```
if (i == 0) if (i == 0) Debug.Assert(false);
    else Debug.Assert(false); else if (i == 0) Debug.Assert(true);
```

Noryoko:
:+1:

PR #1196, opened 2015-08-13
Disable XML analysis when documentation diagnostics are disabled

### LINKED ISSUE #1190: False positive of SA1601 on windows forms - if documentation build XML isn't eanbled.
Just create a new windows forms project under VS2015 and add the StyleCop.Analyzers analyzer.
Resolve all warnings in the IDE by adaption of the code - but do not activate the documentation build.
If you now build the project you will face a additional warning like the following in the build output:

1>C:\temp\TestStyleCopAnalyzers\TestStyleCopAnalyzers\Form1.cs(14,24,14,29): warning SA1601: Partial elements must be documented

Also it seems that the missing documentation build raises a lot of other false positives.
If this is a know behavior please close this issue immediately - and maybe add this fact to the documentation more prominent.

sharwell:
* Fix #1190
* Fix previously unreported issue where enumeration members would receive warnings when documentation diagnostics were disabled


Noryoko:
:+1:

PR #1203, opened 2015-08-13
Fixed issue in CodeFixVerifier

vweijsters:
Fixed an issue where the incremental code fix verifier would terminate early because the new diagnostics set and the old diagnostics set only differed by location.

Also corrected the failing tests in SA1501UnitTests as a result of this bug fix.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
@@ -642,37 +642,8 @@ public class TypeName
             else Debug.Assert(false); else if (i == 13) Debug.Assert(true);//24
     }
 }";
+
             var fixedCode = @"using System.Diagnostics;
-public class TypeName
-{
-    public void Bar(int i)
-    {
-        if (i == 0)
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -746,7 +717,7 @@ public void Bar(int i)
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
             ////await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
-            await this.VerifyCSharpFixAsync(testCode, fixedCode, batchFixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
```
sharwell: :bulb: You can just remove the third parameter. It assumes the batch fixer output matches the incremental fixer if you don't specify something for `batchNewSource`.

PR #1204, opened 2015-08-13
Fix SA1500 wrongly raised for method following object initializer

### LINKED ISSUE #1202: SA1500 (Curly brackets for multi-line statements must not share a line) wrongly raised for extension method on a collection with a collection initializer
The following code gives SA1500 with the analyzer but not with legacy SC:

    public void Fish()
    {
        var someList = new List<int>
                       {
                           1,
                           2,
                           3
                       }.Select(x => x.ToString());
    }

Noryoko:
- [x] Add regression tests
- [x] Fix SA1500 wrongly raised for method following object initializer

Fixes #1202

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CurlyBracketsForMultiLineStatementsMustNotShareLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CurlyBracketsForMultiLineStatementsMustNotShareLine.cs
@@ -205,7 +205,6 @@
                 case SyntaxKind.CommaToken:
                 case SyntaxKind.SemicolonToken:
                 case SyntaxKind.DotToken:
-                case SyntaxKind.QuestionToken:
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CurlyBracketsForMultiLineStatementsMustNotShareLine.cs
```diff
@@ -204,6 +204,8 @@ private static void CheckCurlyBracketToken(SyntaxNodeAnalysisContext context, Sy
                 case SyntaxKind.CloseParenToken:
                 case SyntaxKind.CommaToken:
                 case SyntaxKind.SemicolonToken:
+                case SyntaxKind.DotToken:
+                case SyntaxKind.QuestionToken:
```
sharwell: :exclamation: I don't think we need to include `QuestionToken`. An object creation expression will never result in `null`, so the null propagation operator would be pointless.

PR #1212, opened 2015-08-15
Fix SA1116 handling of omitted argument lists

### LINKED ISSUE #1211: SA1116 throws exception for attribute usage without arguments
The following code results in an exception in SA1116:

```csharp
[Obsolete]
void MethodName() { }
```

This is due to the fact that the use of `Obsolete` does not include an argument list, a situation not currently handled by the analyzer.

sharwell:
* [x] Add regression test for #1211
* [x] Fixes #1211


Maxwe11:
:+1:

PR #1213, opened 2015-08-15
Fix behavior of SA1009 code fix when non-whitespace trivia is included

### LINKED ISSUE #1206: SA1009 should not fire when it's the first character on a line
This is actually two fold, my code was this:
```CSharp
            if (
                symbol == null || // Symbol can be null here if Cast is to a generic parameter for example
                symbol.Equals(declaredType) // If the cast/as/is is to the type in which this statement is contained
                )
```

After applying the fix (on line 4) it became this
```CSharp
            if (
                symbol == null || // Symbol can be null here if Cast is to a generic parameter for example
                symbol.Equals(declaredType) // If the cast/as/is is to the type in which this statement is contained
)
```
Which still violates the rule.

The violation after the fix is a bug. 

For what it's worth, I reformatted it to this to get rid of the violation.
```CSharp
            if (// Symbol can be null here if Cast is to a generic parameter for example
                symbol == null ||

                // If the cast/as/is is to the type in which this statement is contained
                symbol.Equals(declaredType))
```

sharwell:
* [x] Add regression tests for #1206 
* [x] Fix #1206

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/OpenCloseSpacingCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/OpenCloseSpacingCodeFixProvider.cs
@@ -92,7 +92,63 @@
                     triviaList = prevToken.TrailingTrivia.AddRange(token.LeadingTrivia);
 
                     replaceMap[prevToken] = prevToken.WithTrailingTrivia();
-                    replaceMap[token] = token.WithLeadingTrivia(triviaList.WithoutTrailingWhitespace());
+                    if (triviaList.All(i => i.IsKind(SyntaxKind.WhitespaceTrivia) || i.IsKind(SyntaxKind.EndOfLineTrivia)))
+                    {
+                        replaceMap[token] = token.WithLeadingTrivia();
+                    }
+                    else if (token.IsFirstInLine() && token.IsLastInLine())
+                    {
+                        /* This block covers the case where `token` is the only non-trivia token on its line. However,
+                         * the line may still contain non-whitespace trivia which we want the removal process to
+                         * preserve. This code fix only removes the whitespace surrounding `token` if it is the only
+                         * non-whitespace token on the line.
+                         */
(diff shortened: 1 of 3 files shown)


sharwell:
@hvanbakel This was surprisingly complicated. The new logic should take good care of you and your trailing comments. :expressionless: :wink:

sharwell:
@pdelvo @Maxwe11 @Noryoko @vweijsters This update allowed the code fix provider to be used for four other code fixes as well. The pull request that implements those is currently blocked on this review.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/OpenCloseSpacingCodeFixProvider.cs
```diff
@@ -92,7 +92,63 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                            {
+                                trailingTrivia.AddRange(token.LeadingTrivia.Take(lastNewLineLeading + 1));
+                            }
+                            else
+                            {
+                                trailingTrivia.AddRange(token.LeadingTrivia);
+                            }
+
+                            // firstNewLineFollowing was adjusted above to account for the missing case.
+                            trailingTrivia.AddRange(token.TrailingTrivia.Take(firstNewLineFollowing));
```
vweijsters: I think we are missing test cases, as the code above is bugged. `trailingTrivia` is of type `SyntaxTriviaList`, which is an immutable type. That means that the output of the `AddRange` should have been assigned to `trailingTrivia` on lines 127, 131, and 135.

sharwell: I added new tests. If either of the fix two cases is not corrected, at least one of the new tests will fail. I haven't found a way to test the last case, but I fixed it.

Noryoko:
We should add tests for SA1011, SA1012, SA1013, SA1014, SA1015, SA1019, SA1020, and SA1023 since they use the same code fix.

vweijsters:
:+1: Looks good to me, except for the remark of @Noryoko above.

sharwell:
I added issue #1217 for the additional tests.

PR #1216, opened 2015-08-16
Fix SA1514 should not be reported when preceded by a comment at the end of a line

Noryoko:
- [x] Add regression tests for #1215
- [x] Fixes #1215

Issue #1131, opened 2015-08-06
SA1514 (A documentation comment must be preceded by a blank line...) not reported when immediately preceded by a single line comment


oatkins:
The following should, but does not, generate an analyzer warning. (It does in legacy StyleCop.)

    // Something in advance.
    /// <summary>
    /// The YetMore class.
    /// </summary>
    public static class YetMore { }

The above code should be "fixed" by inserting a blank line before the documentation comments, though this would currently add a violation of SA1512 without a fix for #1130.

Noryoko:
Grabbing this.

PR #1219, opened 2015-08-16
Fixed issue with SA1514

vweijsters:
This fixes an issue where a documentation header that is directly preceded by a single comment line that in its turn is preceded by one or more empty lines will not produce the expected diagnostic.

This fixes an issue that was introduced in #1150 and #1216.
It also fixes the root cause for #1131


Noryoko:
:+1:

PR #1226, opened 2015-08-17
Use StringComparison.Ordinal in calls to string.StartsWith

sharwell:
This change resulted in a 9.9% performance improvement for the analysis of Roslyn.sln due to reduced lock contention and a more efficient string comparison operation.


Noryoko:
:+1:

PR #1227, opened 2015-08-17
Avoid boxing SyntaxTriviaList in calls to Any and All with a predicate

sharwell:
This improves the performance of `OpenCloseSpacingCodeFixProvider` and a few other smaller cases.


Noryoko:
:+1:

PR #1232, opened 2015-08-17
Fix SA1024 incorrectly treating explicit interface implementations as private

### LINKED ISSUE #1230: SA1204 Conflict between private static and explicit interface methods
Hi,
There seems to be a bug with explicit interface method classification. The following code gives a violation for SA1204 (or SA1202 if you change the order):

```csharp
private static void ExampleMethod()
{
}

IEnumerator IEnumerable.GetEnumerator()
{
	return GetEnumerator();
}
```

Kind regards,
Artur

Noryoko:
Fixes #1230


sharwell:
:+1:

PR #1250, opened 2015-08-19
Adds codefix that creates default settings file

### LINKED ISSUE #1241: Create a code fix which adds a default settings file to the project
Several diagnostics cannot be properly fixed without a settings file available. A code fix needs to be created to automate this process for the user.

### Offering the code fix

This code fix should only be offered if the `AvailableFiles` property does not contain a `stylecop.json` file. When this file is missing, it should provide this code fix for any diagnostic which requires additional settings to perform the actual fix. For obvious reasons the code fix should not provide a batch fixer.

### Code fix behavior

The code fix should perform the following actions:

1. Add a **stylecop.json** file at the root of the project using the following minimal content:

    ```json
    {
      "settings": {
        "documentationRules": {
          "companyName": "CompanyName"
        }
      }
    }
    ```

2. Include the **stylecop.json** file in the list of `AvailableFiles` for the project.

vweijsters:
- [x] Fixes #1241


StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsFileCodeFixProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+";
+
+        /// <inheritdoc/>
+        public override ImmutableArray<string> FixableDiagnosticIds { get; } =
+            ImmutableArray.Create(
+                SA1633FileMustHaveHeader.DiagnosticId,
+                SA1634FileHeaderMustShowCopyright.DiagnosticId,
+                SA1635FileHeaderMustHaveCopyrightText.DiagnosticId,
+                SA1636FileHeaderCopyrightTextMustMatch.DiagnosticId,
+                SA1637FileHeaderMustContainFileName.DiagnosticId,
```
sharwell: :question: Does this one rely on settings?

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsFileCodeFixProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+                SA1638FileHeaderFileNameDocumentationMustMatchFileName.DiagnosticId,
```
sharwell: :question: Does this one rely on settings?

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsFileCodeFixProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+                SA1639FileHeaderMustHaveSummary.DiagnosticId,
+                SA1640FileHeaderMustHaveValidCompanyText.DiagnosticId,
+                SA1641FileHeaderCompanyNameTextMustMatch.DiagnosticId,
+                SA1649FileHeaderFileNameDocumentationMustMatchTypeName.DiagnosticId);
```
sharwell: :question: Does this one rely on settings?

vweijsters:
@sharwell I added all the file header diagnostics, under the assumption that a single codefix will be made that addresses all of these. That single codefix would ofcourse need the settings.

sharwell:
@vweijsters I'm reviewing this right now. Don't worry about merging it to fix the conflicts as I've already taken care of it. :smile:

Issue #1110, opened 2015-08-02
Code fix: SA1404 (CodeAnalysisSuppressionMustHaveJustification)


sharwell:
Implement a code fix for #93.

hvanbakel:
grabbed

PR #1251, opened 2015-08-20
Added codefix for SA1404

hvanbakel:
PR for #1110

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1404UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1404UnitTests.cs
@@ -4,6 +4,7 @@
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.Diagnostics;
     using StyleCop.Analyzers.MaintainabilityRules;
     using TestHelper;
@@ -57,6 +58,18 @@ public class Foo
             DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(3, 6);
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            var fixTestCode = @"public class Foo
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage(null, null, Justification = """ + SA1404CodeFixProvider.JustificationPlaceholder + @""")]
+    public void Bar()
+    {
+
+    }
+}";
+
+            await this.VerifyCSharpDiagnosticAsync(fixTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixTestCode).ConfigureAwait(false);
         }
 
         [Fact]
@@ -75,6 +88,19 @@ public class Foo
             DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(4, 6);
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            var fixTestCode = @"using SuppressMessageAttribute = System.Diagnostics.CodeAnalysis.SuppressMessageAttribute;
+public class Foo
+{
+    [SuppressMessage(null, null, Justification = """ + SA1404CodeFixProvider.JustificationPlaceholder + @""")]
+    public void Bar()
+    {
+
+    }
+}";
+
+            await this.VerifyCSharpDiagnosticAsync(fixTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixTestCode).ConfigureAwait(false);
         }
 
         [Fact]
@@ -93,6 +119,19 @@ public class Foo
             DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(4, 6);
 
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            var fixTestCode = @"using MySuppressionAttribute = System.Diagnostics.CodeAnalysis.SuppressMessageAttribute;
+public class Foo
+{
+    [MySuppression(null, null, Justification = """ + SA1404CodeFixProvider.JustificationPlaceholder + @""")]
+    public void Bar()
+    {
+
+    }
+}";
+
+            await this.VerifyCSharpDiagnosticAsync(fixTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixTestCode).ConfigureAwait(false);
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <para>To fix a violation of this rule, add a justification to your suppression.</para>
+    /// </remarks>
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(SA1404CodeFixProvider))]
+    [Shared]
+    public class SA1404CodeFixProvider : CodeFixProvider
+    {
+        /// <summary>
+        /// The placeholder to insert as part of the code fix.
+        /// </summary>
+        public const string JustificationPlaceholder = "Code reviewed, suppressed.";
```
sharwell: :thought_balloon: I would argue that this is not actually a justification.

sharwell:
:question: Do we really want to do anything more than handle the case where no justification is provided, and insert an empty string? I lean towards keeping the code fix for *adding* `Justification = ""`, but removing the part of the code fix that inserts a non-empty string.

hvanbakel:
I think Microsoft does the same thing in FxCop, it also adds a standard justification but that one is different from this one. Maybe use that one instead?

hvanbakel:
Fixed issue #1253

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/MaintainabilityResources.resx
```diff
@@ -126,6 +126,12 @@
   <data name="SA1402CodeFix" xml:space="preserve">
     <value>Move type to new file</value>
   </data>
+  <data name="SA1404CodeFixAdd" xml:space="preserve">
+    <value>Add justification</value>
```
sharwell: :bulb: I would just use one message (this message, and name it `SA1404CodeFix`)

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+                var attribute = node as AttributeSyntax;
+                if (attribute != null)
+                {
+                    // In this case there is no justification at all
+                    var fixTitle = MaintainabilityResources.SA1404CodeFixAdd;
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            fixTitle,
+                            token => AddJustificationToAttributeAsync(context.Document, root, attribute),
+                            fixTitle), diagnostic);
```
sharwell: :bulb: This instance of `fixTitle` will need to be `nameof(SA1404CodeFixProvider) + "-Add"` so the operation is still distinct from the one below.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                var argument = node as AttributeArgumentSyntax;
+                if (argument != null)
+                {
+                    var fixTitle = MaintainabilityResources.SA1404CodeFixUpdate;
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            fixTitle,
+                            token => UpdateValueOfArgumentAsync(context.Document, root, argument),
+                            fixTitle), diagnostic);
```
sharwell: :bulb: This instance of `fixTitle` will need to be `nameof(SA1404CodeFixProvider) + "-Update"` so the operation is still distinct from the one above.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            SyntaxNode root = null;
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                if (!this.FixableDiagnosticIds.Contains(diagnostic.Id))
+                {
+                    continue;
+                }
+
+                if (root == null)
```
sharwell: :bulb: I wouldn't worry about this lazy initialization. Roslyn never calls this method without valid diagnostics to provide.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                var attribute = node as AttributeSyntax;
+                if (attribute != null)
+                {
+                    // In this case there is no justification at all
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            MaintainabilityResources.SA1404CodeFix,
+                            token => AddJustificationToAttributeAsync(context.Document, root, attribute),
+                            nameof(SA1400CodeFixProvider) + "-Add"), diagnostic);
```
sharwell: :bulb: `SA1404CodeFixProvider`

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                var argument = node as AttributeArgumentSyntax;
+                if (argument != null)
+                {
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            MaintainabilityResources.SA1404CodeFix,
+                            token => UpdateValueOfArgumentAsync(context.Document, root, argument),
+                            nameof(SA1400CodeFixProvider) + "-Update"), diagnostic);
```
sharwell: :bulb: `SA1404CodeFixProvider` (this one says SA1400)

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                if (!this.FixableDiagnosticIds.Contains(diagnostic.Id))
+                {
+                    continue;
+                }
+
+                var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
```
sharwell: :bulb: This can be lifted outside the `foreach` loop.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1404CodeFixProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+    using System.Composition;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using StyleCop.Analyzers.Helpers;
+    using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
```
sharwell: :bulb: I would prefer to not have the `using static` (personal preference).

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1404UnitTests.cs
```diff
@@ -110,6 +166,18 @@ public void Bar()
...
+            var fixTestCode = @"public class Foo
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage(null, null, Justification = """ + SA1404CodeAnalysisSuppressionMustHaveJustification.JustificationPlaceholder + @""")]
+    public void Bar()
+    {
+
+    }
+}";
+
+            // Do not verify diagnostics as the placeholder will still generate a violation
```
sharwell: :bulb: You could still verify the `fixTestCode` diagnostics, it just won't be empty like we normally have.

PR #1257, opened 2015-08-20
Improved handling of `?` in the presence of syntax errors

### LINKED ISSUE #1256: SA1018 errors on badly placed question mark
I accidentally typed a stray `?` at the end of my class and noticed this bug:

```
namespace StyleCopAnalyzers_Test
{
    public class Class1
    {
        private void Test()
        {
            var hello = "world";
        }
    }
}
?
```

The question mark at the end of the class causes this banner to appear above the code editor:

> SA1018CodeFixProvider encountered an error and has been disabled.

In my real-world project, it seemed to happen as soon as I typed the question mark. In my repro example above, it didn't occur until I clicked on the lightbulb icon to view suggested improvements. Tested in beta008.

sharwell:
I updated the code fix to handle this syntax, but I also updated the analyzer to not report any warnings for the syntax you showed.

Fixes #1256

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1018CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1018CodeFixProvider.cs
@@ -48,7 +48,12 @@
                     continue;
                 }
 
-                var nullableType = (NullableTypeSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan);
+                var nullableType = syntaxRoot.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true) as NullableTypeSyntax;
+                if (nullableType == null)
+                {
+                    continue;
+                }
+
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1018CodeFixProvider.cs
```diff
@@ -48,7 +48,12 @@ public override FixAllProvider GetFixAllProvider()
                     continue;
                 }
 
-                var nullableType = (NullableTypeSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan);
+                var nullableType = syntaxRoot.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true) as NullableTypeSyntax;
```
sharwell: :memo: Failing to include `getInnerMostNodeForTie: true` was the reason we were getting an exception here. The outer node at the location for this syntax error is an `IncompleteMemberSyntax`.

pdelvo:
:+1:

PR #1259, opened 2015-08-21
Fixes that SA1106 is not reported for semicolons after types

### LINKED ISSUE #1254: SA1106 should be reported for the semicolon following a type declaration
The following is legal C# code:

```csharp
class TypeName
{
};
```

However, the `;` at the end is not necessary. It should be treated as an empty statement for the purpose of SA1106.

pdelvo:
Fixes #1254

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1106UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1106UnitTests.cs
@@ -121,7 +121,8 @@ class TestClass
         [InlineData("struct Foo { }")]
         [InlineData("interface IFoo { }")]
         [InlineData("enum Foo { }")]
-        public async Task TestTypeAsync(string declaration)
+        [InlineData("namespace Foo { }")]
+        public async Task TestMemberAsync(string declaration)
         {
             var testCode = declaration + @"
 ;";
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1106UnitTests.cs
```diff
@@ -116,6 +116,21 @@ public void TestMethod()
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
         }
 
+        [Theory]
+        [InlineData("class Foo { }")]
+        [InlineData("struct Foo { }")]
+        [InlineData("interface IFoo { }")]
+        [InlineData("enum Foo { }")]
+        public async Task TestTypeAsync(string declaration)
```
Noryoko: Can you include a namespace declaration?

pdelvo: Im going to add it. i forgot about it because I was just looking at types

pdelvo: Fixed

Noryoko:
:+1:

PR #1260, opened 2015-08-21
Fixes that SA1100 does not support base indexer calls

### LINKED ISSUE #1247: SA1100 does not support inherited indexers
In the following code, SA1100 should be reported on the keyword `base`.

```csharp
class ClassName : List<int>
{
  public int Property { get { return base[0]; } }
}
```

pdelvo:
Fixes #1247


sharwell:
:question: Shouldn't the regression test fail before the fix?

sharwell:
:+1: The regression tests fail for me locally when run specifically on the first commit (before the fix).

PR #1263, opened 2015-08-21
Add additional test for SA1100 with indexers

sharwell:
This test makes sure the new code in #1260 doesn't report false positives.


Noryoko:
:+1:

PR #1265, opened 2015-08-22
Add missing usings

pdelvo:
Fixes the build break.


sharwell:
This was entirely my fault. I ran an organize usings command during the merge and it removed these because I forgot to do a **nuget restore** first.

PR #1266, opened 2015-08-23
Disable SA1609 (PropertyDocumentationMustHaveValue) by default

### LINKED ISSUE #1262: SA1609 - The most annoying new feature.
SA1609 was not turned on by default in the old StyleCop.  It is, without question, the most annoying new feature of the new StyleCop in that the value added is negligible vs. the amount of work needed for compliance.  All of my properties have a standard 'Gets or sets' documentation.  The 'Value' description is redundant.  Is there some justification for this rule being on be default or was it just turned on by mistake?

sharwell:
Fixes #1262

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.ruleset
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.ruleset
@@ -105,6 +105,7 @@
     <Rule Id="SA1411" Action="Error" />
     <Rule Id="SA1600" Action="None" />
     <Rule Id="SA1603" Action="Error" />
+    <Rule Id="SA1609" Action="Warning" />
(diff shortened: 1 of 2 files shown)


pdelvo:
Do we want to enable the diagnostic for this project?

Otherwise: :+1:

sharwell:
> Do we want to enable the diagnostic for this project?

Yes, we'll need to update the ruleset when the project is upgraded to Beta 9.

pdelvo:
Couldn't you just include the ruleset change in this PR? Just set it manually to warning and it should just work when we upgrade.

sharwell:
@pdelvo Updated

pdelvo:
:+1: Looks good to me

PR #1267, opened 2015-08-23
Disallow extra indentation of top-level documentation elements

### LINKED ISSUE #1238: SA1004 should report a warning for indented top-level elements
Currently SA1004 does not report a warning for the following code:

```csharp
/// <summary>
/// The summary text.
/// </summary>
///   <param name="x">The document root.</param>
///    <param name="y">The Xml header token.</param>
private void Method1(int x, int y)
{
}
```

According to the documentation for this rule, it should report a warning here. The exclusion for indented lines should only apply to *content*, such as the following:

```csharp
/// <summary>
///     The summary text.
/// </summary>
/// <param name="x">The document root.</param>
/// <param name="y">The Xml header token.</param>
private void Method1(int x, int y)
{
}
```

sharwell:
Fixes #1238


Noryoko:
:+1:

Issue #1243, opened 2015-08-18
Consider expanding header rules to support company-defined policies


sharwell:
Copyright headers are an aspect of source code which are often outside the ability of a local development team to influence. Unlike many other code style policies, switching to the StyleCop-defined file header structure may not be an option. It would be nice to consider expanding the ability of the file header rules to validate and automatically fix a wider variety of header policies. I'll start the discussion with the following header formats:

### Header 1

:memo: This form is used by my office. At the turn of each year we should see thousands of build warnings for existing code. :disappointed: 

```csharp
// Copyright (c) 2015 ACME, Inc. All rights reserved worldwide.
```

### Header 2

:memo: This form is used by corefx, Roslyn, and others.

```csharp
// Copyright (c) Microsoft. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```

### Header 3

:memo: This form was used in one of my older projects. I would probably switch it to use form 2 above if it was easy.

```
/*
 * Copyright (c) 2013 Sam Harwell, Tunnel Vision Laboratories LLC
 * All rights reserved.
 */
```

### Header 4

:memo: This form is used by two of my source-open (but not open-source) projects.

```csharp
/*
 *  Copyright (c) 2012 Sam Harwell, Tunnel Vision Laboratories LLC
 *  All rights reserved.
 *
 *  The source code of this document is proprietary work, and is not licensed for
 *  distribution. For information about licensing, contact Sam Harwell at:
 *      email@example.com
 */
```

vweijsters:
I think it would be good to offer some sort of support for all of them.
The only reasonable way of implementing that I can think of at this moment is specifying the entire expected header text in the JSON file, with a special character (like `|`) to indicate line-breaks.

sharwell:
> with a special character (like  `|`) to indicate line-breaks.

`\n` would probably be a good special character, since it's the JSON representation of an escaped newline character. :wink: The other possibility is encoding it using an array of strings, where each string in the array is one line.

vweijsters:
I definitely need more JSON knowledge :smile: 
What kind of placeholders do you mean?

sharwell:
For example, I would want the date to be a placeholder, so it knows the difference between a completely improperly formatted header and a header that's correct aside from the date. That way the code fix can update just the date.

oatkins:
[...]

PR #1269, opened 2015-08-23
Create and use a JSON object model for stylecop.json

sharwell:
* Add `copyrightText` to the JSON schema
* Add `variables` to the JSON schema

This addresses some of the concerns in #1243. I believe the remaining concerns can be addressed by selectively disabling the XML structure of the file headers, and using *just* the `copyrightText` value.


vweijsters:
:+1: Looks good to me

vweijsters:
:+1: Latest changes look good as well

DavidDeSloovere:
Might "organizationName" be a better alternative to "companyName"? I'm thinking OSS projects...

sharwell:
@DavidDeSloovere If this was a new feature, that would probably be desirable. However, this is a carry-over from StyleCop "classic" and the settings name was chosen to match the name of the XML attribute that appears in the source code for the standard StyleCop header.

If you really want to get rid of `companyName` in the header, you can use the advanced customization that is coming as part of #1271.

PR #1278, opened 2015-08-25
Implemented configuration settings for documentation rules

### LINKED ISSUE #789: SA1600-1602: support SC rules for whether documentation is required for private and internal symbols

### LINKED ISSUE #829: SA1600-1602: Diagnostic should not be reported for elements inside a private class

### LINKED ISSUE #912: New rule proposal: Non-private elements must be documented
This is a follow-up to discussion on #832.

The proposal is for a rule that is the equivalent of SA1600, SA1602 and SA1602 for non-private, non-public elements only.

This is the equivalent of [CS1591](https://msdn.microsoft.com/en-us/library/zk18c1w9.aspx) but for non-private, non-public elements.

The existing SA160x rules will continue to report violations for elements with *any* visibility, unless the same violation is reported by this new rule or CS1591. This will require modifying the SA160x rules to check whether CS1591 and this new rule is enabled.

sharwell:
* [x] Created new configuration settings
* [x] Updated SA1600 to follow new rules
* [x] Updated SA1601 to follow new rules
* [x] Updated SA1602 to follow new rules

Fixes #829 
Fixes #789 
Supersedes #832 
Closes #912

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
@@ -317,6 +317,11 @@
             return declaredSymbol?.DeclaredAccessibility ?? Accessibility.NotApplicable;
         }
 
+        internal static Accessibility GetDeclaredAccessibility(this EnumMemberDeclarationSyntax syntax)
+        {
+            return Accessibility.Public;
+        }
+
         internal static Accessibility GetDeclaredAccessibility(this DelegateDeclarationSyntax syntax, SemanticModel semanticModel, CancellationToken cancellationToken)
         {
             if (syntax == null)
@@ -447,6 +452,25 @@
             {
                 return declaredAccessibility;
             }
+
+            BaseTypeDeclarationSyntax enclosingType = (BaseTypeDeclarationSyntax)syntax.Parent;
+            Accessibility enclosingAccessibility = enclosingType.GetEffectiveAccessibility(semanticModel, cancellationToken);
+            return CombineEffectiveAccessibility(declaredAccessibility, enclosingAccessibility);
+        }
+
+        internal static Accessibility GetEffectiveAccessibility(this EnumMemberDeclarationSyntax syntax, SemanticModel semanticModel, CancellationToken cancellationToken)
+        {
+            if (syntax == null)
+            {
+                throw new ArgumentNullException(nameof(syntax));
+            }
+
+            if (semanticModel == null)
+            {
+                throw new ArgumentNullException(nameof(semanticModel));
+            }
+
+            Accessibility declaredAccessibility = Accessibility.Public;
 
             BaseTypeDeclarationSyntax enclosingType = (BaseTypeDeclarationSyntax)syntax.Parent;
             Accessibility enclosingAccessibility = enclosingType.GetEffectiveAccessibility(semanticModel, cancellationToken);
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
@@ -1,8 +1,11 @@
 ﻿namespace StyleCop.Analyzers.Helpers
 {
+    using System;
     using System.Collections.Generic;
+    using System.Threading;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
 
     /// <summary>
     /// Contains helper methods for determining an element's access level.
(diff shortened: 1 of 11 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Settings/ObjectModel/StyleCopSettings.cs
```diff
@@ -8,7 +8,7 @@ internal class StyleCopSettings
         /// <summary>
         /// This is the backing field for the <see cref="DocumentationRules"/> property.
         /// </summary>
-        [JsonProperty("documentationRules", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        [JsonProperty("documentationRules", DefaultValueHandling = DefaultValueHandling.Include)]
```
sharwell: :exclamation: This was unintentional.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
```diff
@@ -78,5 +81,508 @@ internal static string GetName(AccessLevel accessLevel)
...
+                throw new ArgumentNullException(nameof(syntax));
+            }
+
+            if (semanticModel == null)
+            {
+                throw new ArgumentNullException(nameof(semanticModel));
+            }
+
+            AccessLevel accessLevel = GetAccessLevel(syntax.Modifiers);
+            switch (accessLevel)
```
Noryoko: :bulb: Move the switch to it's own method for reuse.
```
if (accessLevel != AccessLevel.NotSpecified)
{
   return GetAccessibility(accessLevel);
}
```

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
```diff
@@ -78,5 +81,508 @@ internal static string GetName(AccessLevel accessLevel)
...
+                }
+                else
+                {
+                    return Accessibility.Public;
+                }
+            }
+
+            if (syntax.IsKind(SyntaxKind.ConstructorDeclaration))
+            {
+                return Accessibility.Private;
```
Noryoko: Static constructors are considered public by classic.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
```diff
@@ -78,5 +81,508 @@ internal static string GetName(AccessLevel accessLevel)
...
+            EventDeclarationSyntax eventDeclarationSyntax = syntax as EventDeclarationSyntax;
+            if (eventDeclarationSyntax != null)
+            {
+                if (eventDeclarationSyntax.ExplicitInterfaceSpecifier == null)
+                {
+                    return Accessibility.Private;
+                }
+                else
+                {
+                    return Accessibility.Public;
```
Noryoko: Explicit implementations of interface events are considered private by classic.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
sharwell: Since this rule isn't used for ordering, we can be more strict here. I think treating them as private is a more accurate filter on which teams would require they be documented, especially considering the documentation of static constructors almost never gets included in the published output.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
sharwell: For documentation purposes, the declared accessibility of such a method would match the effective accessibility of the declaring interface, but that behavior is not currently implemented.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
sharwell: Fixed

Noryoko:
:+1:

sharwell:
Moving this back to Beta 10 milestone since it didn't get the required additional review of the new configuration properties.

oatkins:
I look forward to trying this out now I'm back. Thanks for all the work on this.

sharwell:
@oatkins Let me know what you think of the specific properties (names and behavior) I used for this.

pdelvo:
The documentation looks good to me :+1:

PR #1283, opened 2015-08-25
Fix SA1122 should not be reported in case label

### LINKED ISSUE #1281: SA1122 falsely reported in case label.
The following code falsely reports SA1122, because `string.Empty` is not allowed there:
```
switch ("Test string")
{
    case "":
        break;
}
```

Noryoko:
- [x] Add Regression tests for #1281 
- [x] Fixes #1281

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1122UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1122UnitTests.cs
@@ -260,6 +260,31 @@ class ClassName
             await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
         }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1122UnitTests.cs
```diff
@@ -260,6 +260,31 @@ class ClassName
...
+        public async Task TestEmptyStringInCaseLabelNotReportedAsync()
+        {
+            string testCode = @"
+public class TestClass
+{
+    public void TestMethod()
+    {
+        switch (""Test string"")
+        {
+        case """":
```
sharwell: :bulb: Also add the following:

```csharp
case ("" + "a"):
```

Noryoko: Will do.

otac0n:
:+1:

PR #1286, opened 2015-08-27
Fix that TriviaHelper.WithoutLeadingBlankLines can throw exception

### LINKED ISSUE #1285: TriviaHelper.WithoutLeadingBlankLines bug
In #1252 there's now a failing test as a result of TriviaHelper.WithoutLeadingBlankLines throwing an index out of range exception on line 299 with this code:
```CSharp
public class Foo
{
    private int i = 0;

    public int Prop
    {
        /// <summary>
        /// The setter documentation
        /// </summary>
        set
        {
            i = value;
        }

        /// <summary>
        /// The getter documentation
        /// </summary>
        get
        {
            return i;
        }
    }
}
```

pdelvo:
Fixes #1285.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
@@ -313,20 +313,9 @@
 
         private static bool HasBuiltinEndLine(this SyntaxTrivia trivia)
         {
-            switch (trivia.Kind())
-            {
-            case SyntaxKind.SingleLineDocumentationCommentTrivia:
-            case SyntaxKind.PragmaWarningDirectiveTrivia:
-            case SyntaxKind.PragmaChecksumDirectiveTrivia:
-            case SyntaxKind.IfDirectiveTrivia:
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
```diff
@@ -311,6 +311,24 @@ internal static SyntaxToken WithoutLeadingBlankLines(this SyntaxToken token)
             return token.WithLeadingTrivia(newLeadingTrivia);
         }
 
+        private static bool HasBuiltinEndLine(this SyntaxTrivia trivia)
+        {
+            switch (trivia.Kind())
+            {
+            case SyntaxKind.SingleLineDocumentationCommentTrivia:
+            case SyntaxKind.PragmaWarningDirectiveTrivia:
```
sharwell: :exclamation: *All* directive trivia (i.e. the ones where `IsDirective` is true) contain a built-in end line. Some of them were left out here.

pdelvo:
I implemented the feedback

sharwell:
:+1:

PR #1287, opened 2015-08-27
Fix that SA1119 is reported in interpolations

### LINKED ISSUE #1284: SA1119 reports incorrect unnecessary parenthesis in string interpolation
This snippet triggers SA1119:
```
bool flag = false;
string data = $"{(flag ? "yep" : "nope")}";
```

But this is actually [correct syntax](http://stackoverflow.com/questions/31844058/how-to-use-the-ternary-operator-inside-interpolated-string) for using the ternary operator inside a C# 6.0 interpolated string; there's no way to write it without the parenthesis. The code fix (removing the parenthesis) causes the code to be uncompilable.

pdelvo:
Fixes #1284.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1119UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1119UnitTests.cs
@@ -1053,7 +1053,29 @@ public class Foo
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


sharwell:
:thought_balloon: This pull request always prevents parentheses from being removed from interpolated string inlets. It seems like we only want to special case ternary expressions here.

otac0n:
I agree.  In fact, the grammar treats this as a special kind of expression.

pdelvo:
Fixed

PR #1293, opened 2015-08-27
Fix SA1121 SyntaxToken.Text usage should have been SyntaxToken.ValueText

sharwell:
I found this :bug: in code coverage reports, not in a user feedback (thankfully; no one should have code like this).


pdelvo:
:+1: There is _always_ someone doing it. :laughing:

PR #1297, opened 2015-08-27
Update tests to cover the Project and Solution fix all scopes

sharwell:
Testing solution scopes revealed an additional non-deterministic ordering error in `CustomBatchFixAllProvider`, which was fixed. Marking as **enhancement** instead of :bug: because the ordering only affects testability.

@pdelvo You are a good review candidate here.


pdelvo:
:+1:

PR #1298, opened 2015-08-27
Interaction tests

sharwell:
Add unit tests covering additional edge cases in the overlap between different analyzers and the settings file.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/NoXmlFileHeaderUnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/NoXmlFileHeaderUnitTests.cs
@@ -32,6 +32,23 @@
 ";
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/NoXmlFileHeaderUnitTests.cs
```diff
@@ -32,6 +32,23 @@ public class NoXmlFileHeaderUnitTests : DiagnosticVerifier
 ";
 
         /// <summary>
+        /// Verifies that the analyzer will report the correct diagnostics (none for the default case) for a file
+        /// without a header.
```
pdelvo: I dont understand this comment. We are expecting SA1633 here?

sharwell: It was copied and pasted. I rewrote the summary for clarity here and in the original location. :+1:

pdelvo:
:+1:

PR #610, opened 2015-04-04
Implement new block-level documentation diagnostics

sharwell:
This pull request includes the implementation of **three brand-new StyleCop rules**, along with code fixes for each. A description of each of these rules can be found on the associated issue:

* Implement new diagnostic SA1653 PlaceTextInParagraphs (fixes #601)
* Implement new diagnostic SA1654 UseChildBlocksConsistently (fixes #602)
* Implement new diagnostic SA1655 UseChildBlocksConsistentlyAcrossElementsOfTheSameKind (fixes #603)

PR #1304, opened 2015-08-28
Improve concurrency behavior of the generated code cache

sharwell:
On a 28-core machine analyzing Roslyn.sln, the concurrent profiler reported an 85% reduction in the number of contentions inside of `IsGeneratedDocument`, and an 83% reduction in the amount of time threads were blocked inside this method.

This pull request also converts analysis methods to be static where possible.


pdelvo:
:+1: Nice

sharwell:
I'm marking this as blocked on #610 for a bit. I rebased and cleaned up that branch last night and I want to discuss it before having to go through that again. :smile:

Issue #1039, opened 2015-07-24
Generated code should not be analyzed


wekempf:
Generated source code is being analyzed, and since the code is generated every build there's no simple way to fix the warnings/errors produced by these analyzers. I believe generated code should not be analyzed.

I'm new to Rosalyn, but as near as I could figure out there is no general (i.e. known best practice) solution to this issue. It looks like another project has a set of extension methods they use within their analyzers to ignore generated code. [code-craker](https://github.com/code-cracker/code-cracker/blob/90184774d2506d7c38e2578126d3512c93ae6149/src/Common/CodeCracker.Common/Extensions/GeneratedCodeAnalysisExtensions.cs)

sharwell:
Hi @wekempf, and welcome to the project! :smile: 

Our intention is to not analyze generated code, and we have a number of mechanisms in place to prevent it from happening. The following will help narrow it down:

1. The ID of one of the analyzers which is reporting warnings in the file (e.g. "SA0000")
2. The name of one of your generated files which is reporting warnings
3. The header of the generated file (everything before the first `namespace` would be great)
4. The tool used to generate the file (if known)

djheap:
Here's what I'm seeing in a WPF application:

2>C:\Projects\...\obj\Debug\GeneratedInternalTypeHelper.g.cs(1,1,3,1): error SA1518: Code must not contain blank lines at end of file
2>C:\Projects\...\obj\Debug\GeneratedInternalTypeHelper.g.cs(1,1,3,1): error SA1517: Code must not contain blank lines at start of file
2>C:\Projects\...\obj\Debug\GeneratedInternalTypeHelper.g.cs(1,1,1,1): error SA1633: The file header is missing or not located at the top of the file.

The file itself just has 3 blank lines.

I believe it is generated by VS2015 for WPF code-behind or maybe resource elements (at least we aren't using any other build tools that I'm aware of).

pdelvo:
I tried to repro this: I created a new wpf application, added the analyzers and recompiled. The file you are talking about was not generated. I search for it and found one [here](https://raw.githubusercontent.com/jwarzech/NewsTweet/master/obj/Debug/GeneratedInternalTypeHelper.g.cs). When manually adding this file to the project it is correctly detected generated and excluded from analysis (there is an auto generated comment at the top). Could you maybe upload the file on your machine? Maybe VS15's version of this file looks different. Im not sure what I have to do to make it auto generate.

[...]

PR #874, opened 2015-05-24
Restrict the set of files which is excluded by filename

sharwell:
This change removes the automatic exclusion for all filenames where no specific example of this requirement could be found. The code generators for the files that are not eligible for analysis use `<auto-generated>` (or `<autogenerated>`) in the generated code, which causes the exclusion to occur by content instead.

Fixes #848
Fixes #849
Fixes #850
Fixes #851
Fixes #854

PR #1305, opened 2015-08-28
Improve generated code file detection

0x084E:
genereated file names have been "borrowed" from http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.Workspaces/GeneratedCodeRecognition/GeneratedCodeRecognitionServiceFactory.cs


sharwell:
After some discussion, these files were intentionally removed from this check in #874. I'm closing this pull request as **wontfix**, but feel free to open a new issue requesting the addition of particular file names in the check.

Keep in mind that for most cases, we can resolve detection problems as we handled it in #1039 - by assisting non-compliant code generators in updating their output to conform to StyleCop expectations.

0x084E:
Alright, makes sense. I will adjust my generators :)

PR #1308, opened 2015-08-28
SX1401StaticReadonlyFieldsMayBeNonPrivate

0x084E:
SX1401 makes SA1401 treat static readonly fields like constants. No warning is raised if they are non private.

Also I had to add a new parameter "bool enabledByDefault"  to DiagnosticOptionsHelper.IsAnalizerSuppressed. Otherwise the new rule which is disabled by default rules was treated to be always enabled (SpecificDiagnosticOptions did not conain the rule if it was disabled, therefore ReportDiagnostic.Default was returned which was not equal to ReportDiagnostic.Suppress)
This parameter is optional so I haven't changed existing code.


sharwell:
:memo: In C# 6 there is almost no need to ever have a `static readonly` field be exposed. If you remove `readonly` and replace `=` in its declaration with `{ get; }`, you get a property that functions almost identically to the original field.

I'm marking this as **needs discussion**. With the (very) recent introduction of the ability to configure some rules in **stylecop.json**, there are two possible ways to approach this. I expect it to be a major discussion point for Beta 10.

0x084E:
Using stylecop.json sounds like an interesting way. This would make it easily possible to implement whole [StyleCop+ ](https://stylecopplus.codeplex.com/) functionality which I used before to accomplish this.

p.s.
You're right, there is no need (and wasn't before C# 6 although it wasn't a one liner, of course), but I use them to expose constant values which cannot be real const, like Microsoft does for Point.Empty e.g.
I prefer this over a static property in such cases because in my opinion this makes it more clear that it is a constant value.

NikolayIT:
I also would like to see StyleCop+ features implemented (configurable of course).

PR #1314, opened 2015-08-28
Fix SA1513 incorrectly reported on object initializers without assignment

### LINKED ISSUE #1301: SA1513 falsely reported when using object initializer without asigning the object to a variable.
The code below signals SA1513 (Closing curly bracket must be followed by blank line) near the semi-colon, but it shouldn't.
```
if (true)
{
    new Point()
    {
        X = 42,
    };
}
```

Noryoko:
- [x] Add regression tests for #1301
- [x] Fixes #1301

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
@@ -404,7 +404,7 @@ public class Foo
         public async Task TestInvalidAsync()
         {
             var testCode = @"using System;
-
+using System.Collections.Generic;
 public class Foo
 {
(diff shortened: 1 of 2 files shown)


sharwell:
I think this change does not work with the following code. SA1513 should be reported on the `}` before the second return statement.

```csharp
public void Example()
{
    new List<Action>
    {
        () =>
        {
            if (true)
            {
                return;
            }
            return;
        }
    };
}
```

sharwell:
I made a mistake in my comment before. The strange behavior appears if you have an empty statement following the `}`. Notice that the `;` is actually not part of the `if` statement here:

```csharp
public void Example()
{
    new List<Action>
    {
        () =>
        {
            if (true)
            {
                return;
            };
        }
    };
}
```

Given its obscurity, this is something we may just want to file a bug for.

PR #1316, opened 2015-08-28
Variable and parameter naming analyzers and fixes

### LINKED ISSUE #1310: New rule proposal: VariableNamesMustBeginWithLowerCaseLetter
**Category:** Naming
**Title:** Variable names must begin with lower case letter
**Description:** This is very similar to SA1306, but it applies specifically to non-`const` local variables.

This functionality is handled by SA1306 in StyleCop classic, but the commentary in #1307 suggests that we continue to implement SA1306 in a more strict manner here and report variables separately. One advantage of treating local variables separately from fields is renaming them is *always* a non-breaking change (they are not even exposed via Reflection), so the ability to "fix all" on these would be valuable and trusted provided it only touched local variables.

### LINKED ISSUE #1311: New rule proposal: ParameterNamesMustBeginWithLowerCaseLetter
**Category:** Naming
**Title:** Parameter names must begin with lower case letter
**Description:** This is very similar to SA1306, but it applies specifically to parameters of methods and delegates.

This proposal was created based on feedback in #1307. This proposal is separated from #1310 due to the fact that renaming a parameter is a source-breaking change (call sites using named arguments need to be updated).

sharwell:
This pull request adds two new rules:

* SA1312 (VariableNamesMustBeginWithLowerCaseLetter) (fixes #1310)
* SA1313 (ParameterNamesMustBeginWithLowerCaseLetter) (fixes #1311)

It also updates the documentation for SA1306 (FieldNamesMustBeginWithLowerCaseLetter) to explicitly state the deviation from StyleCop "classic".

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1312VariableNamesMustBeginWithLowerCaseLetter.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1312VariableNamesMustBeginWithLowerCaseLetter.cs
@@ -1,0 +1,94 @@
+﻿namespace StyleCop.Analyzers.NamingRules
+{
+    using System.Collections.Immutable;
+    using Helpers;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Microsoft.CodeAnalysis.Diagnostics;
+
+    /// <summary>
+    /// The name of a variable in C# does not begin with a lower-case letter.
+    /// </summary>
+    /// <remarks>
+    /// <para>A violation of this rule occurs when the name of a variable does not begin with a lower-case letter.</para>
+    ///
+    /// <para>If the variable name is intended to match the name of an item associated with Win32 or COM, and thus needs
+    /// to begin with an upper-case letter, place the variable within a special <c>NativeMethods</c> class. A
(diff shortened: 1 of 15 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1312VariableNamesMustBeginWithLowerCaseLetter.cs
```diff
@@ -0,0 +1,94 @@
...
+            LocalDeclarationStatementSyntax parentDeclaration = syntax.Parent as LocalDeclarationStatementSyntax;
+            if (parentDeclaration?.IsConst ?? false)
+            {
+                // this diagnostic does not apply to locals constants
+                return;
+            }
+
+            foreach (VariableDeclaratorSyntax variableDeclarator in syntax.Variables)
+            {
+                if (variableDeclarator == null)
```
Noryoko: :question: Is `null` possible?

Noryoko:
Other than the question. Looks good. :+1:

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1312VariableNamesMustBeginWithLowerCaseLetter.cs
sharwell: I have no idea.

PR #1324, opened 2015-08-29
Permanently disable SA1603 (DocumentationMustContainValidXml)

### LINKED ISSUE #1291: Proposal: Permanently disable rule SA1603 (DocumentationMustContainValidXml)
This rule is a duplication of compiler warning [CS1570](https://msdn.microsoft.com/en-us/library/c20zzdxx.aspx), which means we're slowing down the build but not gaining any functionality from it.

sharwell:
Fixes #1291

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1603DocumentationMustContainValidXml.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1603DocumentationMustContainValidXml.cs
@@ -1,11 +1,7 @@
 ﻿namespace StyleCop.Analyzers.DocumentationRules
 {
-    using System.Collections.Generic;
     using System.Collections.Immutable;
-    using System.Linq;
     using Microsoft.CodeAnalysis;
-    using Microsoft.CodeAnalysis.CSharp;
-    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1603DocumentationMustContainValidXml.cs
```diff
@@ -35,7 +31,7 @@ public class SA1603DocumentationMustContainValidXml : DiagnosticAnalyzer
         private const string HelpLink = "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1603.md";
 
         private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, AnalyzerCategory.DocumentationRules, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, AnalyzerCategory.NamingRules, DiagnosticSeverity.Warning, AnalyzerConstants.DisabledByDefault, Description, HelpLink, WellKnownDiagnosticTags.NotConfigurable);
```
pdelvo: This is not a naming rule

pdelvo:
I wonder if we should also update the docs for the rules

sharwell:
> I wonder if we should also update the docs for the rules

Yes, I think documenting the disabled rules would be a nice addition to #359. I'm not sure yet what the best way to do that is.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1603DocumentationMustContainValidXml.cs
sharwell: I amended the previous commit to correct this.

pdelvo:
:+1:

PR #1329, opened 2015-08-30
Fix SA1312 treating fields as variables

### LINKED ISSUE #1328: SA1312 treats constant fields as variables
Currently SA1312 is being reported for constant fields which start with an upper-case letter. SA1312 should only be reported for non-constant local variables.

sharwell:
* [x] Add regression tests for #1328 
* [x] Fixes #1328


Noryoko:
:+1:

PR #1330, opened 2015-08-30
Internal rule set

sharwell:
Use an internal rule set file instead of specifying warning behavior with the `<NoWarn>` element in project files.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Internal.ruleset
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Internal.ruleset
@@ -4,5 +4,6 @@
   <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp" RuleNamespace="Microsoft.CodeAnalysis.CSharp">
     <Rule Id="CS1573" Action="Hidden" />
     <Rule Id="CS1591" Action="Hidden" />
+    <Rule Id="CS2008" Action="None" />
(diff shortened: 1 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Internal.ruleset
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Rules for StyleCop.Analyzers internal-only projects" Description="Code analysis rules for StyleCop.Analyzers internal-only projects." ToolsVersion="14.0">
+  <Include Path="StyleCop.Analyzers.ruleset" Action="Default" />
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp" RuleNamespace="Microsoft.CodeAnalysis.CSharp">
+    <Rule Id="CS1573" Action="Hidden" />
+    <Rule Id="CS1591" Action="Hidden" />
+    <Rule Id="CS2008" Action="None" />
```
pdelvo: Whyt do we want to suppress this compiler error?

sharwell: The VSIX has no source files.

pdelvo:
:+1:

PR #1331, opened 2015-08-30
Add "Inherit documentation" code fix for missing documentation

### LINKED ISSUE #968: SA1600: Auto-create documentation for overridden/interface impl members.
When SA1600 detects no documentation for an overridden member or a member defined in an interface, it could provide a code fix to copy the documentation for the member from the base class or interface.

Particularly when implementing interfaces i end up doing a lot of copy/pasing (implementing collection interfaces arrgh :ambulance:).

These could be wrapped in `<placeholder>`.

sharwell:
Fixes #968


pdelvo:
:+1:

vweijsters:
Implements the <inheritdoc/> part of #764 
(added the comment for traceability)

Issue #1327, opened 2015-08-30
TriviaHelper does not handle SingleLineDocumentTrivia correctly when removing blank lines


hvanbakel:
Given the following code:

```CSharp
public class Foo
{
    private int i = 0;

    public int Prop
    {

        /// <summary>
        /// The setter documentation
        /// </summary>
        set
        {
            i = value;
        }
    }
}
```

TriviaHelper incorrectly asserts this accessor to not have leading blank lines because of it's documentation. I fixed this in #1252 and added a test for it. However, `WithoutLeadingBlankLines()` also doesn't work and I can't figure out how to fix that method.

hvanbakel:
After discussion, I would suggest moving this method back to SA1501 as it is specific for that. It's not as general purpose as you might guess from the title.

sharwell:
This method appears to be used in the code fixes for both SA1510 and SA1511. It does sound like it needs greatly improved documentation.

hvanbakel:
Grabbed

hvanbakel:
the previous PR was a typo :(

hvanbakel:
#1338 is still a valid PR for this issue though

PR #1252, opened 2015-08-20
Code fix for SA1213

hvanbakel:
Fixes #1105 
Fixes #495

PR #1338, opened 2015-08-31
Fix TriviaHelper handling of SingleLineDocumentTrivia 

hvanbakel:
Fix for #1327


sharwell:
Blocked on #1252.

sharwell:
:question: I don't fully understand why this change is needed. Can you add a test which would fail prior to this change, but after this change passes?

hvanbakel:
Before I couldn't use the function on trivia helper now that that's updated
I can, so I dropped the local implementation in favor of the shared one.

vweijsters:
Just did a review and I have some issues with this:
* Both `WithoutBlankLines` and `WithoutLeadingBlankLines` have the same documentation comments, so its completely unclear what they do.
* The new `WithoutBlankLines` is aimed to replace the old `WithoutLeadingBlankLines`, but it is not functionaly equivalent. The old `WithoutLeadingBlankLines` would only strip leading blank lines in front of the token, and not (as `WithoutBlankLines`) strip all empty lines from the trivia.
* The issue that you try to resolve for the `SA1212SA1213CodeFixProvider` could have been resolved by using a call to `WithoutLeadingWhitespace()` as well. (`secondAccessor.GetFirstToken().LeadingTrivia.WithoutLeadingWhitespace()`)

In my opinion this does not solve #1327. Renaming the original `WithoutLeadingBlankLines` to `WithoutLeadingBlankLinesBeforeToken` and updating the documentation would be the preferred way.

pdelvo:
Do we still want to move this PR forward?

vweijsters:
Based on my earlier comments I would say no.

hvanbakel:
I would drop it too :).

PR #1347, opened 2015-08-31
Fix SA1312 getting reported for event field declarations

### LINKED ISSUE #1340: SA1312 treats short-hand declared events as variables
The following declaration raises SA1312:
```
public event EventHandler Test;
```

A full event declaration does not.
```
public event EventHandler Test<br />
{
    add { }
    remove { }
}
```

sharwell:
Also added regression tests to related rules SA1306 and SA1313.

Fixes #1340


Noryoko:
:+1:

PR #1351, opened 2015-09-01
SA1503 fix all provider

sharwell:
Fixes a huge performance problem with fixing all instances of SA1503 in Roslyn.sln.


pdelvo:
:+1: Great!

PR #1352, opened 2015-09-01
Cache properties reported with diagnostics for OpenCloseSpacingCodeFixProvider

sharwell:
These `ImmutableDictionary<string, string>` instances are trivially calculated in advance.


Noryoko:
Nice! :+1: 

I think SA1008 could also benefit from this change.

PR #1362, opened 2015-09-02
Rewrite StyleCopTester as an asynchronous console application

sharwell:
This change makes it easier to integrate the console application with asynchronous operations in Roslyn.

### DIFF
--- a/StyleCop.Analyzers/StyleCopTester/Program.cs
+++ b/StyleCop.Analyzers/StyleCopTester/Program.cs
@@ -9,6 +9,7 @@
     using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
+    using System.Windows.Threading;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CodeActions;
     using Microsoft.CodeAnalysis.CodeFixes;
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -24,6 +25,35 @@ internal static class Program
...
+                (sender, e) =>
+                {
+                    e.Cancel = true;
+                    cts.Cancel();
+                };
+
+            SynchronizationContext previousContext = SynchronizationContext.Current;
+            try
+            {
+                var context = new DispatcherSynchronizationContext();
```
pdelvo: Do we really want to take a dependency on wpf? With monodevelop supporting roslyn in the future we probably want to be able to run everything on mono too.

sharwell: If we get to a point where the project runs on Mono, this can be replaced with our own `SynchronizationContext` implementation without breaking the rest of the changes. We could probably even keep the same names for the types and just declare them as nested types in the file.

pdelvo:
:+1:

PR #1363, opened 2015-09-02
Correct issues with trivia in code fix for SA1212 and SA1213

sharwell:
This pull request addresses a number of issues in #1252:

* The change to `TriviaHelper.HasLeadingBlankLines` was not valid. The behavior of this method is corrected and a unit test in SA1212 is updated to reflect that a blank line is no longer inserted in the code fix result.
* Failure to assign the result of a call to `WithTrailingTrivia` was resulting in code duplication in the code fix provider.
* New unit tests are added to cover behavior with line comments and block comments. The expected output of these is currently set to match the *actual* behavior. A separate issue will be opened regarding the *desired* behavior.


Noryoko:
:+1:

sharwell:
Thanks! This unblocks #1364.

Issue #1345, opened 2015-08-31
Beta 9 reports "Missing Analyzer Reference"


sharwell:
When 1.0.0-beta009 is installed in a project, Visual Studio reports a MissingAnalyzerReference diagnostic:

> Analyzer assembly '{path}\StyleCopAnalyzers\packages\StyleCop.Analyzers.1.0.0-beta009\analyzers\dotnet\cs\Newtonsoft.Json.dll' depends on 'mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e' but it was not found. Analyzers may not run correctly.

sharwell:
This issue is reported only inside the IDE. The AppVeyor build is still showing up clean with Beta 9.

otac0n:
Removing the analyzer reference to Newtonsoft (Expand References\Analyzers, Right-click Newtonsoft.Json, Remove) appears to fix the warning.

I did this, and eventually I started seeing this *Info Message* showing up in VS:

    Message	AD0001	The User Diagnostic Analyzer 'StyleCop.Analyzers.DocumentationRules.FileHeaderAnalyzers' threw an exception of type 'System.IO.FileNotFoundException' with message 'Could not load file or assembly 'Newtonsoft.Json, Version=7.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The system cannot find the file specified.'.	

    'StyleCop.Analyzers.DocumentationRules.FileHeaderAnalyzers' threw the following exception:
    'System.IO.FileNotFoundException: Could not load file or assembly 'Newtonsoft.Json, Version=7.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The system cannot find the file specified.
    File name: 'Newtonsoft.Json, Version=7.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'
       at StyleCop.Analyzers.SettingsHelper.GetStyleCopSettings(ImmutableArray`1 additionalFiles)
       at StyleCop.Analyzers.SettingsHelper.GetStyleCopSettings(SyntaxTreeAnalysisContext context)
       at StyleCop.Analyzers.DocumentationRules.FileHeaderAnalyzers.HandleSyntaxTreeAxtion(SyntaxTreeAnalysisContext context, Compilation compilation)
[...]

PR #1365, opened 2015-09-02
changed newtonsoft reference to 6.0.8

jimmymain:
based on the [post below](http://stackoverflow.com/questions/30042728/attempting-to-deserialize-an-object-using-json-net-in-windows-10-universal-app-f), I downgraded newtonsoft to version 6.0.8, which solves the reference problem for [https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1345].
it's a temporary solution perhaps, but it does solve the problem until the warnings can be more permanently sorted out.

http://stackoverflow.com/questions/30042728/attempting-to-deserialize-an-object-using-json-net-in-windows-10-universal-app-f


jimmymain:
Referencing 6.0.8 is not the full solution. the version of the newtonsoft.json.dll deployed by nuget needs to be the net45 version (502kb) in size to resolve the problem.

if the portable library is deployed
```
packages\Newtonsoft.Json.6.0.8\lib\portable-net40+sl5+wp80+win8+wpa81 (392kb)
```
the issue will remain.

the deployed version 6.0.8 library must be 
```
packages\Newtonsoft.Json.6.0.8\lib\net45 (502kb)
```

PR #1354, opened 2015-09-01
Implement a custom fix all provider for SA1101

pdelvo:

PR #1374, opened 2015-09-03
Implement custom SA1028 fix all provider and simplified code fixer

pdelvo:
Maybe I am blind but I think this simplified version of the code fixer works too. The fix all provider I implemented here is as expected a lot faster then the batch fixer. Fixing all occurences in MsBuild went down from 44.7s to 2 seconds. And with the changes in #1354 it should go down even more.


sharwell:
Nice and clean. :+1:

sharwell:
This is currently blocked on pdelvo/StyleCopAnalyzers#7, #1377, and #1378.

pdelvo:
I rebased this to master and verified that it is no longer crashing.

PR #1375, opened 2015-09-03
Tester fixes

sharwell:
* Fix TesterDiagnosticProvider not actually providing any document diagnostics (no code fixes were being applied with `/fixall`)
* Fix severe performance problems calculating document diagnostics


pdelvo:
:+1:

PR #1376, opened 2015-09-03
Implement custom SA1515 fix all provider

pdelvo:
Fixes all SA1515 violations (14525) in ~3.5s.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515CodeFixProvider.cs
@@ -47,36 +47,55 @@
             var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
 
             var singleLineComment = syntaxRoot.FindTrivia(diagnostic.Location.SourceSpan.Start);
+            var commentArray = new[] { singleLineComment };
 
-            var newSyntaxRoot = ProcessTriviaList(syntaxRoot, singleLineComment.Token.LeadingTrivia, singleLineComment)
-                                ?? ProcessTriviaList(syntaxRoot, singleLineComment.Token.TrailingTrivia, singleLineComment);
+            var leadingTrivia = FixTriviaList(singleLineComment.Token.LeadingTrivia, commentArray);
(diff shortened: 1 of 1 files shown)


sharwell:
Blocked on #1377 (which enables us to do more extensive testing of this).

pdelvo:
I rebased this to master and verified that it is no longer crashing.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515CodeFixProvider.cs
```diff
@@ -77,5 +78,79 @@ private static SyntaxNode ProcessTriviaList(SyntaxNode syntaxRoot, SyntaxTriviaL
...
+                foreach (var tokenWithTrivia in tokensWithTrivia)
+                {
+                    var token = tokenWithTrivia.Key;
+                    var newLeadingTrivia = FixTriviaList(token.LeadingTrivia, tokenWithTrivia);
+                    var newTrailingTrivia = FixTriviaList(token.TrailingTrivia, tokenWithTrivia);
+
+                    replacements.Add(token, token.WithLeadingTrivia(newLeadingTrivia).WithTrailingTrivia(newTrailingTrivia));
+                }
+
+                return syntaxRoot.ReplaceTokens(replacements.Select(x => x.Key), (oldToken, newToken) => replacements[oldToken]);
```
sharwell: :question: Why `replacements.Select(...)` instead of just `replacements.Keys`?

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515CodeFixProvider.cs
```diff
@@ -77,5 +78,79 @@ private static SyntaxNode ProcessTriviaList(SyntaxNode syntaxRoot, SyntaxTriviaL
...
+            }
+
+            private static SyntaxTriviaList FixTriviaList(SyntaxTriviaList triviaList, IEnumerable<SyntaxTrivia> commentTrivias)
```
sharwell: :question: Do we gain anything (simplicity) by taking this method and `ProcessTriviaList` and creating one method that both the fix all and single fix implementations can use?

pdelvo:
@sharwell I implemented your feedback

PR #1377, opened 2015-09-04
Properly distinguish between linked files in TesterDiagnosticProvider

sharwell:
When the same file on disk appears in multiple projects, the diagnostics could be duplicated during the code fix process, leading to exceptions. This pull request updates the diagnostic provider to track diagnostics according to the original project as well as the path to the file on disk.


pdelvo:
:+1:

PR #1378, opened 2015-09-04
Fixes and simplification of SA1028 (CodeMustNotContainTrailingWhitespace)

### LINKED ISSUE #1373: SA1028 does not appear to catch single space after final closing brace
The following code will pass SA1028, where `·` is a space.

```c-sharp
namespace Foo
{
  public class Bar
  {
    ...
  }
}·
```
Adding multiple spaces will trigger SA1025, and 

```
  }
} // comments goes here·
```

will trigger SA1028. Is this as designed?

sharwell:
* Simplify implementation of SA1028
* Avoid reporting duplicate diagnostics
* Fixes #1373


pdelvo:
:+1:

PR #1380, opened 2015-09-04
Implement custom SA1516 fix all provider

pdelvo:
This was throwing out of memory exception when run on the roslyn codebase. Finishes now in less than 3 seconds on my machine

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516CodeFixProvider.cs
@@ -4,6 +4,7 @@
     using System.Collections.Immutable;
     using System.Composition;
     using System.Linq;
+    using System.Threading;
     using System.Threading.Tasks;
     using Helpers;
     using Microsoft.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516CodeFixProvider.cs
```diff
@@ -94,5 +95,37 @@ private SyntaxNode GetRelevantNode(SyntaxNode innerNode)
...
+            {
+                var diagnostics = await fixAllContext.GetDocumentDiagnosticsAsync(document).ConfigureAwait(false);
+                var syntaxRoot = await document.GetSyntaxRootAsync().ConfigureAwait(false);
+
+                List<SyntaxNode> nodes = new List<SyntaxNode>();
+
+                foreach (var diagnostic in diagnostics)
+                {
+                    var node = syntaxRoot.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
+                    node = GetRelevantNode(node);
```
sharwell: :exclamation: This treatment of `GetRelevantNode` is different from the one above.

* If `GetRelevantNode` can return `null`, any such nodes should not be added to `nodes`
* If `GetRelevantNode` cannot return `null`, the code above (in the single-fix case) should use `.` instead of `?.` to access its members

:bulb: In addition, it seems clear that we do not need to execute `FindNode` inside of `RegisterCodeFixesAsync`. You can move this call into `GetTransformedDocumentAsync` for the single-fix case to maximize the performance of `RegisterCodeFixesAsync`. Note that if you don't do this, you're going to execute `FindNode` twice for every diagnostic during a fix all operation - the first time is used to determine the equivalence keys (in the call to `RegisterCodeFixesAsync`), and the second time is used by the `FixAllInDocumentAsync` operation.

pdelvo:
@sharwell I implemented your feedback

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516CodeFixProvider.cs
```diff
@@ -27,28 +28,32 @@ public class SA1516CodeFixProvider : CodeFixProvider
...
-                if (leadingTrivia != null)
-                {
-                    context.RegisterCodeFix(CodeAction.Create(LayoutResources.SA1516CodeFix, token => GetTransformedDocumentAsync(context, syntaxRoot, node, (SyntaxTriviaList)leadingTrivia), equivalenceKey: nameof(SA1516CodeFixProvider)), diagnostic);
-                }
+                context.RegisterCodeFix(CodeAction.Create(LayoutResources.SA1516CodeFix, token => GetTransformedDocumentAsync(context, syntaxRoot, diagnostic), equivalenceKey: nameof(SA1516CodeFixProvider)), diagnostic);
             }
         }
 
-        private static Task<Document> GetTransformedDocumentAsync(CodeFixContext context, SyntaxNode syntaxRoot, SyntaxNode node, SyntaxTriviaList leadingTrivia)
+        private static Task<Document> GetTransformedDocumentAsync(CodeFixContext context, SyntaxNode syntaxRoot, Diagnostic diagnostic)
```
sharwell: :exclamation: (Missed in the original review for this code fix) Don't pass the code fix context directly to `GetTransformedDocument`. You'll need to change the first parameter to just be the `Document` and add a `CancellationToken` as the last parameter.

pdelvo: Fixed

PR #1382, opened 2015-09-05
Add fix all for OpenCloseSpacingCodeFixProvider

Noryoko:
Fix all on Roslyn for SA1009 reduced from 1818948ms to 1941ms.


pdelvo:
This is great to have :+1: 

This is a lot of code dublication. Couldn't you just refactor out the switch statement and use it in the regular fix as well as the fix all provider?

pdelvo:
Nice :+1:

PR #1386, opened 2015-09-05
Implement a fix all provider for SA1122

sharwell:
This code fix is now able to fix 1617 violations of this rule in Roslyn.sln in 8154ms on my machine (previously ran out of memory before completing when run as a 32-bit process).


pdelvo:
:+1:

PR #1387, opened 2015-09-05
Implement a fix all provider for SA1025

sharwell:
This code fix is now able to fix 237 violations of this rule in Roslyn.sln in 7880ms on my machine (previously took 58811ms on my machine).


pdelvo:
:+1:

PR #1388, opened 2015-09-05
Update SA1016 to use OpenCloseSpacingCodeFixProvider

sharwell:
There are no violations of this rule in Roslyn.sln or CodeContracts.sln.

:warning: As noted in the diff, there is a behavior change included in this pull request.


StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1016UnitTests.cs
```diff
@@ -96,7 +96,7 @@ class ClassName2
 {
 }
 
-[/*comment*/System.Obsolete]
+[/*comment*/ System.Obsolete]
```
sharwell: :warning: **Behavior change:** This code fix now only removes whitespace starting at the token where the violation is reported up through the next non-whitespace token.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TriviaHelper.cs
```diff
@@ -153,7 +153,7 @@ internal static SyntaxTriviaList WithoutTrailingWhitespace(this SyntaxTriviaList
         internal static SyntaxTriviaList WithoutLeadingWhitespace(this SyntaxTriviaList triviaList, bool endOfLineIsWhitespace = true)
         {
             var nonWhitespaceIndex = IndexOfFirstNonWhitespaceTrivia(triviaList, endOfLineIsWhitespace);
-            return (nonWhitespaceIndex >= 0) ? SyntaxFactory.TriviaList(triviaList.Take(nonWhitespaceIndex)) : SyntaxFactory.TriviaList();
+            return (nonWhitespaceIndex >= 0) ? SyntaxFactory.TriviaList(triviaList.Skip(nonWhitespaceIndex)) : SyntaxFactory.TriviaList();
```
sharwell: :warning: **Behavior change:** I have no idea how we missed this before, but `WithoutLeadingWhitespace` was actually returning *only* the leading whitespace...

Noryoko:
The SpacingResources entry can be removed. Looks good other than that. :+1:

PR #1399, opened 2015-09-05
Implement a FixAllProvider for SA1026CodeFixProvider

Noryoko:
87 violations in Roslyn.sln fix all time reduced from 8539ms to 4736ms.


Noryoko:
Good catch.

PR #1401, opened 2015-09-06
Issue1027

robinsedlaczek:
Now the analyzer for 1027 finds tabs before and within comments as well as before documentation comments. The code fix for 1027 replaces them with whitespaces. 

Tabs within documentation comments are a little bit more complicated. Therefore, we have to look for SingleLineDocumentationCommentTrivias in the analyzer. But if we do so, there are some problems in the code fix provider: the relevantText cannot be calculated with the current implementation. The first character of the violating trivia is behind the start of the FullSpan of the violating trivia. So the start of the relevant text gets negative. Some hints how to handle this? I guess, we do not want to accept tabs in documentation comments as well, correct? 

Is it possible that this problem occurs because SingleLineDocumentationCommentTrivias can span multiple lines?

By the way, why does this syntax node is called SingleLineDocumentationCommentTrivia if it can span multiple lines?


sharwell:
> Some hints how to handle this?

Take a look at the new code fix for SA1028. It manipulates spans of text directly instead of dealing with the tree. :smile:

robinsedlaczek:
Great hint! I'll do it! Thanks a lot!

robinsedlaczek:
Please do not review the pull request yet. Implementation is not complete.

PR #1405, opened 2015-09-06
Reference stylecop.json schema file in docs

AArnott:


sharwell:
:bulb: I would prefer to see this as a separate subsection of the Getting Started section, which explains the `"$schema"` property and its advantage (mostly IntelliSense in Visual Studio).

AArnott:
I also wondered where the code fixer was so we can get the original file generated to include this as well. Can you point me to it?

sharwell:
I thought it already did include it? 

I'm AFK for a bit but will check on it soon.

sharwell:
@AArnott Yes, [it's already included](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsFileCodeFixProvider.cs#L25) in the default template.

AArnott:
Oh!

sharwell:
Superseded by #1414.

AArnott:
Great. :) Less work for me.

PR #1409, opened 2015-09-07
Fix that SA1008 is reported in interpolations

### LINKED ISSUE #1407: SA1008 reports wrong warning in string interpolation
If I have the following code:

    fileTitle = $"{fileName}{(FileContainer.HasContentChanged ? "*" : String.Empty)}";

I get a SA1008 warning at the `(` in front of the FileContainer which is wrong IMO.

pdelvo:
Fixes #1407.


sharwell:
:+1:

PR #1414, opened 2015-09-07
Disable SA1633 by default and update configuration documentation

### LINKED ISSUE #1366: SA1636, SA1641 place too heavy a burden on a developer
1. It seems that legacy StyleCop does not enforce SA1636 and SA1641 unless there are stylecop settings. StyleCopAnalyzers enforces it always (except if user explicitly suppresses these on every of their .cs files, or if user explicitly mentions **all** stylecop rules except SA1636 and SA1641 in their custom ruleset).
2. It seems that there is no way to comply with SA1636 and SA1641 unless there is `stylecop.json` (otherwise the comparison will always fail), so the user migrating from old VS versions is forced not only to normalize their file headers, but to create `stylecop.json` as well.
3. `stylecop.json` is [ignored by default VS gitignore on github](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore#L169), so the user, if they use github-supplied gitignore, is forced to modify their `.gitignore` as well (by adding an explicit `!stylecop.json` line in the end), and to maintain it during github-supplied gitignore updates. What's more, it is not entirely obvious that `stylecop.json` will be ignored, so, on first attempt, user will commit changes in their project while leaving `stylecop.json` behind, which means that the project will build successfully with VS 2015 on the said user PC, with VS 2013 on other users PCs, but not with VS 2015 on other users PCs!

As I see it, it is not user-friendly and backwards-compatible at all. Or do I miss something?

Maybe SA1636 and SA1641 should not be checked if there is no `stylecop.json` present (as in legacy StyleCop)?

sharwell:
* ~~Disable SA1633 by default~~ Suppress SA1636 and SA1641 when expected text has not been configured (fixes #1366)
* Document source control requirements for **stylecop.json**
* Document JSON schema functionality for **stylecop.json** (supersedes #1405)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderAnalyzers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderAnalyzers.cs
@@ -312,6 +312,12 @@
                 return;
             }
 
+            if (string.Equals(settings.DocumentationRules.CopyrightText, DocumentationSettings.DefaultCopyrightText, StringComparison.OrdinalIgnoreCase))
+            {
+                // The copyright text is meaningless until the company name is configured by the user.
+                return;
+            }
+
             if (string.CompareOrdinal(copyrightText.Trim(' ', '\r', '\n'), settings.DocumentationRules.CopyrightText) != 0)
             {
                 var location = fileHeader.GetElementLocation(context.Tree, copyrightElement);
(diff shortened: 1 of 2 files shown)


AArnott:
:shipit:

sharwell:
Disabling SA1633 by default did not work as expected, because it was not getting reported with the severity set to suppressed. The updated implementation simply avoids reporting a warning when the company name and copyright text are set to their default values.

oatkins:
:+1:

Issue #1364, opened 2015-09-02
Code fix for SA1212 and SA1213 does not preserve blank lines between accessors


sharwell:
#1363 introduces new unit tests for SA1212 and SA1313. In this code, the `fixedCode` for the following tests does not contain a blank line between the accessors after the code fix is run. However, in each of these cases a blank line is expected.

* `SA1212UnitTests.TestPropertyWithDocumentationAsync`
* `SA1212UnitTests.TestPropertyWithLineCommentAsync`
* `SA1212UnitTests.TestPropertyWithBlockCommentAsync`
* `SA1213UnitTests.TestAddAccessorAfterRemoveAccessorWithLineCommentAsync`
* `SA1213UnitTests.TestAddAccessorAfterRemoveAccessorWithBlockCommentAsync`

Currently blocked on #1363.

hvanbakel:
Grabbed

PR #1417, opened 2015-09-08
Update code fix for SA1212 and SA1213 to preserve blank lines

hvanbakel:
Fix for #1364

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1212UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1212UnitTests.cs
@@ -73,6 +73,65 @@ public class Foo
         }
 
         [Fact]
+        public async Task TestPropertyWithDocumentationNoBlankLineAsync()
+        {
+            var testCode = @"
+public class Foo
+{
+    private int i = 0;
+
+    public int Prop
+    {
+        /// <summary>
+        /// The setter documentation
+        /// </summary>
+        set
+        {
+            i = value;
+        }
+        /// <summary>
+        /// The getter documentation
+        /// </summary>
+        get
+        {
+            return i;
+        }
+    }
+}";
+
+            DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(11, 9);
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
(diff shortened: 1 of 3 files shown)


sharwell:
Nothing stands out to me as a problem here. @Noryoko can you do a second review of this?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1212SA1213CodeFixProvider.cs
```diff
@@ -72,6 +79,12 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
             return document.WithSyntaxRoot(syntaxRoot);
         }
 
+        private static bool HasLeadingBlankLines(AccessorDeclarationSyntax secondAccessor)
```
Noryoko: :bulb: Param type can be generalized to `SyntaxNode`. Can also be extracted as an extension method.

The same can be applied to `GetLeadingTriviaWithoutLeadingBlankLines`.

Noryoko:
Add test for the following.
```csharp
public class Foo
 {
     private int i = 0;
 
     public int Prop
     {
         /// <summary>
         /// The setter documentation
         /// </summary>
         set
         {
             i = value;
         }
         /// <summary>
         /// The getter documentation
         /// </summary>
         get
         {
             return i;
         }
     }
 }";
```

sharwell:
@Noryoko Accessors are not allowed to have XML documentation comments. Are you intentionally testing an abnormal case?

Noryoko:
I grabbed the first test in the list and removed the empty line between them.

hvanbakel:
All comments fixed except for that I didn't introduce an extension method as I'm not claiming this to be as generic as one might expect when I would extract this.

Noryoko:
:+1:

StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1212UnitTests.cs
```diff
@@ -57,6 +57,66 @@ public int Prop
...
+        public async Task TestPropertyWithDocumentationNoBlankLineAsync()
+        {
+            var testCode = @"
+public class Foo
+{
+    private int i = 0;
+
+    public int Prop
+    {
+        /// <summary>
```
sharwell: :bulb: Can you update this test case to use single line comments instead of documentation comments?

PR #1421, opened 2015-09-08
Update SA1313 to ignore matching inherited parameters

### LINKED ISSUE #1372: SA1313 (ParameterNamesMustBeginWithLowerCaseLetter) should ignore matching inherited parameters
SA1313 should be suppressed when all of the following conditions are met:

1. The *n*<sup>th</sup> parameter *X<sub>n</sub>* of method *M* does not start with a lowercase letter
2. *M* overrides a method *M'*
3. The *n*<sup>th</sup> parameter *X'<sub>n</sub>* of method *M'* has exactly the same name as *X<sub>n</sub>*

Example:

```csharp
public interface ISomeType
{
    void Method(int Param1, int Param2, int Param3);
}

public abstract class SomeType : ISomeType
{
    public abstract void Method(
        int Param1,  // OK
        int Param3,  // SA1313
        int param3); // OK
}
```

hvanbakel:
Fixed #1372


sharwell:
:+1:

Issue #1416, opened 2015-09-08
SA1600 (Elements must be documented) throws InvalidCastException


oatkins:
I get the following exception, though I can't see what code is causing this. It has me scratching my head, since I can't work out how the parent of a method declaration could be a compilation unit.

```
Analyzer 'StyleCop.Analyzers.DocumentationRules.SA1600ElementsMustBeDocumented' threw the following exception:
'System.InvalidCastException: Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.CompilationUnitSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.BaseTypeDeclarationSyntax'.
   at StyleCop.Analyzers.Helpers.AccessLevelHelper.GetEffectiveAccessibility(BaseMethodDeclarationSyntax syntax, SemanticModel semanticModel, CancellationToken cancellationToken)
   at StyleCop.Analyzers.DocumentationRules.SA1600ElementsMustBeDocumented.Analyzer.HandleMethodDeclaration(SyntaxNodeAnalysisContext context)
   at StyleCop.Analyzers.AnalyzerExtensions.<>c__DisplayClass2_0`1.<RegisterSyntaxNodeActionHonorExclusions>b__0(SyntaxNodeAnalysisContext c)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c__DisplayClass35_0`1.<ExecuteSyntaxNodeAction>b__1()
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock(DiagnosticAnalyzer analyzer, Action analyze)'.
```

sharwell:
Do you have a method defined outside of a type?

oatkins:
I have 17,000 source files, so I'm not sure!

I didn't think that was ever valid C# syntax. Delegates can be outside of a type, but I wouldn't have thought that they would come through as `BaseMethodDeclarationSyntax`.

Could the exception have been thrown on incomplete code while I was typing in the editor? I can't reproduce that though.

oatkins:
OK. I managed to reproduce this. A public method placed in the global scope is invalid, but still get analyzed, and so you still see the exception.

hvanbakel:
Grabbed

PR #1422, opened 2015-09-08
Expand SA1600 to check if parent is a type

hvanbakel:
Fix for #1416

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
@@ -1,4 +1,6 @@
-﻿namespace StyleCop.Analyzers.Test.DocumentationRules
+﻿using Microsoft.CodeAnalysis;
+
+namespace StyleCop.Analyzers.Test.DocumentationRules
 {
     using System.Collections.Generic;
     using System.Threading;
@@ -13,6 +15,26 @@
     /// </summary>
     public class SA1600UnitTests : DiagnosticVerifier
     {
+        [Fact]
+        public async Task TestRegressionMethodGlobalNamespaceAsync()
+        {
+            // This test is a regression test for https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1416
+            var testCode = @"
+public void Test()
+{
+
+}";
+
+            var expected = new DiagnosticResult
+            {
+                Id = "CS0116",
+                Severity = DiagnosticSeverity.Error,
+                Locations = new[] {new DiagnosticResultLocation("Test0.cs", 2, 13)},
+                Message = "A namespace cannot directly contain members such as fields or methods"
+            };
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
+
         [Fact]
         public async Task TestClassWithoutDocumentationAsync()
         {
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
@@ -15,23 +15,32 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
     /// </summary>
     public class SA1600UnitTests : DiagnosticVerifier
     {
-        [Fact]
-        public async Task TestRegressionMethodGlobalNamespaceAsync()
+        [Theory]
+        [InlineData("public string TestMember;", 15)]
+        [InlineData("public string TestMember { get; set; }", 15)]
+        [InlineData("public void TestMember() { }", 13)]
+        [InlineData("public string this[int a] { get { return \"a\"; } set { } }", 15)]
+        [InlineData("public event EventHandler TestMember { add { } remove { } }", 27)]
+        public async Task TestRegressionMethodGlobalNamespaceAsync(string code, int column)
         {
             // This test is a regression test for https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1416
-            var testCode = @"
-public void Test()
-{
+            var testCode = $@"
+using System;
 
-}";
+{code}";
 
-            var expected = new DiagnosticResult
+            var expected = new[]
             {
-                Id = "CS0116",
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
```diff
@@ -14,6 +16,26 @@
...
+{
+
+}";
+
+            var expected = new DiagnosticResult
+            {
+                Id = "CS0116",
+                Severity = DiagnosticSeverity.Error,
+                Locations = new[] {new DiagnosticResultLocation("Test0.cs", 2, 13)},
+                Message = "A namespace cannot directly contain members such as fields or methods"
```
oatkins: Nice to see a test for this.

Presumably we hard-code our tests to run with English resources?!?

Noryoko:
Please include methods.
```
HandleConstructorDeclaration
HandleDestructorDeclaration
HandlePropertyDeclaration
HandleIndexerDeclaration
HandleFieldDeclaration
HandleDelegateDeclaration
HandleEventDeclaration
HandleEventFieldDeclaration
```

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
```diff
@@ -90,7 +90,12 @@ public void HandleTypeDeclaration(SyntaxNodeAnalysisContext context)
                     return;
                 }
 
-                BaseTypeDeclarationSyntax declaration = (BaseTypeDeclarationSyntax)context.Node;
+                BaseTypeDeclarationSyntax declaration = context.Node as BaseTypeDeclarationSyntax;
```
oatkins: I had a quick look at the code and thought that `AccessLevelHelper` might be a better place to make this change. It has several similar methods, with only two cases where the try-cast pattern isn't used (and therefore where we might get this error). It's the two overloads of `GetEffectiveAccessibility()` for property and method definitions. We should probably fix properties too.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
```diff
@@ -116,6 +121,10 @@ public void HandleMethodDeclaration(SyntaxNodeAnalysisContext context)
                 }
 
                 MethodDeclarationSyntax declaration = (MethodDeclarationSyntax)context.Node;
+                if (!(declaration.Parent is BaseTypeDeclarationSyntax))
```
oatkins: I had a quick look at the code and thought that `AccessLevelHelper` might be a better place to make this change. It has several similar methods, with only two cases where the try-cast pattern isn't used (and therefore where we might get this error). It's the two overloads of `GetEffectiveAccessibility()` for property and method definitions. We should probably fix properties too.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
oatkins: I don't think this is the cause of the exception. This cast *should* be safe AFAIK.

sharwell: Correct, the `context.Node` instance should keep the original cast.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
sharwell: Not sure, but we have other tests where an English message is used. Have you run into problems with that?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
hvanbakel: If that method is not allowed to throw an exception, what would the result be? I think this is a legitimate reason for throwing an exception which is why I solved the problem in the calling code.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
hvanbakel: That's correct, this particular case did not throw an exception here.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
sharwell: It would just return the `declaredAccessibility` value if the parent is not a type.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
oatkins: No. We should probably check though. Anyone tried running the tests on a non-English installation?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
oatkins: @sharwell That sounds reasonable. It doesn't really matter *what* we return because the code is broken anyway. We just need to not throw.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
```diff
@@ -91,6 +91,11 @@ public void HandleTypeDeclaration(SyntaxNodeAnalysisContext context)
                 }
 
                 BaseTypeDeclarationSyntax declaration = (BaseTypeDeclarationSyntax)context.Node;
+                if (declaration == null)
```
oatkins: I don't think we do the null check elsewhere. It's probably redundant.

oatkins:
Looks good to me.

Issue #1429, opened 2015-09-09
UsingCodeFixProvider moves file headers


sharwell:
Two new unit tests are included in #1427 that reveal a bug in the handling of file headers by `UsingCodeFixProvider`:

* `UsingCodeFixProviderUnitTests.VerifyUsingReorderingWithFileHeaderAsync`
* `UsingCodeFixProviderUnitTests.VerifyUsingReorderingWithoutMovingWithFileHeaderAsync`

Currently the code fix provider moves file headers when it reorders using directives. These headers, which are identified by the blank line separating them from the using directives, should not be moved when the using directives are reordered.

PR #1431, opened 2015-09-09
Using code fix now leaves file header in place.

hvanbakel:
Fix for #1429


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/UsingCodeFixProvider.cs
```diff
@@ -503,6 +524,52 @@ private static List<UsingDirectiveSyntax> GenerateUsings(List<UsingDirectiveSynt
...
+                            {
+                                onBlankLine = true;
+                            }
+
+                            break;
+
+                        default:
+                            done = true;
+                            break;
+                    }
```
vweijsters: Should a MultiLineCommentTrivia file header not be handled as well?
Example:
```C#
/*
 * Copyright by FooCorp Inc.
 */

using System;
```

sharwell: It should.

PR #1436, opened 2015-09-09
Update to StyleCop.Analyzers 1.0.0-beta010

sharwell:
:memo: I'm sending a separate pull request to fix the file headers.

PR #1438, opened 2015-09-09
Enable and fix all violations of SA1633 and SA1636

sharwell:


Noryoko:
:+1:  Good catch on the Code Cracker reference in #1436.

PR #1440, opened 2015-09-10
Fix handling of line endings in FileHeaderCodeFixProvider + SA1635 code fix

### LINKED ISSUE #1437: Fix header code fix should normalize line endings
Currently, if the `copyrightText` in **stylecop.json** contains `\n` as a line ending, the file header code fix will insert a LF line ending in the generated header, even if the rest of the file uses CRLF line endings. This code fix should automatically replace `\n` with the expected line ending for the current file.

As a workaround, it is possible to use `\r\n` when configuring the copyright text in **stylecop.json**.

sharwell:
* Fixes #1437 
* Enables FileHeaderCodeFixProvider for SA1635 (FileHeaderMustHaveCopyrightText)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/NoXmlFileHeaderUnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/NoXmlFileHeaderUnitTests.cs
@@ -7,7 +7,7 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
     using System.Threading;
     using System.Threading.Tasks;
     using Analyzers.DocumentationRules;
-    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.Diagnostics;
     using TestHelper;
     using Xunit;
@@ -15,7 +15,7 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
     /// <summary>
     /// Unit tests for file header that do not follow the XML syntax.
     /// </summary>
-    public class NoXmlFileHeaderUnitTests : DiagnosticVerifier
+    public class NoXmlFileHeaderUnitTests : CodeFixVerifier
     {
         private const string SettingsFileName = "stylecop.json";
         private const string TestSettings = @"
@@ -46,9 +46,18 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
 {
 }
 ";
+            var fixedCode = @"// Copyright (c) FooCorp. All rights reserved.
+// Licensed under the ??? license. See LICENSE file in the project root for full license information.
+
+namespace Foo
+{
+}
+";
 
             var expectedDiagnostic = this.CSharpDiagnostic(FileHeaderAnalyzers.SA1633DescriptorMissing).WithLocation(1, 1);
             await this.VerifyCSharpDiagnosticAsync(testCode, expectedDiagnostic, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
         }
 
         /// <summary>
@@ -87,13 +96,13 @@ namespace Bar
         }
 
         /// <summary>
-        /// Verifies that a valid file header built using multi line comments will not produce a diagnostic message.
+        /// Verifies that a valid file header built using multi-line comments will not produce a diagnostic message.
         /// </summary>
         /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
         [Fact]
-        public async Task TestValidFileHeaderWithMultiLineCommentsAsync()
+        public async Task TestValidFileHeaderWithMultiLineComments1Async()
         {
-            var testCodeFormat1 = @"/* Copyright (c) FooCorp. All rights reserved.
+            var testCode = @"/* Copyright (c) FooCorp. All rights reserved.
  * Licensed under the ??? license. See LICENSE file in the project root for full license information.
  */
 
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/NoXmlFileHeaderUnitTests.cs
```diff
@@ -46,9 +46,18 @@ public class NoXmlFileHeaderUnitTests : DiagnosticVerifier
...
+
+namespace Foo
+{
+}
+";
 
             var expectedDiagnostic = this.CSharpDiagnostic(FileHeaderAnalyzers.SA1633DescriptorMissing).WithLocation(1, 1);
             await this.VerifyCSharpDiagnosticAsync(testCode, expectedDiagnostic, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
```
vweijsters: Passing a `CancellationToken.None` to `VerifyCSharpFixAsync` is unnecessary and it does not help readability.

sharwell: I pass a cancellation token in all cases where one can be passed. I would greatly prefer to pass one as described in xunit/xunit#345 but that option is not currently available.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -104,36 +106,39 @@ private static SyntaxNode ReplaceHeader(Document document, SyntaxNode root, Styl
...
-            return root.WithLeadingTrivia(CreateNewHeader(document.Name, settings).Add(SyntaxFactory.CarriageReturnLineFeed).Add(SyntaxFactory.CarriageReturnLineFeed).AddRange(trivia));
+            string newLineText = document.Project.Solution.Workspace.Options.GetOption(FormattingOptions.NewLine, LanguageNames.CSharp);
+            return root.WithLeadingTrivia(CreateNewHeader(document.Name, settings, newLineText).Add(SyntaxFactory.CarriageReturnLineFeed).Add(SyntaxFactory.CarriageReturnLineFeed).AddRange(trivia));
         }
 
-        private static SyntaxNode AddHeader(SyntaxNode root, string name, StyleCopSettings settings)
+        private static SyntaxNode AddHeader(Document document, SyntaxNode root, string name, StyleCopSettings settings)
         {
-            var newTrivia = CreateNewHeader(name, settings).Add(SyntaxFactory.CarriageReturnLineFeed).Add(SyntaxFactory.CarriageReturnLineFeed);
+            string newLineText = document.Project.Solution.Workspace.Options.GetOption(FormattingOptions.NewLine, LanguageNames.CSharp);
```
vweijsters: There is currently no test case that verifies that this option is properly used. I think its good to have one to prevent possible regression issues.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -104,36 +106,39 @@ private static SyntaxNode ReplaceHeader(Document document, SyntaxNode root, Styl
...
+                $"// <copyright file=\"{filename}\" company=\"{settings.DocumentationRules.CompanyName}\">" + newLineText
+                + copyrightText + newLineText
+                + "// </copyright>";
         }
 
-        private static string GetCopyrightText(string copyrightText)
+        private static string GetCopyrightText(string copyrightText, string newLineText)
         {
-            return string.Join("\n// ", copyrightText.Split('\n'));
+            return string.Join(newLineText + "// ", copyrightText.Split('\n'));
```
vweijsters: This will give issues when the newLineText is set to `\n` and the copyright text in the settings file is using `\r\n`.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/NoXmlFileHeaderUnitTests.cs
vweijsters: Ok. 
:question: Given that policy, would it be a good idea to modify syntax of `VerifyCSharpFixAsync` to require a cancellation token as its 3rd parameter? That would improve readability and it would also force consistency over the codebase.

sharwell: I'd rather let this be motivation for DotNetAnalyzers/AsyncUsageAnalyzers#36 and DotNetAnalyzers/AsyncUsageAnalyzers#37, and correct any violations when those are available.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
sharwell: It's hard because last I checked, even the Roslyn formatter does a poor job of handling this setting.

Also, until this is a per-file option, we're going to get it wrong in some cases.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
sharwell: You are correct. My expectation is users will update their **stylecop.json** files to only use `\n` for new lines starting with the next release.

vweijsters: An easy fix could be:
`copyrightText.Replace("\r\n", "\n").Split('\n')`

It could also be done in the SettingsHelper to reduce the performance impact

sharwell: Will do. This is a code fix so operations like this are not as big a deal.

PR #1441, opened 2015-09-10
Implement a code fix provider for SA1626 (SingleLineCommentsMustNotUseDocumentationStyleSlashes)

### LINKED ISSUE #655: Code fix: SA1626 (SingleLineCommentsMustNotUseDocumentationStyleSlashes)
Implement a code fix for #146.

sharwell:
Fixes #655

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1626CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1626CodeFixProvider.cs
@@ -1,0 +1,100 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.DocumentationRules
+{
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Composition;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Text;
+    using StyleCop.Analyzers.Helpers;
+
+    /// <summary>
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1626CodeFixProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        /// <inheritdoc/>
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            foreach (Diagnostic diagnostic in context.Diagnostics)
+            {
+                if (diagnostic.Id != SA1626SingleLineCommentsMustNotUseDocumentationStyleSlashes.DiagnosticId)
+                {
+                    continue;
+                }
```
vweijsters: This is different from a lot of other code fixes. Most use:
```c#
    foreach (Diagnostic diagnostic in context.Diagnostics.Where(d => FixableDiagnosticIds.Contains(d.Id)))
```

vweijsters:
:+1: Looks good beside a minor remark

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1626CodeFixProvider.cs
sharwell: That's because you implemented most of them. :laughing: 

I prefer this pattern because it avoids allocating an enumerator (`context.Diagnostics` is an `ImmutableArray<T>`, which has a custom struct enumerator). Note that `FixableDiagnosticIds.Contains` does not allocate because `ImmutableArray<T>` provides its own implementation of `Contains`.

vweijsters: You may have a point there :laughing: 
:question: Is it worth it to apply the same pattern to all code fixes? (as I mentioned, I prefer consistency :wink:)

sharwell: Actually, we should probably remove the check altogether based on dotnet/roslyn#2643 and dotnet/roslyn#2886.

Issue #1355, opened 2015-09-01
A few rules have encoded HTML in their description.


otac0n:
Files: `StyleCop.Analyzers\**\SA*.cs`
Text: `Description = ".*\&`

Hits:

    StyleCop.Analyzers\DocumentationRules\SA1620GenericTypeParameterDocumentationMustMatchTypeParameters.cs
        private const string Description = "The &lt;typeparam&gt; tags within the Xml header documentation for a generic C# element do not match the generic type parameters on the element.";

    StyleCop.Analyzers\DocumentationRules\SA1622GenericTypeParameterDocumentationMustHaveText.cs
        private const string Description = "A &lt;typeparam&gt; tag within the Xml header documentation for a generic C# element is empty.";

    StyleCop.Analyzers\DocumentationRules\SA1623PropertySummaryDocumentationMustMatchAccessors.cs
        private const string Description = "The documentation text within a C# property’s &lt;summary&gt; tag does not match the accessors within the property.";

This causes the encoded text to show up in the ruleset editor.

This shouldn't be to hard to add a test around.  In my opinion, it would be best to write a test that checks the `documentation\` file for each rule, asserting that the `.md` file contains the text in rule's description.

PR #1447, opened 2015-09-10
Fix encoded HTML tags in description messages

ChrSteinert:
This fixes all encoded HTML tags in description messages for rule classes (See #1355).


sharwell:
:+1:

PR #1449, opened 2015-09-10
Fix that SA1504 code fix removes statements

### LINKED ISSUE #1448: SA1504: Additional setter lines are deleted unintentionally
If I have a something like:

```csharp
public Uri CurrentContentUri
{
    get { return _currentContentUri; }
    private set
    {
        _currentContentUri = value;
        OnPropertyChanged();
    }
}
```

and want to follow the suggestion for SA1504, it will delete the OnPropertyChanged() line. It seems like it deletes all additional setter lines on this rule

image

pdelvo:
Fixes #1448.


vweijsters:
:+1: Looks good to me

PR #1452, opened 2015-09-10
Cleanup for code fixes

vweijsters:
Refactored all code fixes for consistency, based on discussions in #1441

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
@@ -496,7 +496,7 @@ class TypeName
 {
     void MethodName()
     {
-        do { Bar(); } while (false);
+        do { MethodName(); } while (false);
 ";
 
             var fixedTestCode = @"using System;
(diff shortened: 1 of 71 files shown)


sharwell:
I find myself wishing this was multiple commits:

1. The first commit could simply wrap the calls to `RegisterCodeFix`
2. The second commit could remove the `.Where` and `if (...)` conditions that check to see if a diagnostic ID is fixable
3. The third commit could remove `FixableDiagnostics` as a separate field (favoring the `{ get; } = ...` syntax instead)

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -507,12 +507,12 @@ void MethodName()
...
         do
         {
-            Bar();
+            MethodName();
         }
         while (false);
 ";
 
-            await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedTestCode, allowUnsupportedDiagnostics: true).ConfigureAwait(false);
```
sharwell: :question: Why did this one need to be a special case? It seems like a bunch of special behavior was added to the testing framework just to support this single test.

vweijsters: This test contains invalid C# syntax. The closing brace for the method is not present, resulting in a compiler being produced for it in the verify case.
As I wanted to keep the `Assert` in the `CodeFixVerifier` signalling that a test case is producing unexpected errors, I had to go to quite a bit of effort to make this test case work.

vweijsters:
I'll address this in separate pull requests. Keeping this one open until I'm done with those

vweijsters:
Closing this as all content has been addressed in other pull requests.

PR #1460, opened 2015-09-12
Fix handling of '*' followed by '*' or ')'

### LINKED ISSUE #1457: SA1023 needs finetuning: e.g. char** and (char**) are considered invalid
The rule SA1023, Dereference symbol '*' must be followed by a space, tries to correct the following lines:

```
char**
(char*)
```

to

```
char* *
(char* )
```

In both cases, classic StyleCop would accept these uses of the `*` symbol as valid.

sharwell:
Fixes #1457


vweijsters:
:+1: Looks good to me

PR #1461, opened 2015-09-12
Fix behavior of SA1506 code fix when a blank line follows a line comment

### LINKED ISSUE #1456: SA1506 code fix throws ArgumentOutOfRangeException
Exception was thrown in SA1506CodeFixProvider line 71 (with index = -1). Seen when trying to fix https://github.com/icsharpcode/RefactoringEssentials/blob/bb25d62f8610d863a3bc639dc1924585bf4373c9/RefactoringEssentials/Util/SymbolExtensions.cs#L1383

sharwell:
Fixes #1456

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1506CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1506CodeFixProvider.cs
@@ -52,32 +52,50 @@ namespace StyleCop.Analyzers.LayoutRules
             var triviaList = token.LeadingTrivia;
 
             var index = triviaList.IndexOf(SyntaxKind.SingleLineDocumentationCommentTrivia);
-            for (; index < triviaList.Count - 1; index++)
+
+            int currentLineStart = index + 1;
+            bool onBlankLine = true;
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1506CodeFixProvider.cs
```diff
@@ -52,32 +52,50 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
-                }
-            }
+                case SyntaxKind.EndOfLineTrivia:
+                    if (onBlankLine)
+                    {
+                        for (int i = 0; i < currentIndex - currentLineStart + 1; i++)
+                        {
+                            triviaList = triviaList.RemoveAt(currentLineStart);
+                            currentIndex = currentLineStart - 1;
+                        }
```
vweijsters: :question: What is the intention for the `for` loop, as it will only perform a single iteration.

sharwell: The assignment to `currentIndex` should have appeared outside the loop. I replaced the loop with a call to a new extension method `RemoveRange` to help clarify the code.

vweijsters:
:+1: Looks good to me

PR #1463, opened 2015-09-12
Add regression test for element preceded by attribute

### LINKED ISSUE #737: SA1516 should require a line before attributes
If attribute is added to type or member, there should be blank line enforced above it, unless 
* there is another attribute or xml doc comment in the line above
* element decorated with attribute is first element of its parent (e.g. first member in the type, or first type in a namespace)
* attribute is added inline - e.g. attributes added for parameters should not trigger this rule

Rationale - currently SC enforces blank line before element if there are no attributes and xml doc comments at all. It also enforces blank line above xml doc comment. Attributes feel like missing piece in the whole puzzle

sharwell:
Closes #737


vweijsters:
:+1: Looks good to me

PR #1466, opened 2015-09-13
SA1027 fix all provider and bug fixes

### LINKED ISSUE #1246: SA1027 (TabsMustNotBeUsed) code fix ignores trivia
e.g. assume all lines indented by tabs

		/// <summary>
		/// foo bar
		/// </summary>
		public void FooBar()

after code fix we will receive:

		/// <summary>
		/// foo bar
		/// </summary>
        public void FooBar()

code indented by spaces but documentation still indented by tabs

### LINKED ISSUE #1465: Implement a fix all provider for SA1027 (TabsMustNotBeUsed)
This diagnostic is still relying on the batch fixer. A custom fix all provider should be implemented for maximum efficiency in large solutions.

sharwell:
* Improved efficiency of SA1027 code fix by using text changes instead of trivia changes
* Fix analysis of comments; uses tests from @robinsedlaczek :+1: (Fixes #1246)
* Implement a custom fix all provider (Fixes #1465)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027CodeFixProvider.cs
@@ -59,7 +59,7 @@ namespace StyleCop.Analyzers.SpacingRules
 
             TextLine startLine = sourceText.Lines.GetLineFromPosition(span.Start);
             string text = sourceText.ToString(TextSpan.FromBounds(startLine.Start, span.End));
-            StringBuilder replacement = new StringBuilder();
+            StringBuilder replacement = new StringBuilder(indentationOptions.TabSize * span.Length);
             int column = 0;
             for (int i = 0; i < text.Length; i++)
             {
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027CodeFixProvider.cs
```diff
@@ -49,53 +49,84 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
-                SourceText sourceText = await document.GetTextAsync(cancellationToken).ConfigureAwait(false);
-                relevantText = sourceText.ToString(new TextSpan(violatingTrivia.FullSpan.Start - firstTriviaIndex, firstTriviaIndex + violatingTrivia.FullSpan.Length));
-            }
+        private static TextChange FixDiagnostic(IndentationOptions indentationOptions, SourceText sourceText, Diagnostic diagnostic)
+        {
+            TextSpan span = diagnostic.Location.SourceSpan;
 
+            TextLine startLine = sourceText.Lines.GetLineFromPosition(span.Start);
+            string text = sourceText.ToString(TextSpan.FromBounds(startLine.Start, span.End));
+            StringBuilder replacement = new StringBuilder();
```
Maxwe11: Why not to pass `text.Length` as capacity? I know that resulting string will be larger than `text.Length` but anyway it will reduce amount of SB chunks.

pdelvo: I dont think using a capacity that is too small in all possible cases (if the tab size is not set to 1) is a good idea. `indentationOptions.TabSize * text.Length` might be a good approximation.

This is only part of a code fix so not providing a capacity is not a big deal.

sharwell: I used `span.Length` instead of `text.Length`.

Maxwe11:
:+1:

PR #1467, opened 2015-09-13
Update SA1017 to use OpenCloseSpacingCodeFixProvider

sharwell:
This resulted in a slight change in the way certain trivia is treated, as shown in the tests. The new behavior should be more consistent with the expectations of users.


Noryoko:
:+1:

PR #1469, opened 2015-09-13
Remove unnecessary missing token checks

### LINKED ISSUE #539: FindToken returns default on failure, not missing
If `SyntaxNode.FindToken` cannot find the requested token, it returns `default(SyntaxToken)`. This result is different from a missing token, so code which needs a concrete token needs to use our `IsMissingOrDefault()` extension method rather than just check the `IsMissing` property before proceeding.

sharwell:
In each of these cases, the diagnostic analyzers already ensure that the token at the reported location is not missing.

Closes #539


Noryoko:
:+1:

PR #1470, opened 2015-09-13
SA1400 (AccessModifierMustBeDeclared) now ignores operators

### LINKED ISSUE #591: SA1400 (AccessModifierMustBeDeclared) should not analyze operators
The C# language requires operators include an access modifier (`public`). This requirement makes SA1400 unnecessary for these declarations and the analysis for them should be removed.

sharwell:
Fixes #591


Maxwe11:
LGTM

PR #1471, opened 2015-09-13
Add additional tests for SA1613 with <inheritdoc/>

### LINKED ISSUE #631: Additional test case for SA1613 (ElementParameterDocumentationMustDeclareParameterName)
We should include a test case to ensure SA1613 violations are still reported if the documentation includes an `<inheritdoc/>` element, since many other documentation warnings are suppressed in this case.

sharwell:
Fixes #631


Noryoko:
:+1:

PR #1491, opened 2015-09-17
Improve IsFirstInLine and IsLastInLine performance

pdelvo:
IsFirstInLine was showing up as a hot path when measuring performance. 2.3% of the time spent analyzing roslyn was spent in this method. With this change the time goes down to 0.3%.

GetNextToken and GetPreviousToken are very expensive and we should try to avoid using them if possible.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/HelperTests/TokenHelperTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/HelperTests/TokenHelperTests.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.Test.HelperTests
+{
+    using System.Linq;
+    using Analyzers.Helpers;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Xunit;
+
+    public class TokenHelperTests
+    {
+        [Fact]
+        public void TestIsFirstInLineSpecialCase()
+        {
+            string testCode = @"class MyClass {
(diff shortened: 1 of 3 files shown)


sharwell:
:question: What does this implementation of `IsFirstInLine` return for the `struct` token in the following code?

```
class MyClass {
} /// <summary>
/// Text
/// </summary>
struct MyStruct { }
```

pdelvo:
I fixed the bug you mentioned and implemented tests for that. Strange that this was not catched by one of our other tests.

StyleCop.Analyzers/StyleCop.Analyzers.Test/HelperTests/TokenHelperTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            string testCode = @"class MyClass {
+} /// <summary>
+/// Text
+/// </summary>
+struct MyStruct { }";
+            var syntaxTree = CSharpSyntaxTree.ParseText(testCode);
+            var root = syntaxTree.GetRoot();
+
+            Assert.True(root.FindToken(64).IsFirstInLine());
```
sharwell: :bulb:

```csharp
Assert.True(root.DescendantTokens().First(i => i.IsKind(SyntaxKind.StructKeyword)).IsFirstInLine());
```

StyleCop.Analyzers/StyleCop.Analyzers.Test/HelperTests/TokenHelperTests.cs
```diff
@@ -0,0 +1,45 @@
...
+} /// <summary>
+/// Text
+/// </summary>
+struct /**
+Foo
+**/ MyStruct { }";
+            var syntaxTree = CSharpSyntaxTree.ParseText(testCode);
+            var root = syntaxTree.GetRoot();
+
+            Assert.True(root.FindToken(64).IsLastInLine());
```
sharwell: :bulb:

```csharp
Assert.True(root.DescendantTokens().Single(i => i.IsKind(SyntaxKind.StructKeyword)).IsLastInLine());
```

PR #1493, opened 2015-09-18
Reduce trivia list allocations

pdelvo:
After running a memory profiling session I saw that we are allocating a lot of SyntaxTriviaList structs on the heap because of casting them to IReadOnlyList. 
snip_20150918031719

In this PR I reduce the number of allocations by not casting to an Interface and using the struct directly. I also made DualTriviaListHelper a struct and I made sure we dont cast it to IReadOnlyList. There is still a lot of room for improvement.

I also changed a few spots where we used Linq on structs.


pdelvo:
This is the result after this change:
snip_20150918114446

PR #1503, opened 2015-09-18
Convert public types to internal

### LINKED ISSUE #1502: Change public types in StyleCop.Analyzers.dll to be internal
Currently the assembly is unintentionally advertising a public API. All types in the assembly should be changed to internal until the following concerns are addressed:

1. A breaking changes policy is fully-defined.
2. A distribution strategy which cannot result in conflicting analyzer assemblies is clear.

If these items are resolved in the future and it's clear that we want to share code with other assemblies, we can propose exposing particular types and members for that purpose.

A unit test should be added to verify that we are not exposing types from the assembly.

> :warning: Code which currently depends on public types to be exposed will break as a result of this change. Any such assemblies are expected to instead use the same approach we used for `BatchFixAllProvider`:
>
> 1. Copy the required code from this project into that assembly
> 2. Update the license accompanying the library to properly reference this project (see #1485 for an example where we referenced Roslyn)

sharwell:
Fixes #1502


Maxwe11:
:+1:

Issue #1432, opened 2015-09-09
SA1633 should allow multi-line comment file headers


Noryoko:
The following examples produce no errors in classic, but are flagged by `FileHeaderAnalyzers`.

```csharp
/* <copyright file="Test.cs" company="Test">
 Copyright (c) Test. All rights reserved.
 </copyright> */
```

```csharp
/*
 <copyright file="Test.cs" company="Test">
 Copyright (c) Test. All rights reserved.
 </copyright>
*/
```

vweijsters:
Grabbing this

Issue #1356, opened 2015-09-01
File headers: leading spaces should be ignored in copyright messages on lines other than the first


oatkins:
If the correct header for a file is

    // <copyright file="fish.cs" company="Face">
    //   Copyright (c) Face.
    //   All rights reserved.
    // </copyright>

Then the settings file must contain:

    "copyrightText": "Copyright (c) Face.\n   All rights reserved."

Note that, after the "\n" there must be exactly three spaces to match the indent of the second line of the message.

This was also true in legacy StyleCop, but it would be nicer if you could just have: 

    "copyrightText": "Copyright (c) Face.\nAll rights reserved."

sharwell:
:memo: The code fix will insert exactly one space at the beginning of each line. If you plan to use the code fix and want more than one space, you'll still need to code them in.

oatkins:
OK, though I don't think "exactly one space" will be helpful for most preexisting projects. For example, our own tests use three, and [this example](http://blogs.msdn.com/b/sourceanalysis/archive/2008/05/25/configuring-file-header-company-and-copyright.aspx) from the old StyleCop blog uses a whopping five!

sharwell:
Exactly one space is the most flexible policy. It's easy to add spaces, but difficult to remove them. Zero spaces might even make sense but that would lead to violations of the "include a space in comments" rule.

otac0n:
I would actually support treating all spans of whitespace `[\s\r\n]+` as a single space for validation purposes.

@sharwell Does it make sense for the code fix to introduce spacing for the first line, but none of the others?  Also, what if a user wants to switch away from XML comments?  Ideally this would not require changes to the copyright text configuration.

I would therefore recommend that the code fix introduce context-sensitive indentation for all lines in the copyright text.

    // <copyright ...>
    //   Line 1
    //   Line 2
    // </copyright>

And:

    // Line 1
    // Line 2

sharwell:
@otac0n I discourage the indentation of content within XML tags. It cramps the right margin for and offers little to no benefit when syntax highlighting colors are updated to a more muted color for the XML tags.

otac0n:
@sharwell All the more reason for SyleCop's Code Fix to apply the correct indentation automatically.

vweijsters:
If I understand the @oatkins correctly, he is asking for the SA1636 analyzer to ignore any leading / trailing space on a line when checking it against the copyright text in the settings file. 
[...]

PR #1505, opened 2015-09-19
File header code fix should only replace the copyright element

### LINKED ISSUE #1434: File header code fix should only replace the copyright element
An XML file header may contain additional elements, including the following:

1. Horizontal bars (the series of `-----------------` characters)
2. A `<summary>` element
3. Other XML elements
4. Other text outside of XML elements

## Expected behavior

When the file header code fix corrects the `<copyright>` element, each of these other elements should be preserved.

## Actual behavior

When the file header code fix corrects the `<copyright>` element, the entire header is replaced (each of the items is lost).

peteri:
Fixes for SA1634 & SA1636 to deal with other copyright header elements. If the header is already valid then this only replaces the <copyright> tag contents. Otherwise it does the previous behaviour of replacing the whole header.

Also removed trailing spaces on blank lines in copyright text.

Fixes #1434

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
@@ -140,11 +140,6 @@ namespace StyleCop.Analyzers.DocumentationRules
                 }
             }
 
-            if (isMalformedHeader)
-            {
-                copyrightTriviaIndex = null;
-            }
-
             // Remove copyright lines in reverse order
             removalList.Reverse();
             foreach (var triviaLine in removalList)
@@ -155,23 +150,20 @@ namespace StyleCop.Analyzers.DocumentationRules
             string newLineText = document.Project.Solution.Workspace.Options.GetOption(FormattingOptions.NewLine, LanguageNames.CSharp);
 
             var newHeaderTrivia = CreateNewHeader(document.Name, settings, newLineText);
-            if (copyrightTriviaIndex.HasValue)
+            if (!isMalformedHeader && copyrightTriviaIndex.HasValue)
             {
-                if (inCopyright)
-                {
-                    newHeaderTrivia = newHeaderTrivia.Add(SyntaxFactory.CarriageReturnLineFeed);
-                }
-
-                trivia = trivia.ReplaceRange(trivia[copyrightTriviaIndex.Value], newHeaderTrivia);
-                return root.WithLeadingTrivia(trivia);
+                // Replace copyright element in place.
+                return root.WithLeadingTrivia(trivia.ReplaceRange(trivia[copyrightTriviaIndex.Value], newHeaderTrivia));
             }
             else
             {
+                // Add blank line if we don't already have comments at top of file.
                 if ((trivia.Count == 0) || (trivia[0].Kind() != SyntaxKind.SingleLineCommentTrivia))
                 {
                     newHeaderTrivia = newHeaderTrivia.Add(SyntaxFactory.CarriageReturnLineFeed);
                 }
 
+                // Insert header at top of the file.
                 return root.WithLeadingTrivia(newHeaderTrivia.Add(SyntaxFactory.CarriageReturnLineFeed).AddRange(trivia));
             }
         }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
@@ -3,9 +3,11 @@
 
 namespace StyleCop.Analyzers.DocumentationRules
 {
+    using System;
     using System.Collections.Immutable;
     using System.Composition;
     using System.Linq;
+    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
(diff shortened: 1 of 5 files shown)


sharwell:
I'm looking forward to this one!

peteri:
Don't think they duplicate (although I did get merge grief)
Currently the fixer doesn't deal with #1432 but now the other code is in place I'm happy to have a look.

peteri:
Realised I had some unreachable code, so removed it and cleaned up the logic, also removed an inheritdoc xml comment tag that had appeared somehow.

vweijsters:
Valid XML file headers are not handled (See #1432)

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -106,8 +140,32 @@ private static SyntaxNode ReplaceHeader(Document document, SyntaxNode root, Styl
...
+            var newHeaderTrivia = CreateNewHeader(document.Name, settings, newLineText);
+            if (!isMalformedHeader && copyrightTriviaIndex.HasValue)
+            {
+                // Replace copyright element in place.
+                return root.WithLeadingTrivia(trivia.ReplaceRange(trivia[copyrightTriviaIndex.Value], newHeaderTrivia));
+            }
+            else
+            {
+                // Add blank line if we don't already have comments at top of file.
+                if ((trivia.Count == 0) || (trivia[0].Kind() != SyntaxKind.SingleLineCommentTrivia))
```
vweijsters: `trivia[0].Kind() != SyntaxKind.SingleLineCommentTrivia` can be replaced by `!trivia[0].IsKind(SyntaxKind.SingleLineCommentTrivia)`

vweijsters:
I'm missing test cases for a single line copyright block. For example:
```C#
// <copyright file=""Test0.cs"" company=""FooCorp""> Copyright (c) FooCorp. All rights reserved.</copyright>
// <author>John Doe</author>

namespace Test
{
}
```

vweijsters:
:thought_balloon: The code is currently very difficult to analyze for correctness because there is no separation between cases where an XML header is expected or not, depending on the settings (StyleCopSettings.DocumentationRules.XmlHeader). My proposal would be to have two `ReplaceHeader` implementations, one for XML based headers and one for non-XML headers.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -56,48 +57,81 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                            var closingTag = triviaLine.ToFullString().Contains("</copyright>") ||
+                                (openingTag && triviaLine.ToFullString().Trim().EndsWith("/>"));
+                            if (openingTag)
+                            {
+                                inCopyright = !closingTag;
+                                copyrightTriviaIndex = i;
+                            }
+                            else
+                            {
+                                if (inCopyright)
```
vweijsters: :thought_balloon: Reformatting to `else if (inCopyright)` would improve readability

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -56,48 +57,81 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
             var settings = document.Project.AnalyzerOptions.GetStyleCopSettings();
 
             var fileHeader = FileHeaderHelpers.ParseFileHeader(root);
-            var newSyntaxRoot = fileHeader.IsMissing ? AddHeader(document, root, document.Name, settings) : ReplaceHeader(document, root, settings);
+            var xmlFileHeader = FileHeaderHelpers.ParseXmlFileHeader(root);
+            var newSyntaxRoot = fileHeader.IsMissing ? AddHeader(document, root, document.Name, settings) : ReplaceHeader(document, root, settings, xmlFileHeader.IsMalformed);
```
vweijsters: The code above should select the correct parsing method based on the `XmlHeader` setting.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -106,8 +140,32 @@ private static SyntaxNode ReplaceHeader(Document document, SyntaxNode root, Styl
                 }
             }
 
+            // Remove copyright lines in reverse order
+            removalList.Reverse();
+            foreach (var triviaLine in removalList)
+            {
+                trivia = trivia.RemoveAt(triviaLine);
+            }
```
vweijsters: This can be implemented more efficiently using an for loop with a decreasing indexer.

vweijsters:
The `AddHeader` method retrieves the correct newline to use and the adds `SyntaxFactory.CarriageReturnLineFeed` instead of the retrieved newline setting.

(Pre-existing issue)

vweijsters:
There is an issue if the current file header has indentation. The copyright part would end up without any indentation, while the indentation for the rest of the header is preserved.

The previously implementation would remove all indentation, which is not correct, but preferable to having a mixed mode.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -106,8 +140,32 @@ private static SyntaxNode ReplaceHeader(Document document, SyntaxNode root, Styl
...
+            {
+                // Add blank line if we don't already have comments at top of file.
+                if ((trivia.Count == 0) || (trivia[0].Kind() != SyntaxKind.SingleLineCommentTrivia))
+                {
+                    newHeaderTrivia = newHeaderTrivia.Add(SyntaxFactory.CarriageReturnLineFeed);
+                }
+
+                // Insert header at top of the file.
+                return root.WithLeadingTrivia(newHeaderTrivia.Add(SyntaxFactory.CarriageReturnLineFeed).AddRange(trivia));
+            }
```
vweijsters: Line 163 and 167 should use the newLineText, instead of `SyntaxFactory.CarriageReturnLineFeed`

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
peteri: I'll review this once I've had a look at doing the multiline comments, but I think you're correct and two lots of code would be cleaner.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
peteri: Didn't spot that (will change)

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
peteri: Okay I'll change that.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
peteri: Re: The indentation @sharwell in #1356 goes with the view that spaces should be added in the settings to the copyright text, I tend to agree as it gives a decent level of control and is at least predictable (whereas a magic autoindenter can go wrong easily).

Re Lines 163 and 167 this needs to be SyntaxFactory.EndOfLine(newLineText);
Is there an issue already raised for the other instances in the code base?
Feels like something that really needs an helper method somewhere.

vweijsters: I added a helper method for the end of line in #1511, but that is not merged yet.

Regarding the indentation I think there is an unclarity. I was talking about leading spaces before the start of the single line comment and that cannot be handled by additional spaces in the settings.

peteri: Ah see what you mean now. You're talking about 

```C#
   // <author>Peter</author>
   // <copyright>FooBar</copyright>
   // <summary>Stuff<summary>
```

with the old code would have been turned into

```C#
// <copyright>FooBar</copyright>
```

but now looks like this

```C#
   // <author>Peter</author>
// <copyright>FooBar</copyright>
   // <summary>Stuff<summary>
```

sharwell:
@peteri Please let me know if you get stuck on this. As I'm sure you're aware, this is a very important bug fix to get out to users. :+1: I checked with @vweijsters today and this is the current status as we currently see it:

> #1505 still has issues with multiline comment headers as described in #1432. I consider that blocking

peteri:
I had an attempt yesterday, but made too many changes at once between running tests and made a complete hash of it.
I will have a think while I'm at work and get the implementation done this evening (I'm in the UK). I'll let you know how far I get.

peteri:
Looking pretty good for finishing tomorrow.

peteri:
Could probably do with some more edge case tests but should be there now.

vweijsters:
:+1: Most of it looks good.  Multiline comment trivia are handled properly now. There is one remark left that I would prefer to be addressed.

peteri:
Hmm the single line comment check is so that this code 

    // <summary>Foo</summary>

comes out like this after the fix is run (I've used a single line comment for the copyright here so it's fairly brief)

```
// <copyright>Single line</copyright>
// <summary>Foo</summary>
```

but
```
#define STUFF
```

Looks like this after the fix

```
// <copyright>Single line</copyright>

#define STUFF
```

So for the multi line comment case (which can't be valid xml otherwise we would have gone down the other path)

```
/************** 
*  DON'T MOVE *
**************/
```

Will end up looking like this option (A):

```
// <copyright>Single line</copyright>

/************** 
*  DON'T MOVE *
**************/
```

Rather than option (B) below
```
// <copyright>Single line</copyright>
/************** 
*  DON'T MOVE *
**************/
```

My personal preference is for option A (a happy accident of the code) but I'm really not that fussy.

I'm out after work this evening but I can make a change tomorrow (Thursday) if it's wanted.

sharwell:
Thanks for all the hard work on this y'all! :smile:

Issue #1482, opened 2015-09-15
Implement SA1129 (DoNotUseDefaultValueTypeConstructor)


sharwell:
**Category:** Readability
**Title:** Use `default(T)` to create default value type instances
**Description:** A default instance of a value type `T` may be obtained by either `default(T)` or `new T()`. In these scenarios, the `default(T)` syntax should be used.

**Rationale:** This is especially applicable for code using `ImmutableArray<T>` (a value type). Code which formerly used `List<T>` or `T[]` (both reference types) need to be very careful not to mistake `new ImmutableArray<T>()` for the same behavior. This syntax results in a default `ImmutableArray<T>` instance, whose behavior more closely resembles `null` than any non-null instance (especially in regards to exceptions being thrown). Similar behavior can be seen for `Nullable<T>`.

:memo: This proposal only refers to the distinction between `default(T)` and `new T()`. Other default values, including `CancellationToken.None`, `0`, `0.0f`, `IntPtr.Zero`, and `RegexOptions.None`, would *not* produce a warning.

:memo: This proposal was inspired by code observed in [dotnet/Roslyn#5180 (comment)](https://github.com/dotnet/roslyn/pull/5180#discussion_r39443514).

vweijsters:
:+1:

Maxwe11:
:+1:

otac0n:
I'm :-1:, because the advice boils down to "never use parameterless constructors for structs" which, to me, is quite dubious.

Basically, if `default(MyStruct)` and `new MyStruct()` behave differently, it would stand to reason that the author of the struct did this intentionally.  That is to say, there was a case in the author's mind that lead them to intentionally create a parameterless constructor that behaves differently than `default(T)`.  

So, in the *general* case, this seems wrong to me.

This *does* seems like very good advice for the Immutable Array case, so I think it should probably be added to Code Analysis.

pdelvo:
A struct can't have a parameterless constructor. So default(MyStruct) and new MyStruct() behave the same.

otac0n:
@pdelvo Fair point.  :+1:

sharwell:
@otac0n Does your :+1: mean you just retracted the :-1: vote, or are you adding a :+1: in favor of the proposal?

Tinister:
public struct Struct
    {
        public int Property { get; }
    
        public Struct(int value = 1)
        {
            this.Property = value;
        }
    }
    
[...]

Issue #1451, opened 2015-09-10
Implement SA1130 (UseLambdaSyntax)


otac0n:
**Category:** Readability
**Rule:** Use lambda syntax over delegate syntax.
**Description:** Replace usages of the `delegate` syntax with lambda expressions.
**Rationale:** Lamda syntax is used more widely and reads easier since it uses fewer characters.

These would yield a warning:

    foo += delegate { x = 0; };
    bar += delegate() { y = 0; };
    baz += delegate(int a, int b) { return a + b; };

This would not:

    var x = A(() => { }); // Expression
    var y = A(delegate { }); // Action

    private Expression<Action> A(Expression<Action> expression)
    {
        return expression;
    }

    private Action A(Action action)
    {
        return action;
    }

sharwell:
:+1:

ChrSteinert:
:+1:

Maxwe11:
:+1:

ChrSteinert:
Why would `var y = A(delegate { });` not yield a warning?

Maxwe11:
@ChrSteinert what do you mean? This rule not implemented yet :smile:

ChrSteinert:
I know, but I want to implement it ;) Sorry, for poor wording.
So the question is, why SHOULD mentioned snipped yield no warning. I'd argue that using delegate anywhere is not wanted. But the initial proposal says that this should be ok. So, why should this be ok?

pdelvo:
Using a lambda expression would change the semantics because now you would call a different overload. The question is if we want to have an exception for these cases or if we want to fix the ambiguosity in a code fix by adding a cast.

otac0n:
Because in that context, blindly converting to the lambda syntax would
change which overload gets called. This is the only case (as far as I was
able to work out) where the syntaxes are treated differently, and I expect
it to be exceedingly rare.

ChrSteinert:
Ahhh of course, now I get it. 
An really ingesting exception for a possible rule.

pdelvo:
Im grabbing this

Issue #863, opened 2015-05-18
Implement SA1131 (UseReadableConditions)


sharwell:
**Category:** Readability
**Rule:** Use readable conditions (alternatively: Do not use [Yoda conditions](http://en.wikipedia.org/wiki/Yoda_conditions))
**Description:** The equality and inequality operators `==` and `!=` are generally symmetric. In combination with #862, there is no longer a need to write expressions with a literal as the first operand. This rule is described as follows:

> The first operand of an equality operator should not be a literal (or `null`), unless the second operand is also a literal (or `null`).

vweijsters:
:+1: 

Yoda conditions had their uses in C / C++, but given the strict rules about valid conditions in C#, it no longer makes sense.

Przemyslaw-W:
+1
Similarly, if we stick to "use readable conditions", this could also cover
```if (booleanExpression)``` or ```if (!booleanCondition)
instead of
```if (booleanCondition == true)``` or ```if (booleanCondition == false)```

2015-05-18 21:49 GMT+02:00 Vincent Weijsters <notifications@github.com>:

**Comment by sharwell** [comment]
@Przemyslaw-W avoiding `== true` and `== false` should be a different rule, and needs further discussion due to potentially problematic interaction with #862.

**Comment by pdelvo** [comment]
grabbed.

**Comment by alxyr** [comment]
I'd like to :+1: the proposal from @Przemyslaw-W to generate a warning when an expression of boolean type is compared to `false` or `true`. For example, the warning should be generated in the following cases:
```C#
if (completedSuccessfully == true)
if (control.IsVisible == false)
if (dataStorage.GetValue("name") == true)
if ((bool)data[i] == true)
```

pdelvo:
Im currently working on this. Do we want to report on all occurences of '==' that has a constant value on its left and a non constant value on its right or only built in ones? You could write a non commutative == operator.

sharwell:
@pdelvo We should assume that operator `==` is commutative. Are there any cases where changing the order of operands can change which operator gets resolved (I'm thinking of implicit type conversions here)?

nriding:
I don't like this rule. I've been coding C/C++/C# for a long time. I've always written my comparisons <constant> == <value>, it's how all my code is formatted. I find it easier to read. The most important term come first. After updating StyleCop I now have to disable this rule in all my projects.

GregReddick:
[...]

Issue #825, opened 2015-05-11
Implement SA1132 (DoNotCombineFields)


sharwell:
**Category:** Readability
**Title:** Do not combine field definitions
**Description:** C# allows more than one field to be defined in a single field declaration syntax. When fields are combined, it becomes difficult to impossible to properly document them, and difficult to identify each field defined in a type.

**Example:**

The following code shows an example which would produce this warning:

```csharp
int x, y;
```

To resolve the violation, define in field separately:

```csharp
int x;
int y;
```

vweijsters:
:+1: Big +1 for me, that's one feature they should have never introduced IMHO.

johnkoerner:
I agree with this one 100%.

PR #1508, opened 2015-09-19
Add stub analyzers for new readability rules

sharwell:
This pull request assigns diagnostic IDs and adds initial documentation and a stub analyzer for each of the following proposals:

#805 
#825 
#863 
#1451
#1482 

If accepted, I'll update the title of each of the proposal issues to reflect the newly-assigned diagnostic IDs.


documentation/SA1130.md
```diff
@@ -0,0 +1,73 @@
...
+
+The following code shows the equivalent variable declarations using the more familiar lambda syntax.
+
+```csharp
+Action a = () => { x = 0; };
+Action b = () => { y = 0; };
+Func<int, int, int> c = (m, n) => m + n;
+```
+
+:memo: It is not always possible to replace an anonymous method with an equivalent lambda expression. For example, the
```
pdelvo: So no violation should be reported in this cases?

sharwell: That is correct.

pdelvo:
:+1:

PR #1509, opened 2015-09-19
Fixes that SA1101 analyzer doesnt work for generic names

### LINKED ISSUE #1500: Missing SA1101:PrefixLocalCallsWithThis for generic method calls
SA1101 is apparently not issued for generic method calls with explicitly specified generic arguments. The provided CodeFix skips this case as well.
```cs
public class Test_SA1101 : List<string>
{
    public void Foo()
    {
        Converter<string, int> cvt = null;

        ConvertAll(cvt); // SA1101, OK
        this.ConvertAll(cvt); // no SA1101, OK
        ConvertAll<int>(cvt); // no SA1101, NOT OK!!!
        this.ConvertAll<int>(cvt); // SA1101, OK
    }
}
```

Could you please have a look? Thanks!

pdelvo:
Fixes #1500

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1101UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1101UnitTests.cs
@@ -264,6 +264,43 @@ namespace StyleCop.Analyzers.Test.ReadabilityRules
         }
 
         [Fact]
+        public async Task TestPrefixLocalCallsWithThisWithGenericArgumentsAsync()
+        {
+            string testCode = @"public class Test_SA1101
+{
+    public void Foo()
+    {
+        ConvertAll(42); // SA1101, OK
+        this.ConvertAll(42); // no SA1101, OK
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1101UnitTests.cs
```diff
@@ -264,6 +264,43 @@ public void InstanceMethodName(int ParameterName)
...
+        public async Task TestPrefixLocalCallsWithThisWithGenericArgumentsAsync()
+        {
+            string testCode = @"public class Test_SA1101
+{
+    public void Foo()
+    {
+        ConvertAll(42); // SA1101, OK
+        this.ConvertAll(42); // no SA1101, OK
+        ConvertAll<int>(42); // no SA1101, NOT OK!!!
+        this.ConvertAll<int>(42); // SA1101, OK
```
sharwell: :bulb: This should say `no SA1101`

pdelvo: fixed

PR #1510, opened 2015-09-19
Implement SA1130

### LINKED ISSUE #1451: Implement SA1130 (UseLambdaSyntax)

pdelvo:
Fixes #1451.

visual studio provides a (local) code fix (without fix all). We probably want to implement one ourself.


sharwell:
I'd like to see a test included for overloads like the following:

```csharp
void Method(Action action) { }
void Method(Expression<Func<int>> expression) { }
```

pdelvo:
@sharwell I added that test

PR #1513, opened 2015-09-20
Remove unnecessary code

pdelvo:
We already had a variable for the `NullableTypeSyntax`.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1018NullableTypeSymbolsMustNotBePrecededBySpace.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1018NullableTypeSymbolsMustNotBePrecededBySpace.cs
@@ -68,8 +68,7 @@ namespace StyleCop.Analyzers.SpacingRules
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1018NullableTypeSymbolsMustNotBePrecededBySpace.cs
```diff
@@ -68,8 +68,7 @@ private static void HandleQuestionToken(SyntaxNodeAnalysisContext context)
                 return;
             }
 
-            NullableTypeSyntax parentNullableTypeSyntax = questionToken.Parent as NullableTypeSyntax;
-            if (parentNullableTypeSyntax != null && parentNullableTypeSyntax.ElementType.IsMissing)
+            if (nullableType != null && nullableType.ElementType.IsMissing)
```
sharwell: :bulb: `nullableType` can't be null here.

pdelvo: fixed

PR #1516, opened 2015-09-20
Implement SA1132 + unit tests

### LINKED ISSUE #825: Implement SA1132 (DoNotCombineFields)

Maxwe11:
Fixes #825

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1132UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1132UnitTests.cs
@@ -1,0 +1,119 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.Test.ReadabilityRules
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Analyzers.ReadabilityRules;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using TestHelper;
+    using Xunit;
+
+    public class SA1132UnitTests : CodeFixVerifier
+    {
+        [Theory]
+        [InlineData("private int a;")]
+        [InlineData("public event System.Action a;")]
+        public async Task TestValidDeclarationAsync(string declaration)
+        {
+            var testCode = GetTemplate(declaration);
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [InlineData("private int a, b;", "private int a;\r\n    private int b;")]
+        [InlineData("public event System.Action a, b;", "public event System.Action a;\r\n    public event System.Action b;")]
+        public async Task TestInvalidDeclarationAsync(string declaration, string fixedDeclaration)
+        {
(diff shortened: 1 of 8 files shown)


vweijsters:
:+1: Looks good to me

Maxwe11:
@vweijsters could you review second commit please? :smile:

vweijsters:
:+1:

vweijsters:
The current code fix code will duplicate any trivia that are not part of the variable declarations. For example:
```C#
public class TestClass
{
#if DEBUG
    private int field1, field2;
    private int field3;
#endif
}
```
will be fixed to:
```C#
public class TestClass
{
#if DEBUG
    private int field1;
#if DEBUG
    private int field2;
    private int field3;
#endif
}
```

Maxwe11:
@vweijsters see 89e77c4 for a fix. Previous commits just squash as is.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1132CodeFixProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (fieldDeclaration != null)
+            {
+                VariableDeclarationSyntax declaration = fieldDeclaration.Declaration;
+                SeparatedSyntaxList<VariableDeclaratorSyntax> variables = declaration.Variables;
+                VariableDeclaratorSyntax first = variables.First();
+                var newFieldDeclarations = new List<BaseFieldDeclarationSyntax>(variables.Count);
+
+                foreach (VariableDeclaratorSyntax variable in variables)
+                {
+                    var variableDeclarator = SyntaxFactory.SeparatedList(new[] { variable });
```
sharwell: :bulb: The downside of this syntax is it forces the code to use this overload even if a future release includes an overload of `SeparatedList` that takes exactly one argument.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1132UnitTests.cs
```diff
@@ -0,0 +1,123 @@
...
+        [InlineData("public event System.Action a;")]
+        public async Task TestValidDeclarationAsync(string declaration)
+        {
+            var testCode = GetTemplate(declaration);
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [InlineData("private int a, b;", "private int a;\r\n    private int b;")]
```
sharwell: :thought_balloon: I find this hard to read. I think I would have preferred two separate test methods instead.

vweijsters:
:+1: The fix for directives is OK. The code fix will still duplicate leading comments / trialing comments, but that is purely cosmetic and can be fixed later if it annoys people.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1132CodeFixProvider.cs
Maxwe11: fixed

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1132UnitTests.cs
Maxwe11: fixed

PR #1517, opened 2015-09-20
Improve SA1642 and SA1643 code fix

### LINKED ISSUE #415: SA1642 (constructor summary text) special case for incorrect class name
When SA1642 is reported, it is always reported for the entire `<summary>` element. In each of the following scenarios, a more specific location should be used for reporting the warning.

1. When the name of the class is not wrapped in a `<see>` element, but the text is otherwise correct. For example, the warning reported for the following code should only underline the word `Customer`:

    ```csharp
    /// <summary>
    /// Initializes a new instance of the Customer class.
    /// </summary>
    public Customer() { }
    ```

1. As in the first case, but now considering an incorrect class name. For example:

    ```csharp
    /// <summary>
    /// Initializes a new instance of the NotTheRightType class.
    /// </summary>
    public Customer() { }
    ```

1. When the name of the class specified in the `cref` attribute does not match the name of the current class, but the text is otherwise correct. In this case, the warning should be reported for the `cref` value instead of the entire `<summary>` element. For example:

    ```csharp
    /// <summary>
    /// Initializes a new instance of the <see cref="NotTheRightType"/> class.
    /// </summary>
    public Customer() { }
    ```

pdelvo:
Fixes #415.

With these changes the code fix does a better job detecting mistakes in the standard text. If the standard text has the wrong class name in its text it will fix that instead of inserting a new standard text at the beginning of the summary. Examples:

It replaces a text form of the class name with a see tag:
```csharp
/// <summary>
/// Initializes a new instance of the Foo class.
/// </summary>
```
->
```csharp
/// <summary>
/// Initializes a new instance of the <see cref="Foo"/> class.
/// </summary>
```
It fixes a wrong see tag:
```csharp
/// <summary>
/// Initializes a new instance of the <see cref="Bar"/> class.
/// </summary>
```
->
```csharp
/// <summary>
/// Initializes a new instance of the <see cref="Foo"/> class.
/// </summary>
```
And it fixes a wrong class text with the correct class name:
```csharp
/// <summary>
/// Initializes a new instance of the Bar class.
/// </summary>
```
->
```csharp
/// <summary>
/// Initializes a new instance of the <see cref="Foo"/> class.
/// </summary>
```

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1642SA1643CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1642SA1643CodeFixProvider.cs
@@ -7,6 +7,7 @@ namespace StyleCop.Analyzers.DocumentationRules
     using System.Collections.Immutable;
     using System.Composition;
     using System.Linq;
+    using System.Text.RegularExpressions;
     using System.Threading.Tasks;
     using Helpers;
     using Microsoft.CodeAnalysis;
@@ -119,9 +120,13 @@ namespace StyleCop.Analyzers.DocumentationRules
             }
 
             string newLineText = document.Project.Solution.Workspace.Options.GetOption(FormattingOptions.NewLine, LanguageNames.CSharp);
-            var list = BuildStandardText(typeDeclaration.Identifier, typeParameterList, newLineText, standardText[0], standardText[1]);
 
-            var newContent = node.Content.InsertRange(0, list);
+            string trailingString = string.Empty;
+
+            var newContent = RemoveMalformattedStandardText(node.Content, typeDeclaration.Identifier, standardText[0], standardText[1], ref trailingString);
+
+            var list = BuildStandardText(typeDeclaration.Identifier, typeParameterList, newLineText, standardText[0], standardText[1] + trailingString);
+            newContent = newContent.InsertRange(0, list);
             var newNode = node.WithContent(newContent).AdjustDocumentationCommentNewLineTrivia();
 
             var newRoot = root.ReplaceNode(node, newNode);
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1642SA1643CodeFixProvider.cs
```diff
@@ -53,13 +54,19 @@ public override FixAllProvider GetFixAllProvider()
...
+                var xmlElementSyntax = node as XmlElementSyntax;
+
+                if (xmlElementSyntax != null)
                 {
-                    continue;
+                    context.RegisterCodeFix(CodeAction.Create(DocumentationResources.SA1642SA1643CodeFix, token => GetTransformedDocumentAsync(context.Document, root, xmlElementSyntax), equivalenceKey: nameof(SA1642SA1643CodeFixProvider)), diagnostic);
+                }
+                else
+                {
+                    var xmlEmptyElementSyntax = node as XmlEmptyElementSyntax;
```
sharwell: :bulb: Don't assume the result of `as` is non-null. Either use a normal cast instead, or include a null check.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1642SA1643CodeFixProvider.cs
```diff
@@ -53,13 +54,19 @@ public override FixAllProvider GetFixAllProvider()
                     continue;
                 }
 
-                var node = root.FindNode(diagnostic.Location.SourceSpan, findInsideTrivia: true) as XmlElementSyntax;
-                if (node == null)
+                var node = root.FindNode(diagnostic.Location.SourceSpan, findInsideTrivia: true) as XmlNodeSyntax;
```
sharwell: We've had some problems in the past where failing to specify `getInnermostNodeForTie: true` resulted in exceptions when running the code fix on documents with syntax errors. In addition, the `as XmlNodeSyntax` appears to be unnecessary. I recommend the following:

```csharp
SyntaxNode node = root.FindNode(diagnostic.Location.SourceSpan, findInsideTrivia: true, getInnermostNodeForTie: true);
```

pdelvo:
@sharwell i implemented your feedback

PR #1518, opened 2015-09-20
Implemented SA1129 analyzer, code fix, and tests

### LINKED ISSUE #1482: Implement SA1129 (DoNotUseDefaultValueTypeConstructor)

vweijsters:
Fixes #1482

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
@@ -62,6 +62,7 @@ namespace StyleCop.Analyzers.Test.ReadabilityRules
         var v1 = new TestStruct(1);
         var v2 = new TestStruct(1) { TestProperty = 2 };
         var v3 = new TestStruct() { TestProperty = 2 };
+        var v4 = new TestStruct { TestProperty = 2 };
     }
 
     private struct TestStruct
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1129CodeFixProvider.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        /// <inheritdoc/>
+        public override FixAllProvider GetFixAllProvider()
+        {
+            return FixAll.Instance;
+        }
+
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            foreach (var diagnostic in context.Diagnostics.Where(d => this.FixableDiagnosticIds.Contains(d.Id)))
```
Maxwe11: There is no need in `Where` clause

Maxwe11:
one minor comment but otherwise :+1:

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        [Fact]
+        public async Task VerifyValueTypeWithParametersCreationAsync()
+        {
+            var testCode = @"public class TestClass
+{
+    public void TestMethod()
+    {
+        var v1 = new TestStruct(1);
+        var v2 = new TestStruct(1) { TestProperty = 2 };
+        var v3 = new TestStruct() { TestProperty = 2 };
```
sharwell: Nice catch finding this one!

:exclamation: You need to add one more for the following case (I tried adding it during the merge but it resulted in an AD0001 failure):

```csharp
var v4 = new TestStruct { TestProperty = 2 };
```

sharwell:
I'm missing some tests:

### `struct` generic type constraint

The following code *should* produce a violation.

```csharp
public void Method<T>()
    where T : struct
{
    T value = new T(); // SA1129
}
```

### `new` generic type constraint

The following code *should not* produce a violation.

```csharp
public void Method<T>()
    where T : new()
{
    T value = new T();
}
```

vweijsters:
Updated with CR feedback

Issue #1501, opened 2015-09-18
SA1101: "Prefix calls with this" incorrectly handles properties as expression bodies


lowkay:
When using expression bodies for properties the diagnostic for SA1101 is incorrectly detected when the property name is the same as a type/enum, i.e.

```C#
    public enum Lifetime
    {
        Persistent,

        Transient,

        Scoped
    }

   public class Example
    {
        public Lifetime Lifetime => Lifetime.Persistent;
        //                          ^ Diagnostic SA1101 incorrectly detected here.
    }
```

It is not detected if this is a declared getter:

```C#

   public class Example
   {
      public Lifetime Lifetime
      {
           get
           {
               return Lifetime.Persistent;
           }
      }
   }
```

sharwell:
*Weird* :bug: catch!

pdelvo:
Im currently looking into this and I have trouble reproducing this. Is the code you posted exactly the one that was reporting SA1101? Which version were you using?

lowkay:
Howdy,

It's using the latest pre-release from nuget. Take a look at the example solution here:
https://github.com/lowkay/StyleCopAnalyzerBugs/tree/master/Bug1501

pdelvo:
![bug1501](https://cloud.githubusercontent.com/assets/1710209/9989629/79c0a7a4-605a-11e5-9f8c-13f69f7724a0.PNG)
I tried to repro this with your project but I dont see any problem. I tried both beta12 (left) and the current master (right). Maybe someone else can try to repro it?

lowkay:
hmm, strange - no squiggle, but the build fails with SA1101 violations:

![image](https://cloud.githubusercontent.com/assets/6552669/9989780/6dd7a8b8-6053-11e5-8119-9ffdb332a9af.png)

pdelvo:
Wow thats weird! It shows up for me in the msbuild output as well but not the error list.

lowkay:
We've experienced some errors not getting into the error list before, always thought it was a VS bug - as we have quite a large solution and VS isn't always good at updating the error list and as I understand it adding the diagnostic is all you need to do and VS should update the relevant UX...

pdelvo:
[...]

PR #1523, opened 2015-09-21
Implement a workaround for #1501

pdelvo:
Unfortunatly the bug does not show up in unit tests so I cant write one.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1101PrefixLocalCallsWithThis.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1101PrefixLocalCallsWithThis.cs
@@ -189,6 +189,20 @@ namespace StyleCop.Analyzers.ReadabilityRules
                 {
                     return;
                 }
+
+                // This is a workaround for https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1501 and can
+                // be removed when the underlying bug in roslyn is resolved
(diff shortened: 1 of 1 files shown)


sharwell:
Does the bug also appear for static members of other types? For example, what if the `Lifetime` type from #1501 was implemented as follows:

```csharp
public class Lifetime
{
    public static readonly Lifetime Persistent = new Lifetime();
}
```

pdelvo:
yes it does. I removed the enum check

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1101PrefixLocalCallsWithThis.cs
```diff
@@ -189,6 +189,20 @@ private static void HandleIdentifierNameImpl(SyntaxNodeAnalysisContext context,
...
                     return;
                 }
+
+                // This is a workaround for https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1501 and can
+                // be removed when the underlying bug in roslyn is resolved
+                if (nameExpression.Parent is MemberAccessExpressionSyntax)
+                {
+                    var parentSymbol = context.SemanticModel.GetSymbolInfo(nameExpression.Parent, context.CancellationToken).Symbol as IFieldSymbol;
+
+                    if (parentSymbol.ContainingType.Name == symbol.Name)
```
sharwell: :question: Does this also need to check if the member is static?

PR #1526, opened 2015-09-22
Implement SA1305 and unit tests

### LINKED ISSUE #81: SA1305: FieldNamesMustNotUseHungarianNotation
http://www.stylecop.com/docs/SA1305.html

Noryoko:
Fixes #81

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1305UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1305UnitTests.cs
@@ -1,0 +1,205 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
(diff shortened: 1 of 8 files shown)


sharwell:
I'm torn. I really question the value added by this rule but you appear to have done a good job with the implementation. I'm going to do some experiments with known larger code bases and see what I find.

sharwell:
CodeContracts.sln currently contains around 1000 violations of this rule. While I did find a few cases where Hungarian notation (the bad kind) was used, it seemed like the majority of reported violations were actually due to other causes. Examples include:

| Name | Meaning |
| --- | --- |
| `bAsSet` | `b as Set<T>` |
| `ehEnd` | Exception handler end |
| `mdDecoder` | Metadata decoder |
| `ilDecoder` | IL decoder |
| `ctDecoder` | Contract decoder |
| `idCount` | Number of IDs |
| `opString` | Operator string |

sharwell:
:bulb: The diagnostic is still marked as `DisabledNoTests`. Depending on the outcome of #1530, it should be changed to either `EnabledByDefault` or `DisabledByDefault`.

PR #1536, opened 2015-09-24
Added logic to escape characters in XmlComments in the fixer.

peteri:
I realised we don't cope with &, <, > and friends in the wrap xml header logic.
Fix & test code attached.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
@@ -353,8 +353,8 @@ namespace StyleCop.Analyzers.DocumentationRules
 
         private static string WrapInXmlComment(string prefixWithLeadingSpaces, string copyrightText, string filename, StyleCopSettings settings, string newLineText)
         {
-            string encodedFilename = new XText(filename).ToString();
-            string encodedCompanyName = new XText(settings.DocumentationRules.CompanyName).ToString();
+            string encodedFilename = new XAttribute("t", filename).ToString().Substring(2).Trim('"');
+            string encodedCompanyName = new XAttribute("t", settings.DocumentationRules.CompanyName).ToString().Substring(2).Trim('"');
             string encodedCopyrightText = new XText(copyrightText).ToString();
 
             return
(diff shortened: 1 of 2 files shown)


peteri:
Oh meant to say apologies if I should have raised an issue then attached the pull request to it.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -352,9 +353,13 @@ private static SyntaxTriviaList CreateNewHeader(string prefixWithLeadingSpaces,
 
         private static string WrapInXmlComment(string prefixWithLeadingSpaces, string copyrightText, string filename, StyleCopSettings settings, string newLineText)
         {
+            string encodedFilename = new XText(filename).ToString();
+            string encodedCompanyName = new XText(settings.DocumentationRules.CompanyName).ToString();
```
sharwell: This doesn't handle `"` characters properly, since it appears in an attribute. It should use the following instead:

```csharp
string encodedFilename = new XAttribute("t", filename).ToString().Substring(2).Trim('"');
string encodedCompanyName = new XAttribute("t", settings.DocumentationRules.CompanyName).ToString().Substring(2).Trim('"');
```

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -352,9 +353,13 @@ private static SyntaxTriviaList CreateNewHeader(string prefixWithLeadingSpaces,
...
+            string encodedCopyrightText = new XText(copyrightText).ToString();
```
sharwell: :memo: This one is correct.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderCodeFixProvider.cs
peteri: Well I learnt something new today (Never thought about it before...) I'll fix and add quotes into the company name for the tests.

peteri:
Good catch, let me know if you want me to squash it down.

PR #1546, opened 2015-09-25
Exclude compiler-generated code from coverage analysis

sharwell:


pdelvo:
:+1:

sharwell:
I'm not so sure this is what we want. The goal was to exclude the unreachable portions of the state machine generated for `async` methods, but it turned out to completely exclude all `async` methods as well as iterator blocks. I'm not sure what the right approach is yet.

PR #1549, opened 2015-09-26
Add Hungarian notation configuration documentation

### LINKED ISSUE #1542: Document naming rules
Pull request #1526 added new settings to **stylecop.json**. These settings should be documented in [Configuration.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md).

Noryoko:
Fixes #1542

### DIFF
--- a/documentation/Configuration.md
+++ b/documentation/Configuration.md
@@ -164,3 +164,40 @@ The following example shows a configuration file which requires developers to do
   }
 }
 ```
+
+## Naming Rules
+
+This section describes the features of naming rules which can be configured in **stylecop.json**. Each of the described properties are configured in the `namingRules` object, which is shown in the following sample file.
+
+```json
+{
+  "settings": {
+    "namingRules": {
+    }
+  }
+}
+```
+
+### Hungarian Notation
+
+The following properties are used to configure allowable Hungarian notation prefixes in StyleCop Analyzers.
+
+| Property | Default Value | Summary |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


documentation/Configuration.md
```diff
@@ -164,3 +164,25 @@ The following example shows a configuration file which requires developers to do
...
+}
+```
+
+### Hungarian Notation
+
+The following properties are used to configure allowable Hungarian notation prefixes in StyleCop Analyzers.
+
+| Property | Default Value | Summary |
+| --- | --- | --- |
+| `allowCommonHungarianPrefixes` | **true** | Specifies whether a set of common prefixes ("as", "at", "by", "do", "go", "if", "in", "is", "it", "no", "of", "on", "or", "to") should be allowed |
```
sharwell: :thought_balloon: Thinking of the summary:

> Specifies whether common non-Hungarian notation prefixes should be allowed. When true, to two-letter words 'as', 'at', 'by', 'do', 'go', 'if', 'in', 'is', 'it', 'no', 'of', 'on', 'or', and 'to' are allowed to appear as prefixes for variable names.

documentation/Configuration.md
```diff
@@ -164,3 +164,25 @@ The following example shows a configuration file which requires developers to do
...
+| `allowedHungarianPrefixes` | `[ ]` | Specifies allowable Hungarian notation prefixes |
```
sharwell: :thought_balloon: 

> Specifies additional prefixes which are allowed to be used in variable names. See the example below for more information.

Then right after the table you could have something like this:

> The following example shows a settings file which allows the common prefixes as well as the custom prefixes 'md' and 'cd'.
>
> ```json
> {
>   "settings": {
>     "namingRules": {
>       "allowedHungarianPrefixes": [
>         "cd",
>         "md"
>       ]
>     }
>   }
> }
> ```

Issue #1144, opened 2015-08-07
Guide to replace StyleCop with StyleCop.Analyzers


hickford:
I'd like to replace StyleCop in our project with StyleCop.Analyzers. [StyleCop "classic"](https://stylecop.codeplex.com/) is slow and tedious (you have to fix everything [by hand](http://stackoverflow.com/a/12407781/284795)), whereas I understand StyleCop.Analyzers can fix errors automatically. That sounds 

Is there a how-to guide? 

1. How to uninstall StyleCop (StyleCop.dll, StyleCop.targets, StyleCopTask)
2. How to install StyleCop.Analyzers
3. How to configure StyleCop.Analyzers to use your existing [`StyleCop.Settings`](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/422), or convert it to a `.ruleset` 

I'm sure others would find it useful.

bsivanov:
It seems a [guy at stackoverflow](http://stackoverflow.com/a/31741929/259206) created an [F# script](http://fssnip.net/ss) that migrate the Settings.StyleCop to the .ruleset format. Maybe it could be included in the guide, eventually with some modifications.

journeyman:
@bsivanov @hickford hi, i tried to explain the usage in the comments. But if there're any questions, you can ask me directly and we probably can generalise it a bit (or rewrite completely :smiley: )

meh-uk:
I would like to see this.

Issue #1099, opened 2015-08-02
Add migrating documentation


pdelvo:
I saw that a few people find that configuring rules is not straight forward. We should probably include documentation and maybe even a script that converts Settings.StyleCop files into rule sets.

See
http://stackoverflow.com/questions/31313249/stylecop-integration-with-visual-studio-2015-rc/31741929#31741929
and
https://twitter.com/pedrolamas/status/627593865323761664

journeyman:
I made small F# script (probably not very generic one, but works for me). Maybe someone will find it useful
http://fssnip.net/ss

PR #1550, opened 2015-09-26
created Migrate.md

journeyman:
addresses requests from: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1099 and https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1144


documentation/Migrate.md
```diff
@@ -0,0 +1,16 @@
+## Migration from Settings.StyleCop to FxCop.ruleset (using F#)
```
sharwell: :thought_balloon: I'm a bit torn on this. It looks like a lot of work for relatively little gain. It's quite easy in my experience to use the rule set editor to configure the severity of rules. I have the following additional specific concerns:

1. This only covers the severity of rules. Users will still have to configure **stylecop.json** separately and/or reconsider the severity of rules in cases where the behavior of this project does not precisely match StyleCop "classic".
1. This script could enable rules which are currently disabled because they are either not implemented or are completely untested. It seems like it could manually configure the severity of rules which are permanently disabled, such as SA1603 (DocumentationMustContainValidXml) (see #1324). When rules are permanently disabled, a diagnostic descriptor for them is retained in order for the status site to extract information from them (it doesn't use it currently), but the diagnostic is hidden from the rule set editor interface.

documentation/Configuration.md
```diff
@@ -6,6 +6,7 @@ StyleCop Analyzers is configured using two separate mechanisms: code analysis ru
 
    * Enable and disable individual rules
    * Configure the severity of violations reported by individual rules
+   * migrate from old **StyleCop.Settings** to new .ruleset format (see [Migrate.md](documentation/Migrate.md))
```
sharwell: :thought_balloon: This bulleted list is intended to describe the functionality of the new rule set files, but this new item describes something else. A better placement for this as part of the descriptive text outside of the list itself.

pdelvo:
Do we want to move this PR forward?

dmitryvhf (approved):
Script checked. Easy in using.

sharwell:
Closing this one since the changes as proposed have become outdated. The .ruleset format has been superseded by .editorconfig (and later by .globalconfig).

PR #1551, opened 2015-09-26
Improve SA1407 and SA1408 fix all provider

pdelvo:
This simplifies the fix all provider. This might also be a little bit more efficient because we dont need to annotate the tree.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1407SA1408FixAllProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1407SA1408FixAllProvider.cs
@@ -3,6 +3,7 @@
 
 namespace StyleCop.Analyzers.MaintainabilityRules
 {
+    using System.Collections.Generic;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CodeFixes;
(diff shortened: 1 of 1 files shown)


sharwell:
:+1:

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1407SA1408FixAllProvider.cs
```diff
@@ -40,10 +36,10 @@ internal sealed class SA1407SA1408FixAllProvider : DocumentBasedFixAllProvider
...
 
-                root = root.ReplaceNode(node, node.WithAdditionalAnnotations(NeedsParenthesisAnnotation));
+                nodes.Add(node);
             }
 
-            return root.ReplaceNodes(root.GetAnnotatedNodes(NeedsParenthesisAnnotation), this.AddParentheses);
+            return root.ReplaceNodes(nodes, (originalNode, rewrittenNode) => this.AddParentheses(originalNode, rewrittenNode));
         }
 
         private SyntaxNode AddParentheses(SyntaxNode originalNode, SyntaxNode rewrittenNode)
```
sharwell: :bulb: This can be static.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1407SA1408FixAllProvider.cs
```diff
@@ -25,13 +24,10 @@ internal sealed class SA1407SA1408FixAllProvider : DocumentBasedFixAllProvider
...
 
-            var root = await newDocument.GetSyntaxRootAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
+            List<SyntaxNode> nodes = new List<SyntaxNode>();
 
-            // First annotate all expressions that need parenthesis with a temporary annotation.
-            // With this annotation we can find the nodes that need parenthesis even if
-            // the source span changes.
             foreach (var diagnostic in diagnostics)
             {
                 SyntaxNode node = root.FindNode(diagnostic.Location.SourceSpan);
```
sharwell: :thought_balloon: We should investigate whether this needs `getInnermostNodeForTie: true`.

PR #1561, opened 2015-09-27
Restore separate phases for test and code coverage

### LINKED ISSUE #1560: CI reports "success" when test failed
https://ci.appveyor.com/project/sharwell/stylecopanalyzers/build/1.0.1897

sharwell:
Fixes #1560

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,7 @@ before_build:
 build:
   project: StyleCopAnalyzers.sln
   verbosity: minimal
-test_script:
+after_test:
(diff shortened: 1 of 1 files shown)


sharwell:
@stevepeak The codecov documentation for C# projects recommends the solution implemented as part of this pull request. If there was a way to instead combine the test and code coverage passes, it would substantially improve the build times for your users. The only problem I observed by using `test_script` instead of `after_test` is the one described in #1560.

stevepeak:
Just to make sure I'm understanding this: 

- If you run `codecov` in `test_script` you may get false positives from AppVeyor because the `codecov` command was successful?
- Does AppVeyor quit upon the first failure in `test_script`?

Sorry for my lack of answer atm.

sharwell:
@stevepeak I actually found out a way to do this last night!

I modified the code coverage gathering step (which uses OpenCover) to report the status of XUnit instead of the status code for OpenCover itself. AppVeyor picked up on the change and now we can run our code coverage and test steps together. You can see the result in #1561.

:memo: In the event of a test failure, no data is sent to codecov.io. I do not know if this is the same as the original behavior (is the `after_test` script run if the `test_script` fails?). Either way, for me this is not a problem and combining the steps is a welcome performance improvement.

stevepeak:
Would you suggest this be noted in our documentation? Thank you :)

sharwell:
I filed codecov/example-csharp#3 with my suggestions :+1:

PR #1574, opened 2015-09-29
SA1026: handle nodes instead of syntax trees

Maxwe11:


vweijsters:
:+1: Looks good to me

PR #1575, opened 2015-09-29
Remove SA1111 analysis of attribute brackets

### LINKED ISSUE #1557: SA1111 message for attribute brackets is confusing
Currently SA1111 is reported for `]` character in the following code:

```csharp
[
Obsolete()
]
```

The message provided for this diagnostic is:

> Closing parenthesis must be on line of last parameter

The code fix for the diagnostic changes the code to the following:

```csharp
[
Obsolete()]
```

:memo: Note that SA1111 is *not* reported for the following code:

```csharp
[
Obsolete
]
```

I'm not sure what the expected behavior is, but I'm fairly sure it's one of the following:

1. SA1111 never should have been reported. If it was not reported for the original sample above, then it also would not have been confusing.
2. If SA1111 is supposed to be reported here, then both of the following changes need to be made:
  1. Use a different message which is less confusing.
  2. Update the code to still report the diagnostic for the second example.

sharwell:
Fixes #1557

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1111UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1111UnitTests.cs
@@ -1140,18 +1140,8 @@ using System.Diagnostics;
 public class FooAttribute : System.Attribute
 {
 }";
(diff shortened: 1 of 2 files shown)


vweijsters:
:+1: Looks good to me

PR #1576, opened 2015-09-29
Improved SA1500 FixAll behavior

### LINKED ISSUE #1565: Implement a fix all provider for SA1500 (CurlyBracketsForMultiLineStatementsMustNotShareLine)
I received an exception trying to fix all instances of this violation in a project today. The problem stems from poor memory behavior of the batch fixer, and both the error and the performance problems could be resolved by implementing a custom fix all provider for this diagnostic.

vweijsters:
Fixes #1565

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CodeFixProvider.cs
@@ -26,7 +26,7 @@ namespace StyleCop.Analyzers.LayoutRules
     internal class SA1500CodeFixProvider : CodeFixProvider
     {
         /// <inheritdoc/>
-        public override ImmutableArray<string> FixableDiagnosticIds { get; } = 
+        public override ImmutableArray<string> FixableDiagnosticIds { get; } =
             ImmutableArray.Create(SA1500CurlyBracketsForMultiLineStatementsMustNotShareLine.DiagnosticId);
 
         /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500CodeFixProvider.cs
```diff
@@ -229,28 +240,39 @@ private static bool ContainsStartOfLine(SyntaxToken token, int startLine)
...
-                    return replacementToken;
+                    return null;
                 }
 
-                return base.VisitToken(token);
+                SyntaxNode syntaxRoot = await document.GetSyntaxRootAsync().ConfigureAwait(false);
+
+                var tokens = diagnostics
+                    .Select(diagnostic => syntaxRoot.FindToken(diagnostic.Location.SourceSpan.Start))
+                    .OrderBy(token => token.SpanStart);
```
Maxwe11: I'd like to see `List<T>` here instead of a bunch of iterators, lambdas, etc.
We already have a bunch of `List<T>` [with cached lambda for sorting](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/9702db1a9ac1934bf8431e4c6850173dfb4c34c8/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1626CodeFixProvider.cs#L93)

vweijsters: I don't agree with that. The readability of the current construction is better in my opinion and this is not performance critical code, it will be called at most once per document.

:question: @sharwell @pdelvo how do you guys think about this?

sharwell: The current construction is fine in this context. I would prefer to see `ToImmutableArray` included in the assignment to `tokens` instead of at the usage location.

vweijsters:
Updated

sharwell:
:+1:

PR #1578, opened 2015-09-29
Made SA1508 implementation more robust

### LINKED ISSUE #1534: SA1508 can throw ArgumentOutOfRangeException
Previously we had a report of this in #938. Today I received a screenshot from another developer who saw this:

image

Still no known way to reproduce...

vweijsters:
Fixes #1534


sharwell:
Thanks for figuring this one out :+1:

PR #1580, opened 2015-09-30
SA1401 now ignores static readonly fields

### LINKED ISSUE #1320: Public Static Readonly Fields Trigger SA1401
```
public static readonly MyClass Empty = new MyClass();
```

SA1401 is triggered telling me that the field must be private.  Recommend ignoring the rule for public static readonly fields, as this is the .NET paradigm for creating empty objects.

vweijsters:
Fixes #1320

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1401FieldsMustBePrivate.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1401FieldsMustBePrivate.cs
@@ -3,6 +3,7 @@
 
 namespace StyleCop.Analyzers.MaintainabilityRules
 {
+    using System;
     using System.Collections.Immutable;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1401FieldsMustBePrivate.cs
```diff
@@ -49,14 +50,15 @@ public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
...
 
-        private void AnalyzeField(SymbolAnalysisContext symbolAnalysisContext)
+        private static void AnalyzeField(SymbolAnalysisContext symbolAnalysisContext)
         {
             var fieldDeclarationSyntax = (IFieldSymbol)symbolAnalysisContext.Symbol;
-            if (!this.IsFieldPrivate(fieldDeclarationSyntax) &&
-                this.IsParentAClass(fieldDeclarationSyntax) &&
+            if (!IsFieldPrivate(fieldDeclarationSyntax) &&
+                !IsStaticReadonly(fieldDeclarationSyntax) &&
+                IsParentAClass(fieldDeclarationSyntax) &&
```
Maxwe11: Why we need semantic model for this rule?

Maxwe11:
LGTM

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1401FieldsMustBePrivate.cs
vweijsters: I have no clue, I didn't make the initial implementation, just added the static readonly stuff.

PR #1581, opened 2015-09-30
Added custom build targets to nuget package to simplify setting item group for settings file

oatkins:
Added custom .targets and .props file with code to make it easy to set the item group for stylecop.json to StyleCopSettings. Include StyleCopSettings in AdditionalFileItemNames group.

I wouldn't be surprised if you've though of this already... and have a good reason why we shouldn't include it. But maybe it's just that nobody got around to it ;-)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/build/StyleCop.Analyzers.props
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/build/StyleCop.Analyzers.props
@@ -1,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <AdditionalFileItemNames>$(AdditionalFileItemNames);StyleCopSettings</AdditionalFileItemNames>
+  </PropertyGroup>
+
+</Project>
--- a/StyleCop.Analyzers/StyleCop.Analyzers/build/StyleCop.Analyzers.props
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/build/StyleCop.Analyzers.props
@@ -1,8 +1,0 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
-<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
-  <PropertyGroup>
-    <AdditionalFileItemNames>$(AdditionalFileItemNames);StyleCopSettings</AdditionalFileItemNames>
-  </PropertyGroup>
-
-</Project>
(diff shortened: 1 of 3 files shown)


sharwell:
The primary reason this wasn't addressed yet is no one took the time to do it.

That said, I do have a few concerns and I'm not sure what the best approach is.

1. This approach only works for users with standard project files. DNX users working with the new file-system projects would need a different solution.
2. This change is a workaround for an ongoing Roslyn limitation; see dotnet/roslyn#4655, dotnet/roslyn#5096, dotnet/roslyn#5104. Depending on the final resolution of these issues, our approach may change. If we implement the solution proposed in this pull request, users may be forced to change the way **stylecop.json** is added a second time if we adopt some other solution provided by the infrastructure.

If we could eliminate the requirement that users manually set the build type of **stylecop.json** to `StyleCopSettings`, we would somewhat address the second concern by automating the first change. For example, we could have a target automatically added to the build which identifies any files named **stylecop.json** with the build action `None` and automatically adds them to the set of additional files.

StyleCop.Analyzers/StyleCop.Analyzers/build/StyleCop.Analyzers.props
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup Condition="'$(BuildingInsideVisualStudio)' == 'true'">
```
sharwell: This would need to occur unconditionally, since command line builds still need to be aware of the **stylecop.json** file in order to produce accurate results.

oatkins:
@sharwell Thanks. Good point about the condition being unwanted.

> DNX users... would need a different solution.

Yup, I saw that discussion. It didn't seem to negate the value of making this work for MSBuild/VS projects though.

> This change is a workaround for an ongoing Roslyn limitation.

Maybe so, though my suspicion is that they *intended* for items to be added to the AdditionalFiles group indirectly, and that this is the reason why "AdditionalFiles" doesn't show up on the VS menu.

> We could have a target automatically added to the build which identifies any files named stylecop.json with the build action  None  and automatically adds them to the set of additional files.

It would be really nice if users could get settings working without having to make changes to the project file directly. The approach you describe would achieve that too. I'd be happy to change the diff to make it work that way instead if you think that's better.

sharwell:
So it seems we can use the following **.targets** file and omit the **.props** file completely. This implementation would not require users to perform any additional steps to enable the **stylecop.json** settings file after executing the code fix which adds it to a project (aside from *possibly* needing to reload the project; I haven't tested this specific case).

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Target Name="InjectStyleCopSettings"
          BeforeTargets="CoreCompile">
    <ItemGroup>
      <StyleCopSettings Include="@(None)"
                        Condition="'%(Filename)%(Extension)' == 'stylecop.json'"/>
      <AdditionalFiles Include="%(StyleCopSettings.Identity)" />
    </ItemGroup>
  </Target>
</Project>
```

sharwell:
:+1:

PR #1582, opened 2015-09-30
Initial work on a fix all provider for renames

pdelvo:
With this PR I implement a RenameFixAllProvider that can batch fix rename changes. It is more robust compared to the batch fixer because it only applies changes that can be applied at the same time without effecting each other. because of limitations in roslyns renamer it cant detect conflicts in renames itself but it can prevent conflicts between all rename tasks.

I need to write some unit tests that test the new batch renaming. Also I have to write some more code to prevent some more renames in partial types or derived types. Also the renamer can be used in more places (probably all SA13XX rules). Currently this is used for SA1300, SA1303, SA1304, SA1307, SA1311, SA1306, SA1312 and SA1313. This is currently approx. 10x faster than the batch fixer.

This also fixed that namespace renames were not applied globally. A test for this is also missing right now.

 This is still a work in progress but I want to get some early feedback on this.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/RenameHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/RenameHelper.cs
@@ -3,11 +3,17 @@
 
 namespace StyleCop.Analyzers.Helpers
 {
+    using System;
+    using System.Collections.Generic;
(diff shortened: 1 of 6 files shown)


pdelvo:
I'm closing this for now. This turns out to be a lot more complicated then I originally thought.

PR #1592, opened 2015-10-03
Fix directive handling in ElementOrderCodeFixProvider

### LINKED ISSUE #1588: SA1201 code fix does not respect conditional compilation statements
Consider the following piece of code:

```
namespace ClassLibrary1
{
    public class Class1
    {
        public void Method1()
        {
        }

#if DEBUG
        public string Property1 { get; }
#endif
    }
}
```

This will trigger:
```
Warning	SA1201	A property should not follow a method.	ClassLibrary1
```

Choosing `Fix element order` for `Property1` will result in the following code:

```
namespace ClassLibrary1
{
    public class Class1
    {
#if DEBUG
        public string Property1 { get; }

        public void Method1()
        {
        }
#endif
    }
}
```

i.e., `Method1` has been moved into the `#if DEBUG` code block - changing the semantics of the code.

Noryoko:
- [x] Add regression tests for SA1201
- [ ] Fixes #1588

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1201UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1201UnitTests.cs
@@ -306,6 +306,44 @@ public struct FooStruct { }
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// This is a regression test for DotNetAnalyzers/StyleCopAnalyzers#1588:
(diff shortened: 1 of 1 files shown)


Maxwe11:
I suspect with this change we still offer code fix which will do nothing or I'm wrong?

pdelvo:
I don't think this is sufficient. E.g.:
```csharp
namespace ClassLibrary1
{
    public class Class1
    {
        public void Method1()
        {
        }

#if true
        public string Property1 { get; }
        public string Property2 { get; }
        public string Property3 { get; }
#endif

        private int field1;
    }
}
```

We can't move Property2 in this case.

pdelvo:
@Noryoko Do you still plan to work on this or can we close this for now?

sharwell:
Closing this since we haven't heard back from the author and I don't have permissions to apply changes to the working branch Noryoko:fix-1588.

PR #1593, opened 2015-10-04
Fix escaped quotes in XML attributes

### LINKED ISSUE #1590: XmlSyntaxFactory.TextAttribute does not escape attribute values correctly
The helper method `XmlSyntaxFactory.TextAttribute` allows a `value` to contain characters which are not valid unescaped in an XML attribute.

| `name` | `value` | Expected | Actual |
| --- | --- | --- | --- |
| `name` | `value` | `name="value"` | `name="value"` |
| `name` | `"value"` | `name="&quot;value&quot;"` | `name=""value""` :exclamation: |
| `name` | `value&2` | `name="value&amp;2"` | `name="value&amp;2"` |

sharwell:
Fixes #1590


vweijsters:
:+1: Looks good to me

PR #1598, opened 2015-10-05
Implement SA1131 diagnostic + code fix + fix all provider + tests

### LINKED ISSUE #863: Implement SA1131 (UseReadableConditions)

pdelvo:
Fixes #863.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
@@ -1,0 +1,238 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.Test.ReadabilityRules
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Analyzers.ReadabilityRules;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using TestHelper;
+    using Xunit;
+
+    public class SA1131UnitTests : CodeFixVerifier
+    {
+        [Fact]
+        public async Task TestYodaComparismAsync()
+        {
+            var testCode = @"
+using System;
+public class TypeName
+{
+    public void Test()
+    {
+        int i = 5;
+        const int j = 6;
+        if (j == i) { }
+    }
+}";
+            var fixedCode = @"
+using System;
+public class TypeName
+{
+    public void Test()
+    {
+        int i = 5;
+        const int j = 6;
+        if (i == j) { }
+    }
+}";
+
+            var expected = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(9, 13),
+            };
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestYodaComparismAsAnArgumentAsync()
+        {
+            var testCode = @"
+using System;
+public class TypeName
+{
+    public void Test()
+    {
+        int i = 5;
+        const int j = 6;
+        Test(j == i);
+    }
+    public void Test(bool arg) { }
+}";
+            var fixedCode = @"
+using System;
+public class TypeName
+{
+    public void Test()
+    {
+        int i = 5;
+        const int j = 6;
+        Test(i == j);
+    }
+    public void Test(bool arg) { }
+}";
+
+            var expected = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(9, 14),
+            };
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestYodaComparismOutsideIfAsync()
+        {
+            var testCode = @"
+using System;
+public class TypeName
+{
+    public void Test()
+    {
+        int i = 5;
+        const int j = 6;
+        bool b = j == i;
+    }
+}";
+            var fixedCode = @"
+using System;
+public class TypeName
+{
+    public void Test()
+    {
+        int i = 5;
+        const int j = 6;
+        bool b = i == j;
+    }
+}";
+            var expected = new[]
+            {
+                this.CSharpDiagnostic().WithLocation(9, 18),
+            };
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/ReadabilityResources.resx
```diff
@@ -432,11 +432,14 @@
   <data name="SA1130Title" xml:space="preserve">
     <value>Use lambda syntax</value>
   </data>
+  <data name="SA1131CodeFix" xml:space="preserve">
+    <value>Switch arguments.</value>
```
sharwell: :bulb: ~~arguments~~ operands

sharwell:
This implementation doesn't quite follow the [documentation for this rule](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1131.md) (which happens to be more detailed than you see for many other rules).

The main differences are:

1. It only considers "Any expression which evaluates to a constant value at compile time" to be a literal or constant value. My biggest questions relate to the handling of `null`, `default(T)`, and `static readonly` fields.
2. It only looks at operator `==`. We certainly need to include `!=` as well. For the relational operators which are listed, we either need to include them or update the documentation.

pdelvo:
`null` and `default(T)` are constants. Everything but the two last entries could be removed from that list and it would still be correct.

sharwell:
What does `GetConstantValue` return for `null`?

pdelvo:
HasValue is true and Value is null.

sharwell:
What does it return for `default(T)`, where `T` is a type defined in the assembly being compiled?

sharwell:
> Everything but the two last entries could be removed from that list and it would still be correct.

I believe you are correct, but I included them for clarity. The reader may or may not realize they are equivalent. :smile:

pdelvo:
For a class its null. It doesnt work for structs. So I have to look for default(..).

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1131UseReadableConditions.cs
```diff
@@ -33,7 +36,24 @@ internal class SA1131UseReadableConditions : DiagnosticAnalyzer
         /// <inheritdoc/>
         public override void Initialize(AnalysisContext context)
         {
-            // TODO: Implement analysis
+            context.RegisterCompilationStartAction(HandleCompilationStart);
+        }
+
+        private static void HandleCompilationStart(CompilationStartAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionHonorExclusions(HandleBinaryExpression, SyntaxKind.EqualsExpression);
```
Maxwe11: Why only `EqualsExpression` ?

Maxwe11:
What about these
> - A string literal
- null
- default(T) (for any type T)
- A reference to a static readonly field, such as IntPtr.Zero
test cases?

pdelvo:
I updated this PR

Maxwe11:
:+1:

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1131CodeFixProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                var syntaxRoot = await document.GetSyntaxRootAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
+
+                var nodes = diagnostics.Select(diagnostic => (BinaryExpressionSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true));
+
+                return syntaxRoot.ReplaceNodes(nodes, (originalNode, rewrittenNode) => TransformExpression(rewrittenNode));
+            }
+        }
+    }
+}
```
Maxwe11: nit: new line missing

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
```diff
@@ -0,0 +1,544 @@
...
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using TestHelper;
+    using Xunit;
+
+    public class SA1131UnitTests : CodeFixVerifier
+    {
+        [Theory]
+        [InlineData("==", "==")]
+        [InlineData("!=", "!=")]
+        [InlineData(">=", "<=")]
```
sharwell: :exclamation: This operator pairing is incorrect. It should be:

| Old | New |
| --- | --- |
| `==` | `==` |
| `!=` | `!=` |
| `>=` | `<` :exclamation:  |
| `<=` | `>` :exclamation:  |
| `>` | `<=` :exclamation:  |
| `<` | `>=` :exclamation:  |

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
```diff
@@ -0,0 +1,544 @@
...
+            };
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [InlineData("==", "==")]
+        [InlineData("!=", "!=")]
+        [InlineData(">=", "<=")]
```
sharwell: :exclamation: This operator pairing is incorrect.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
```diff
@@ -0,0 +1,544 @@
...
```
sharwell: :exclamation: This operator pairing is incorrect.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
```diff
@@ -0,0 +1,544 @@
...
```
sharwell: :exclamation: This operator pairing is incorrect.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
```diff
@@ -0,0 +1,544 @@
...
```
sharwell: :exclamation: This operator pairing is incorrect.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1131UnitTests.cs
```diff
@@ -0,0 +1,544 @@
...
+    }}
+}}";
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+
+        [Theory]
+        [InlineData("==", "==")]
+        [InlineData("!=", "!=")]
+        [InlineData(">=", "<=")]
```
sharwell: :exclamation: This operator pairing is incorrect.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/ReadabilityResources.resx
```diff
@@ -432,11 +432,14 @@
   <data name="SA1130Title" xml:space="preserve">
     <value>Use lambda syntax</value>
   </data>
+  <data name="SA1131CodeFix" xml:space="preserve">
+    <value>Switch operants.</value>
```
sharwell: :bulb: operands is misspelled.

PR #1599, opened 2015-10-05
SA1516 no longer counts attributes when determining if a declaration …

### LINKED ISSUE #1595: SA1516 reports violation between fields if the first one has an attribute
I don't think the following code should cause an SA1516 violation.

```csharp
public class SomeClass
{
  [CanBeNull]
  private Foo foo;
  private bool bar;
}
```

vweijsters:
…is multiline

Fixes #1595


sharwell:
:+1:

PR #1605, opened 2015-10-07
SA1513 no longer reported for closing brace in first part of conditio…

### LINKED ISSUE #1583: SA1513 reported when I wouldn't expect it
I'm getting a violation of SA1513 (ClosingCurlyBracketMustBeFollowedByBlankLine) reported when I'm not expecting it. It's possible it's what you guys expect, but on the off chance it's a bug I thought I'd report it :)

``` CSharp
var target = isIFrame
    ? new Dictionary<string, string>
        {
            { "target", "_parent" }
        } // <-- SA1513 Violation reported here
    : new Dictionary<string, string>();
```

**Note**: The following code is not raising the issue, but personally I think it's an unnecessary new line that isn't helping readability.

``` CSharp
var target = isIFrame
    ? new Dictionary<string, string>
        {
            { "target", "_parent" }
        }

    : new Dictionary<string, string>();
```

 What do you think?

vweijsters:
…nal expression

Fixes #1583

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1513ClosingCurlyBracketMustBeFollowedByBlankLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1513ClosingCurlyBracketMustBeFollowedByBlankLine.cs
@@ -249,6 +249,13 @@ namespace StyleCop.Analyzers.LayoutRules
                         return;
                     }
 
+                    if (nextToken.IsKind(SyntaxKind.ColonToken)
+                        && IsPartOf<ConditionalExpressionSyntax>(token))
+                    {
+                        // the close brace is in the first part of a conditional expression.
+                        return;
+                    }
+
                     if (nextToken.IsKind(SyntaxKind.AddKeyword)
                         || nextToken.IsKind(SyntaxKind.RemoveKeyword)
                         || nextToken.IsKind(SyntaxKind.GetKeyword)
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1513ClosingCurlyBracketMustBeFollowedByBlankLine.cs
```diff
@@ -249,6 +249,13 @@ private void AnalyzeCloseBrace(SyntaxToken token)
                         return;
                     }
 
+                    if (nextToken.IsKind(SyntaxKind.ColonToken)
+                        && IsPartOf<ConditionalExpressionSyntax>(token))
```
sharwell: I'm trying to think of a case where `}` is followed by `:` but is not part of a conditional expression.

Places a colon can appear:

* Conditional expressions
* Labels and switch cases (can't be preceded by `}`)
* Named arguments (can't be preceded by `}`)
* The target of an attribute (can't be preceded by `}`)
* Base type list (can't be preceded by `}`)
* Initializer expression (can't be preceded by `}`)

Are there any others?

sharwell: :memo: I simplified this condition to just the following during the merge:

```csharp
if (nextToken.IsKind(SyntaxKind.ColonToken))
```

Even if there was another case, whatever that case is would also be exempt from this rule.

PR #1609, opened 2015-10-08
Fix that SA1003 does not allow postfix unary expressions as last elem…

### LINKED ISSUE #1600: SA1003 violation if using postfix increment in object initializer
The following code triggers a violation but didn't in StyleCop 'classic'.

```csharp
new Foo
{
  Foo = "foo",
  Bar= i++
};
```

If you use the code fix the generated code causes an SA1500 violation.

pdelvo:
…ents in an anonymous object creation or initializers.

Fixes #1600.


vweijsters:
:+1: Looks good to me

sharwell:
:+1:

PR #1612, opened 2015-10-09
Add a string builder cache

pdelvo:
I copied over the pool implementation from roslyn and used the StringBuilderPool in all places were we used a string builder instance before. We still allocate new instances in our tests but I don't think this matters too much.


pdelvo:
I have to fix some more style violations. I have some issues with visual studio 2015 update 1 ctp not reporting them to me.

pdelvo:
I fixed this PR

PR #1616, opened 2015-10-11
SA1313 codefix will remove underscore if necessary

### LINKED ISSUE #1604: SA1313 code fix does not remove _
Consider this code:
```csharp
public Token(string _tokenText)
        : this(_tokenText, ScannerState.ParseValue, false)
    {
    }
```
SA1313 is correctly reported for _tokenText, but the code fix is not removing the underscore.

vweijsters:
Fixes #1604 

As a side effect, leading underscores are also removed for SA1312 in the code fix.
As SA1306 is not reported for fields starting with an underscore, the change does not affect the SA1306 code fix.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1312UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1312UnitTests.cs
@@ -148,12 +148,19 @@ public class TypeName
     }
 }";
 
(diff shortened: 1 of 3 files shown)


sharwell:
:question: Can you add tests for the name `_` and `__` in each of the analyzers which use this code fix?

vweijsters:
Testcases have been added SA1312 and SA1313. No testcases where added for SA1306, as this one will ignore underscores. Also added a bug fix

sharwell:
:+1:

PR #1625, opened 2015-10-14
Removed unnecessary diagnostic id checks

vweijsters:
- Removed the `.Where` and `if (...)` conditions that check to see if a diagnostic ID is fixable. See discussion in #1441 
- Fixed failing SA1513 unit test 
- Modified `CodeFixVerifier` to only present supported diagnostics to a code fix provider

This is part 1 of the breakup of #1452


Maxwe11:
:shipit:

sharwell:
Why are we not seeing an AppVeyor build failure due to the SA1513 test?

sharwell:
Ah, looks like that test only runs the code fix, but doesn't have the assertions for diagnostics reported before/after the fix.

sharwell:
:+1:

PR #1630, opened 2015-10-14
Removed FixableDiagnostics fields

vweijsters:
- Removed all `FixableDiagnostics` fields and assigned the immutable array directly to the `FixableDiagnosticsId` override

This is part 2 of the breakup of #1452


sharwell:
:+1:

PR #1631, opened 2015-10-14
Reformatted all RegisterCodeFix calls

vweijsters:
- Reformatted all `RegisterCodeFix` calls to use the same layout
- renamed all `t` or `token` to the proper `cancellationToken`
- removed `equivalenceKey:` as it was unnecessary
- added missing equivalence key for `SA1302CodeFixProvider`

This is part 3 of the breakup of #1452


sharwell:
I like this change. I need to make another pass over it just to make sure nothing changed aside from formatting, but I don't expect to find any problems based on what I've seen so far.

PR #1649, opened 2015-10-18
Implemented SX1101 (incl. tests, codefix, and docs)

### LINKED ISSUE #1601: New rule proposal: SX1101 DoNotPrefixLocalMembersWithThis
**Category:** Readability
**Title:** Do not prefix local members with this
**Description:** A violation of this rule occurs whenever the code contains a call to an instance member of the local class or a base class which is prefixed with `this.`.
**Rationale:** This rule as an alternative rule for SA1101 (PrefixLocalCallsWithThis). This rule is most likely to be used in conjunction with SX1309 (FieldNamesMustBeginWithUnderscore) (and/or SX1309S (StaticFieldNamesMustBeginWithUnderscore)). A typical usage scenario when using members starting with underscores is to not use the `this.` prefix. I think we should offer our users the possibility to enforce this. This rule will be disabled by default.

vweijsters:
Fixes #1601

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SX1101DoNotPrefixLocalMembersWithThis.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SX1101DoNotPrefixLocalMembersWithThis.cs
@@ -3,9 +3,13 @@
 
 namespace StyleCop.Analyzers.ReadabilityRules
 {
+    using System;
+    using System.Collections.Generic;
     using System.Collections.Immutable;
+    using System.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
 
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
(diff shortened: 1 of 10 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SX1101DoNotPrefixLocalMembersWithThis.cs
```diff
@@ -0,0 +1,49 @@
...
+        private static void HandleCompilationStart(CompilationStartAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionHonorExclusions(HandleThisExpression, SyntaxKind.ThisExpression);
+        }
+
+        private static void HandleThisExpression(SyntaxNodeAnalysisContext context)
+        {
+            if (context.Node.Parent.IsKind(SyntaxKind.SimpleMemberAccessExpression))
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, context.Node.GetLocation()));
```
pdelvo: What if this is necessary because e.g. the name of a field is the same as the name of a parameter?

StyleCopAnalyzers.sln
```diff
@@ -217,6 +217,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "documentation", "documentat
 		documentation\SA1650.md = documentation\SA1650.md
 		documentation\SA1651.md = documentation\SA1651.md
 		documentation\SA1652.md = documentation\SA1652.md
+		SX1101.md = SX1101.md
```
pdelvo: The file should be in the documentation directory

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SX1101DoNotPrefixLocalMembersWithThis.cs
vweijsters: Good point, I'll address that

vweijsters:
Updated after CR

sharwell:
:+1:

PR #1651, opened 2015-10-19
Small refactorings

NikolayIT:
- Replaced two calls to Count() calls (with O(n)) with faster calls to .Length and .Count (O(1))
- Few typos fixed
- Removed code redundancies


vweijsters:
:+1: Looks good to me

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1407SA1408FixAllProvider.cs
```diff
@@ -38,7 +38,7 @@ internal sealed class SA1407SA1408FixAllProvider : DocumentBasedFixAllProvider
                 nodes.Add(node);
             }
 
-            return root.ReplaceNodes(nodes, (originalNode, rewrittenNode) => AddParentheses(originalNode, rewrittenNode));
+            return root.ReplaceNodes(nodes, AddParentheses);
```
pdelvo: These changes are actually hurting performance because this will allocate a delegate every time. The lambda expression is cached.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -284,7 +284,7 @@ private static string FormatDiagnostics(ImmutableArray<DiagnosticAnalyzer> analy
             {
                 var diagnosticsId = diagnostics[i].Id;
 
-                builder.AppendLine("// " + diagnostics[i].ToString());
+                builder.AppendLine("// " + diagnostics[i]);
```
sharwell: :bulb:

```
builder.Append("// ").AppendLine(diagnostics[i]);
```

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/DiagnosticVerifier.cs
```diff
@@ -300,7 +300,7 @@ private static string FormatDiagnostics(ImmutableArray<DiagnosticAnalyzer> analy
                     {
                         Assert.True(
                             location.IsInSource,
-                            string.Format("Test base does not currently handle diagnostics in metadata locations. Diagnostic in metadata:\r\n", diagnostics[i]));
+                            string.Format("Test base does not currently handle diagnostics in metadata locations. Diagnostic in metadata:{0}\r\n", diagnostics[i]));
```
sharwell: :thought_balloon: It looks like the diagnostic was intended to be placed after the newline.

PR #1653, opened 2015-10-19
Allow directive trivia in parameter list

### LINKED ISSUE #1623: SA1009,1111,1114,1115 trigger when using compiler directives in method signatures
Unfortunately due to Microsoft renaming a bunch of interfaces, we had to use compiler directives to pick the right interface for the version being compiled.

It would be great if Stylecop would ignore compiler directive lines and not treat them as whitespace.

````
#pragma warning disable SA1111, SA1115, SA1114, SA1009
        public EventNotificationStatus ProcessEvent(
#if TFS20151
            IVssRequestContext requestContext,
#else
            TeamFoundationRequestContext context,
#endif
            NotificationType notificationType,
            object notificationEventArgs,
            out int statusCode,
            out string statusMessage,
            out ExceptionPropertyCollection properties)
#pragma warning restore SA1111, SA1115, SA1114, SA1009
        {
````

And:

```
        private Uri GetCollectionUriFromContext(
+#if TFS20151
+            IVssRequestContext requestContext
+#else
+            TeamFoundationRequestContext requestContext
+#endif
+            )
         {
```

vweijsters:
Fixes #1623

SA1114 and SA1115 have been adapted for this.
I purposely implemented the smallest fix possible for this, as I'm not certain that we want the same changes for all cases handled by the analyzers. Both analyzers now only support directive trivia in parameters lists of all types of method declarations.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1114UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1114UnitTests.cs
@@ -1067,6 +1067,35 @@ public class SomeOtherClass
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
         }
 
+        /// <summary>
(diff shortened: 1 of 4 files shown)


sharwell:
:+1:

PR #1656, opened 2015-10-19
Refactor: cleanup SupportedDiagnostics declarations

vweijsters:
Cleaned up all `SupportedDiagnostics` declarations inline with earlier refactorings of code fix providers.


sharwell:
:+1:

PR #1658, opened 2015-10-20
SA1113 reporting failure (proof of concept)

sharwell:
I've been observing cases where SA1113 is reported in the editor (for IntelliSense purposes), but not being reported when compiling with MSBuild. This pull request is an attempt to reproduce this reporting failure in a reference environment (the AppVeyor build environment).


tmeschter:
@sharwell Any chance the analyzer is failing to load with an AD0001 diagnostic? You might have to bump up the MSBuild logging verbosity and then search the build log to find it.

Issue #1357, opened 2015-09-01
File headers: allow pattern matching or multiple allowable headers to support headers that must include a date


oatkins:
I agree with @sharwell that it's a pain to have a file header that includes a date. However, this might not always be avoidable. On one project, the header includes the _year_ when the file was last updated.

It would be helpful if there was a way to allow, say, "Copyright (C) Flubbersoft, {year}" where **year** is defined as

```
\d{4}
```

(regular expression) or even

```
{ "2014", "2015" }
```

(array of permissible values)

I can make my own analyzer for this case if it's rare, but would be interested to know whether it would be helpful to anyone else too.

sharwell:
So what you're saying is, if you update all the headers to say 2015, it will still be correct?

Because that's exactly what I did in my work project to handle exactly this case. :smile:

oatkins:
We were asked not to update the date unless the file had been changed.

oatkins:
Of course, if you *do* update the date, then the file has definitely been changed... but I don't think that's what was meant ;-)

mreu:
+1 Same here.

bktechy:
We have the same issue.  We don't update the date unless the file has been changed.  With a large solution, changing all file headers to a new date is cumbersome.

Having a way to support this use case would be very helpful.

sharwell:
At this point, the project already supports two specific and particularly practical use cases:

1. Cases like the .NET Foundation uses, which minimizes the chances that a project header ever needs to be changed (no date included at all, but instead references another file containing the information).
2. The strongest possible copyright protection under US copyright law, which is to use the year in which the project started as the copyright date and not change the year for future files (new files get back-dated).

I've seen other strategies employed, but they all seem to lose sight of the actual goals. When 95 years rolls around, one of three things will inevitably happen:

1. You don't care about the code enough to try and protect it anymore
2. The code is so old no one cares to copy it anymore aside from fair use for historical preservation purposes
3. We will all be dead so the whole thing is moot

My recommendation is to use the inflexibility of this tool as rationale to change the policy as it applies to your project.

Nihlus:
Aside from dates, a pattern matching feature would be very useful in terms of headers where the file author is included - a common pattern in open-source collaborations. As it is, having two files with the same header where the only difference is the author (or list of authors) is not possible. A way to have wildcards or list value matching would solve this issue.

sharwell:
[...]

PR #1661, opened 2015-10-20
Use legally-correct copyright notices.

Travis-Snoozy:
The existing copyright support uses the (legally-unrecognized) (C), and does not support year lists due to the literal-matching nature of copyrightText.

This change fixes the above problems by adding a new {copyright} variable for use with copyrightText. The default value for copyrightText is also updated to use {copyright} instead of the hard-coded "Copyright (C)" text.

{copyright} provides the following features:
- When suggesting a fix, automatically expands to "Copyright © " followed by the current year.
- When checking copyright headers, {copyright} matches "Copyright © " followed by a comma-separated list of four-digit years, and/or year-ranges (two four-digit years separated with a hyphen).

This change adds new unit tests to verify this new functionality, and updates the existing tests to work correctly with the new behaviors.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/FileHeaderTestBase.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/FileHeaderTestBase.cs
@@ -22,7 +22,7 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
   ""settings"": {
     ""documentationRules"": {
       ""companyName"": ""FooCorp"",
-      ""copyrightText"": ""Copyright (c) FooCorp. All rights reserved.""
+      ""copyrightText"": ""{copyright} {companyName}. All rights reserved.""
     }
   }
 }
@@ -53,7 +53,7 @@ namespace Bar
         public async Task TestValidFileHeaderAsync()
         {
             var testCode = @"// <copyright file=""Test0.cs"" company=""FooCorp"">
-//   Copyright (c) FooCorp. All rights reserved.
+//   Copyright © 2015 FooCorp. All rights reserved.
 // </copyright>
 // <summary>This is a test file.</summary>
 
@@ -74,7 +74,7 @@ namespace Bar
         {
             var testCode = @"//----------------------------------------
 // <copyright file=""Test0.cs"" company=""FooCorp"">
-//   Copyright (c) FooCorp. All rights reserved.
+//   Copyright © 2015 FooCorp. All rights reserved.
 // </copyright>
 // <summary>This is a test file.</summary>
 //----------------------------------------
@@ -82,6 +82,79 @@ namespace Bar
 namespace Bar
 {
 }
+";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the literal use of {copyright} in a header does not match the configured token {copyright}.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestLiteralCopyrightAsync()
+        {
+            var testCode = @"// <copyright file=""Test0.cs"" company=""FooCorp"">
+//   {copyright} FooCorp. All rights reserved.
+// </copyright>
+// <summary>This is a test file.</summary>
+
+namespace Bar
+{
+}
+";
+
+            var expectedDiagnostic = this.CSharpDiagnostic(FileHeaderAnalyzers.SA1636Descriptor).WithLocation(1, 4);
+            await this.VerifyCSharpDiagnosticAsync(testCode, expectedDiagnostic, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that {copyright} correctly requires "Copyright ©" followed by at least a 4-digit year.
+        /// </summary>
+        /// <param name="copyright">The (malformed) string to use in the {copyright} position.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Theory]
+        [InlineData("© 1000")]
+        [InlineData("Copyright 1000")]
+        [InlineData("Copyright ©")]
+        [InlineData("Copyright © 93")]
+        [InlineData("Copyright © 1000-93")]
+        [InlineData("Copyright © 1000, 93")]
+        public async Task TestMalformedCopyrightAsync(string copyright)
+        {
+            var testCode = $@"// <copyright file=""Test0.cs"" company=""FooCorp"">
+//   {copyright} FooCorp. All rights reserved.
+// </copyright>
+// <summary>This is a test file.</summary>
+
+namespace Bar
+{{
+}}
+";
+
+            var expectedDiagnostic = this.CSharpDiagnostic(FileHeaderAnalyzers.SA1636Descriptor).WithLocation(1, 4);
+            await this.VerifyCSharpDiagnosticAsync(testCode, expectedDiagnostic, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that {copyright} correctly accepts multiple years (lists, spans, and mixtures of the two).
+        /// </summary>
+        /// <param name="copyright">The string to use in the {copyright} position.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Theory]
+        [InlineData("Copyright © 1000, 1003")]
+        [InlineData("Copyright © 1000-1003")]
+        [InlineData("Copyright © 1000-1003, 1005, 1007-1010")]
+        public async Task TestMultipleCopyrightYearsAsync(string copyright)
+        {
+            var testCode = $@"// <copyright file=""Test0.cs"" company=""FooCorp"">
+//   {copyright} FooCorp. All rights reserved.
+// </copyright>
+// <summary>This is a test file.</summary>
+
+namespace Bar
+{{
+}}
 ";
 
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1633UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1633UnitTests.cs
@@ -3,6 +3,7 @@
 
 namespace StyleCop.Analyzers.Test.DocumentationRules
 {
+    using System;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis.CodeFixes;
@@ -38,8 +39,8 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
         [Fact]
         public async Task TestValidFileHeaderNoContentAsync()
         {
-            var testCode = @"// <copyright file=""Test0.cs"" company=""FooCorp"">
-//   Copyright (c) FooCorp. All rights reserved.
+            var testCode = $@"// <copyright file=""Test0.cs"" company=""FooCorp"">
+//   Copyright © {DateTime.Now.Year} FooCorp. All rights reserved.
 // </copyright>
 ";
 
@@ -53,26 +54,26 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
         [Fact]
         public async Task TestValidMultilineCommentFileHeadersAsync()
         {
-            var testCode1 = @"/* <copyright file=""Test0.cs"" company=""FooCorp"">
-  Copyright (c) FooCorp. All rights reserved.
+            var testCode1 = $@"/* <copyright file=""Test0.cs"" company=""FooCorp"">
+  Copyright © {DateTime.Now.Year} FooCorp. All rights reserved.
 </copyright> */
 ";
 
-            var testCode2 = @"/*
+            var testCode2 = $@"/*
 <copyright file=""Test1.cs"" company=""FooCorp"">
-  Copyright (c) FooCorp. All rights reserved.
+  Copyright © {DateTime.Now.Year} FooCorp. All rights reserved.
 </copyright>
 */
 ";
 
-            var testCode3 = @"/*<copyright file=""Test2.cs"" company=""FooCorp"">
-  Copyright (c) FooCorp. All rights reserved.
+            var testCode3 = $@"/*<copyright file=""Test2.cs"" company=""FooCorp"">
+  Copyright © {DateTime.Now.Year} FooCorp. All rights reserved.
 </copyright>*/
 ";
 
-            var testCode4 = @"/*
+            var testCode4 = $@"/*
  * <copyright file=""Test3.cs"" company=""FooCorp"">
- *   Copyright (c) FooCorp. All rights reserved.
+ *   Copyright © {DateTime.Now.Year} FooCorp. All rights reserved.
  * </copyright>
 */
 ";
@@ -87,17 +88,17 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
         [Fact]
         public async Task TestValidFileHeaderWithDirectivesAsync()
         {
-            var testCode = @"#define MYDEFINE
+            var testCode = $@"#define MYDEFINE
 #if (IGNORE_FILE_HEADERS)
 #pragma warning disable SA1633
 #endif
 // <copyright file=""Test0.cs"" company=""FooCorp"">
-//   Copyright (c) FooCorp. All rights reserved.
+//   Copyright © {DateTime.Now.Year} FooCorp. All rights reserved.
 // </copyright>
 
 namespace Bar
-{
-}
+{{
+}}
 ";
 
             var expected = this.CSharpDiagnostic(FileHeaderAnalyzers.SA1633DescriptorMissing).WithLocation(1, 1);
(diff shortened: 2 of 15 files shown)


sharwell:
I would prefer the text `Copyright ©` appear as literal text in the `copyrightText` property of **stylecop.json**, and the new placeholder only apply to the year ranges.

sharwell:
:memo: Note that the default header's use of `Copyright (c)` complies with the recommended form of notice described under **Visually Perceptible Copies** in [Copyright Notice](http://www.copyright.gov/circs/circ03.pdf) by the US Copyright Office. It is the word 'Copyright' as opposed to the symbol '(c)' which meets this condition.

sharwell:
I've marked this as **needs discussion**. The primary new feature added by this pull request is support for year lists. This is related to #1357.

The main areas of concern I have with this proposal are the following:

1. The US Copyright Office does not suggest the use of or prescribe a form for copyright notices listing multiple years. This means the form used for multiple years is established on a per-company or per-legal-team basis, and thus only applies to a subset of our potential users which are still using headers with year(s) listed.
2. The copyright years cannot be validated for correctness by StyleCop Analyzers. Maintenance of the proposed form imposes the maximum burden for development teams, and I would prefer to discourage its use as much as possible, in as many places as possible. My hope is that as more projects adopt notices which intentionally omit the copyright year(s), the entire software industry will benefit.

The OSS license used for this repository of analyzers does permit users to create a new analyzer which specifically handles copyright headers in a custom format and use that analyzer within their project. As long as the diagnostic IDs for file header diagnostics are changed, the custom analyzer and StyleCop.Analyzers could be used together in the same repository.

Travis-Snoozy:
Yes, Copyright is valid, but (C) is meaningless; perpetuating its use is just silly, so fix it or drop it. Omission if a year is the primary flaw here, regardless. The implementation being tied to the placement of Copyright was intentional for multiple reasons; I'll elaborate further once I have a real keyboard.

http://www.copyright.gov/circs/circ01.pdf

Page 4, "Form of Notice for Visually Perceptible Copies" specifies the year as one of three critical points.

I'm perfectly happy to not handle multi-years, but ONE year is at minimum required.

--
Travis

sharwell:
> Yes, Copyright is valid, but (C) is meaningless; perpetuating its use is just silly, so fix it or drop it.

The current text is the closest we can get to © while only using characters which have the same binary representation in Windows-1252 and UTF-8. If any change is made to the default, I expect it to be the result of ongoing discussions in the .NET Foundation. Note that you can currently customize the `copyrightText` value in **stylecop.json**, and if you use © there it will enforce the use of that mark in source files as well.

> I'm perfectly happy to not handle multi-years, but ONE year is at minimum required.

We use this at my office by creating a variable which holds a fixed year (currently 2015).

```json
{
  "$schema": "https://raw.githubusercontent.com/DotNetAnalyzers/StyleCopAnalyzers/master/StyleCop.Analyzers/StyleCop.Analyzers/Settings/stylecop.schema.json",
  "settings": {
    "documentationRules": {
      "companyName": "Our Company",
      "copyrightText": "Copyright (c) {year} {companyName}. All Rights Reserved.",
      "xmlHeader": false,
      "variables": {
        "year": "2015"
      }
    }
  }
}
```

I think you can make a great case for expanding on this. A special variable (name TBD) which could be used in the `copyrightText` value with the following semantics:

* When suggesting a fix, automatically expands to the current year.
* When checking copyright headers, matches any four-digit year.

Travis-Snoozy:
Still not at a kb, so bear with me --

The placeholder needs to match a valid Copyright symbol as the prev. token. Thus 

{copyright} => ((Copyright)|©) [12]\d{3}  

or similar. Trying to date-match in isolation is an ultimately useless activity, as we can't tell a date from any other sort of 4-digit # -- it's the direct proximity to the magic Copyright word that gives this sequence of digits importance, and the Copyright word in turn needs those digits. I considered trying to do full regex support as the ultimate generic way to solve this, but that plan fell over when suggested fixes had to be considered.

Re. encoding, that shouldn't be a concern: either tag the necessary files with a UTF-8 "BOM", or set up the HTTP server to serve the appropriate file encoding in the headers, and recode as appropriate to the output file's encoding. That said, AFAIK, all .NET languages need to support Unicode source files anyway*. If it's a Super Big Deal, then just drop the (C) altogether. Being squeamish about Unicode vs. legacy text seems oddly quaint, though. Very 90s. :)

Cheers,

--
Travis

* I've noticed the analyzers seemed to choke when I used © in the name of a C# variable, so I presume there are bugs hiding in the code. I'll probably poke at that a bit later.

vweijsters:
I think that the original test cases should not have been modified. Those were valid test cases and should not be modified for an optional feature.

vweijsters:
:question: What is the point of supporting comma-separated list of years or year ranges?

Travis-Snoozy:
50% sure I only touched tests that failed after my change (a shocking number did), but it's entirely possible some got dragged along for the ride. The refactoring was a rather boring autopilot haze.

Regardless, none of the tests' underlying flavor should have changed. If you have specific cases you'd like to try reverting, feel free to point them out.

--
Travis

sharwell:
> 50% sure I only touched tests that failed after my change (a shocking number did)...

This was caused by changing the default value of the `copyrightText` configuration property. Even if we update the code to support a special replacement variable with a "flexible" definition for the year, the default value for `copyrightText` should not be changed to use it. Once the default value is restored, the previously created tests should work again without changes.

GregReddick:
In addition, circular 3 gives more information about the proper form of copyright notices. http://www.copyright.gov/circs/circ03.pdf The (c) has no meaning. There was a court case where the C was enclosed in an octagon instead of a circle and it was ruled invalid.The year should be present. In the United States since 1 March 1989, not having a copyright notice does not mean that the work is uncopyrighted, however it does cause problems. (Not registering the copyright with the Library of Congress limits the damages that can be collected.)

Travis-Snoozy:
Now that I'm at a keyboard... wall of text engage!

> I would prefer the text Copyright © appear as literal text in the copyrightText property of stylecop.json, and the new placeholder only apply to the year ranges.

[...]

> I think you can make a great case for expanding on this. A special variable (name TBD) which could be used in the copyrightText value with the following semantics:
> * When suggesting a fix, automatically expands to the current year.
> * When checking copyright headers, matches any four-digit year.


From what I can tell, the year should come _immediately after_ either "Copyright", "Copr.", or "©". Thus, a generic year-only match is not a semantically equivalent substitute (although it would technically work, essentially you could put {year} tags anywhere, and it would not guarantee the enforcement of a full, legally recognized copyright statement).

A full regex/pattern system would be the logical extension as a fully-flexible and data-driven solution, but that would require a lot of plumbing, and lead to interesting requirements in terms of interacting with the suggested fixes code. It's a lot of work just to support a single templating scenario.

What I could do is put in another configuration knob that lets the customer chose an appropriate value (e.g., copyrightPhrase set by an enum like Full, Abbreviated, Symbol), and have that do the special expansion that checks for a word.

Sadly, this can't be an implicit rule that requires a four-digit year-like thing after any instance of "Copyright", "Copr.", or "©", because those words/symbols could potentially show up in the copyright header and not actually be the intended statement of copyright. For example: "This program is protected by U.S. Copyright law" would trigger improperly. The user has to be able to convey the explicit semantic intent of "this is a copyright notice," and that means, as far as I can tell, the whole notice needs to be handled as a single, explicitly-stated token.

> The year of first publication is intentionally omitted from the default notice[...]

[...]

> My hope is that as more projects adopt notices which intentionally omit the copyright year(s), the entire software industry will benefit.

[...]

> The US Copyright Office does not suggest the use of or prescribe a form for copyright notices listing multiple years.

[...]

> What is the point of supporting comma-separated list of years or year ranges? 

[Guidance given by the copyright office](http://www.copyright.gov/circs/circ01.pdf) specifies a single year after the "Copyright", "Copr.", or "©". This represents, at least in the US, the "best practice," and that is exactly what StyleCop is supposed to be helping us conform to. By the same reasoning, **I intend to remove the multi-year support from this change**.

>The existing copyright support uses the (legally-unrecognized) (C)[...]

[...]

>It is the word 'Copyright' as opposed to the symbol '(c)' which meets this condition.

[...]

>The (c) has no meaning.

Best practice would seem to be to not have (C) -- everyone seems to acknowledge it's not a legally recognized sequence. It's cruft. Kill the cruft. :)

>Even if we update the code to support a special replacement variable with a "flexible" definition for the year, the default value for copyrightText should not be changed to use it.

[...]

> If any change is made to the default, I expect it to be the result of ongoing discussions in the .NET Foundation.

[...]

>In the United States since 1 March 1989, not having a copyright notice does not mean that the work is uncopyrighted, however it does cause problems. (Not registering the copyright with the Library of Congress limits the damages that can be collected.)

Omitting copyright years may seem like a trivial choice, but as Greg points out, there are real legal ramifications of making this choice. Upon learning this, I would consider it to be not only incorrect, but professionally unethical to continue with a default that does not enforce a year-labeled copyright notice.

> We use this at my office by creating a variable which holds a fixed year (currently 2015).

This was the first option I reached for when I tried to get my codebase going... sadly, it leads to StyleCop errors in the source code unless all the files use the current year. Since it is improper and incorrect to change all the files to the latest year (especially if the code in the files hasn't changed), this sadly this is a non-solution to the bigger problem. Again, we need an unambiguous semantic declaration of "this is a copyright notice here," and the engine needs to be smart enough to treat it properly.

sharwell:
I'm going to go a bit out of order on the responses.

> This represents, at least in the US, the "best practice," and that is exactly what StyleCop is supposed to be helping us conform to.

The suggested guidance is the recommended practice to follow in order to maximize the ability to obtain statutory damages following an infringement. In my experience (and [IANAL](https://en.wikipedia.org/wiki/IANAL)), source code files tend to derive only marginal benefits from this practice because they fall into one of two categories:

1. Open source software, where the work tends to not be for (direct) profit, and the original author is less likely to statutory damages as the result of an infringement.
2. Closed source software, where the source code files are not publicly distributed.

Rather than focus on a hypothetical outcome of obtaining statutory damages, the default behavior of the copyright headers in this project focuses on behavior which can be easily adopted and is appropriate at minimum for a majority of open source projects. The default header is essentially as easy to maintain as not having a header at all (maximum developer efficiency), but does provide a notice regarding the copyright holder for the work. Individual projects may expand on this by customizing the copyright text; as you can see in this project we added brief information regarding the license the source code is provided under and the location where the complete license text may be obtained.

> From what I can tell, the year should come immediately after either "Copyright", "Copr.", or "©". Thus, a generic year-only match is not a semantically equivalent substitute (although it would technically work, essentially you could put {year} tags anywhere, and it would not guarantee the enforcement of a full, legally recognized copyright statement).

Copyright laws vary by country. By using simple replacement variables and customizable copyright text, we maximize the ability of the code to meet the needs of a varied audience. If a generic year could be written as `{creationYear}` (for example), then the copyright notice recommended by the US Copyright Office could be configured by the following line in **stylecop.json**. Further restrictions on the use of `creationYear` in the text (e.g. by requiring it appear after 'Copyright') only serve to limit its overall capability.

```
"copyrightText": "© {creationYear} {companyName}. All rights reserved."
```

> Omitting copyright years may seem like a trivial choice, but as Greg points out, there are real legal ramifications of making this choice.

IMO, the long-term benefits to the developer community as a whole outweigh the marginal legal benefits provided by the more complete notice. This sentiment appears to be backed by de-facto use of notices without years in projects recently published by major organizations (including Microsoft), and is now being considered as a more official recommendation coming from the .NET Foundation itself.

If you can get the .NET Foundation to change its recommendation such that it includes a date in the copyright notice, then it would be much more compelling evidence that the StyleCop Analyzers default is not suitable for most .NET projects.

Travis-Snoozy:
Two-and-half things first...

First: I'm disappearing until next week (this was supposed to be a small one-night change to take care of a snag in getting a project spun up). Don't be alarmed.


Half: The way the current patch works relies on "Copyright ©" being rare/unique enough in actual header text that it can be back-converted into a token. If that is straight-out unacceptable, reject this pull request.

Second: I *can* do a year-only solution. However, It will be quite difficult, and pretty much be 100% based on regexes. It will be much slower than the line-by-line ordinal comparison presently used. Before I embark, I want some level of guarantee that if I code that up, get it to work, document it, and write passing tests for it, that I'm not going to get a bunch of philosophical, ideological, or theoretical-performance pushback. :)

Onward to ideological debate...

>IMO, the long-term benefits to the developer community as a whole outweigh the marginal legal benefits provided by the more complete notice.

Name one benefit of not having a year.

None of the benefits listed on the link provided are valid for years. Point by point:

1. Keeping the headers "as brief as possible [...] to make it easy to apply the right header" is bunk if you're using StyleCop (save for the fact that 2015 StyleCop doesn't handle dates right... but that's a catch-22 ;)).
2. "Minimize [...] polluting git history" is bunk: you should set the copyright header only once, and not be updating the year; this is more-relevant to not having each file copyrighted to a different person.
3. "[...] file header being incorrect by getting out of date over time as code is refactored" is partially valid, but weak; it can be worked around by following best-practices for one class per file, file named as the class. Again, more relevant to per-person attribution.
4. "Make sure we give proper credit [...]" and...
5. "clearly identify [...] who originally submitted a contribution" again has to do with dropping per-person attribution and consolidating that in a single CONTRIBUTORS file. 
6. "[...]properly attributing other open source included in our projects[...]" has more to do with the LICENSE file, and (if code is included as source), the copyright notice should absolutely not be tampered with (there are legal implications surrounding that).

The only explicit reference to removing the copyright year states it's done solely "to avoid unnecessary churn in the code base." That argument is completely invalid if single-year notices are used. When the file is created, the copyright notice should be in it: there should be exactly 0 churn.

> This sentiment appears to be backed by de-facto use of notices without years in projects recently published by major organizations (including Microsoft), and is now being considered as a more official recommendation coming from the .NET Foundation itself.

Now, as in right now, as in two days ago, the same day I submitted this patch. It's not a standard, it's only just been proposed, and it has made exactly zero good points for actually removing the year. Trying to dodge to "de facto" over an actual codified legal standard also doesn't seem like a particularly convincing point. Refusing to look at the consequences of the choice because they are hypothetical also doesn't strengthen this position: if the hypothetical consequences are unimportant, we shouldn't even put a copyright notice in. Copyright is automatic in the US, and even then, copyrights are only important in the hypothetical situation that somebody tries to violate our license. :)

Microsoft, as a large corporation, has its own agenda. It isn't going to be interested in chasing down infringement of its sample code or other such items it releases to the wild. Microsoft is also a predominately closed-source operation, and it is less-likely (though still possible, with so many people internally having access to so much code) that single file(s) will be stolen and then used. If infringement happens, and they care, they're only going to care because huge profits were reaped, or Microsoft's huge profits were eaten into -- statutory damages are absolute chump change to begin with for Microsoft, and they would never opt to take them. This policy makes perfect sense for a company like Microsoft.

For open-source work, where actual damages to the copyright holder may be zero (and where the profits of the infringer may be small by comparison to the whole product the infringement took place in), statutory damages may be the only actual value a "little guy" can get. I refer you to [BusyBox](https://en.wikipedia.org/wiki/BusyBox#GPL_lawsuits), who out of half a dozen settlements, actually won one the single case that went to court... *with statutory damages only*. This code was embedded into hundreds, if not thousands of real devices that went out, but BusyBox still went for statutory rather than actual damages+profits... hardware is, after all, a low-margin industry.

sharwell:
Based on information in [Circular 3](http://www.copyright.gov/circs/circ03.pdf), we see that the impact of including inexact dates is the following:

* If the date is earlier than the actual publication date, then the notice is valid but the copyright protection period is potentially shorter than it otherwise would be.
* If the date is later than the actual publication date, then the notice is equivalent to omitted.

During certain operations such as refactoring which can create new files or move content from one existing file to another, it may be difficult to determine if the content of the new file is original content which was not previously created, or simply relocation of content which was already created (and potentially published). Failure to be exactly correct when choosing a date for the header could result in the invalidation of the notice. The safest way to ensure indisputable protection for all content in the project is to use a copyright notice on all source files which is dated in the year that work on the project was started. For example, this repository would contain a notice dated 2014 on all files, regardless of the time when they were added. The protection would then extend to the year 2109.

This strategy is already supported by StyleCop Analyzers. Considering that C# 1.0 was released in the year 2000, we can assume that projects using StyleCop Analyzers have an earliest date of publication for C# code no earlier than the year 2000. As of today, the discussions in this issue therefore only impact the ability to seek statutory damages for copyright infringements which occur between 2095 (at the earliest) and 2109. I feel like this buys us a little bit of time to really think about a solution which protects developers while minimizing the amount of work required to maintain accurate copyright notices.

> Before I embark, I want some level of guarantee that if I code that up, get it to work, document it, and write passing tests for it, that I'm not going to get a bunch of philosophical, ideological, or theoretical-performance pushback.

The best you can get on this front is filing a proposal describing the specific behavior you want to support and the manner in which you would expose it (e.g. changes to semantics of **stylecop.json**, etc.), and then waiting for it to be approved before starting the implementation. This covers the "...philosophical, ideological..." concerns. In addition, there are several active participants in the project who expressed a great deal of interest in finding ways to incorporate approved functionality in a manner that does not negatively impact performance, including but *definitely* not limited to myself, @pdelvo, and @vweijsters. That said, if things don't work out here then remember two things:

1. It's not just you. I've had three of my own proposals get rejected after I took the time to prepare, implement, and test them (and they weren't simple rules to implement).
2. The license used by code in this repository allows you to create another analyzer which behaves differently. Roslyn allows you to install and use many different analyzer packages in the same project, and you can use a single rule set file to control everything. Several people who use StyleCop.Analyzers have other analyzers installed as part of a complete solution tailored to their product. If your changes don't get merged here, it definitely doesn't mean they were a waste of time.

As a final note, please consider writing up some of your concerns in the .NET Foundation thread on this topic. So far the conversation has been quite one-sided.

Travis-Snoozy:
Thanks for putting in a reference back to this thread on the .NET foundation discussion -- I don't have an account over there, and really I'd rather be coding than adding more accounts to my keyring. :)

Would it be preferred to file a fresh issue with the "proposal" tag, or pile on to #1357?

sharwell:
Either way works. :smile:

PR #1674, opened 2015-10-23
Remove references from analyzers to Microsoft.CodeAnalysis.Workspaces

### LINKED ISSUE #1659: Violations of SA1113 are not reported in the MSBuild output
See #1658 for a clear example. The pull request introduces two violations of SA1113, in addition to violations of SA1116 and SA1117. When you open the solution in Visual Studio, all of these warnings are reported in the Error List window. When you build the project, only SA1116 and SA1117 are included in the build output (see [AppVeyor for example](https://ci.appveyor.com/project/sharwell/stylecopanalyzers/build/1.0.2044)).

:memo: SA1113 is implemented in [SA1113CommaMustBeOnSameLineAsPreviousParameter](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1113CommaMustBeOnSameLineAsPreviousParameter.cs).

sharwell:
Analyzers are not allowed to directly or indirectly reference the **Microsoft.CodeAnalysis.Workspaces** assembly. This means analyzers cannot reference members of code fix providers.

* Refactor several classes to make it easier to spot violations of this rule.
* Fixes #1659, which was caused by the use of properties from `TokenSpacingCodeFixProvider` inside of various analyzers.


vweijsters:
:+1: Changes look good to me.

What a nasty kind of dependency problems. Do you think its worth to have some unit tests that will verify this kind of dependency?

sharwell:
@vweijsters I took a more extreme approach. With two assemblies, it is now impossible for our analyzers to reference types that would cause this problem.

vweijsters:
:+1:

PR #1676, opened 2015-10-23
Fix build warning that snuck in

sharwell:
In addition to reordering the field (which fixed the warning), I made it `readonly` and immutable.


pdelvo:
:+1:

PR #1683, opened 2015-10-24
Reduce allocations

sharwell:
This change reduces the number of allocations which are required to execute `HandleCompilationStart` for the various analyzers. These calls are made many times while users are working in the IDE, so the benefits of this change are most visible there. The potential benefit is at least hundreds of allocations each time the user makes a change while typing in the editor.

Currently the work is complete for documentation and layout rules, and partially complete for the remaining rules.


pdelvo:
:+1:

PR #1702, opened 2015-10-28
Added a code fix provider for the SA1206 diagnostic.

bexxx:


bexxx:
Will start a new PR. Just realized that @sharwell extracted all code fixes to a new assembly.

Issue #1693, opened 2015-10-26
SA1649 code fix does not remove files from linked projects


sharwell:
When renaming a source file **A.cs** &rarr; **B.cs**, the code fix for SA1649 properly updates each project which previously included **A.cs** to now include **B.cs**. However, it only *removes* **A.cs** from the project which was the context for the rename operation.

During the rename, the original file needs to be removed from all projects which contained it.

vweijsters:
Grabbing this

sharwell:
Blocked on dotnet/roslyn#6261.

bjornhellander:
The root cause simply seems to be a missing call to RemoveDocument here:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1649CodeFixProvider.cs#L66-L71

Issue #1665, opened 2015-10-21
SA1649 code fix (rename file) in Shared Project copies the file into all importing projects


AArnott:
Repro steps:
1. Have two ordinary class library projects that both reference a Shared Project in the solution.
2. Add a source file to the Shared project with a file name that will generate an SA1649 warning.
3. Use the code fix to fix the file name.

**Expected**
The file is renamed in the shared project

**Actual**
Both referencing projects now contain the file directly.

I suspect this is actually a Roslyn bug. If so, I'm happy to file that, just let me know. Otherwise feel free to file it yourself and link to this bug.

vweijsters:
Grabbing this

sharwell:
@vweijsters I had assumed this wasn't possible. If you solve it then... that would be awesome.

vweijsters:
Not sure, just saying that I'm having a go at it :smile:

sharwell:
Blocked on dotnet/roslyn#6261.

bjornhellander:
@AArnott / @sharwell: I have been looking into all the problems related to the SA1649 fixer and it's generally looking good, but this specific case related to shared projects seems to have its own problems. Do you have any feeling for how common shared projects actually are?

AArnott:
@bjornhellander  AFAIK they were only well used for Windows 8 targeting. Since then, they've been used as utility folders (where no project imports them at all) by some creative folks. I don't think I have any Shared Projects in any of my repos any more.

bjornhellander:
The root cause seems to be a missing (optional) file path argument in the call to AddDocument here:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1649CodeFixProvider.cs#L66-L71
This has the effect of copying the file into the project instead of adding it as a link.

sharwell:
The upstream feature this needs is available in the Roslyn version shipping first with Visual Studio 16.7. It'll need to be accessed via lightup.

PR #1714, opened 2015-10-30
Disabled SA1649 code fix for linked files

vweijsters:
Changed the behavior of the SA1649 code fix to no longer offer a code fix when a document has linked documents. This was done to protect the user from getting unexpected copied or moved source code files.

This prevents #1693 and #1665 from happening. Proper fixing can only be done when dotnet/roslyn#6261 is resolved.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1649CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1649CodeFixProvider.cs
@@ -34,6 +34,13 @@ namespace StyleCop.Analyzers.DocumentationRules
         /// <inheritdoc/>
         public override Task RegisterCodeFixesAsync(CodeFixContext context)
         {
+            // Do not offer a codefix when there are linked documents.
+            // Roslyn will currently not handle this properly, creating a copy of the file, instead of a link.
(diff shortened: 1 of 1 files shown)


sharwell:
The frustrating thing here is the add/remove document functionality actually does work when you have multiple project files *in the same folder*. This is a layout that many projects I work on use.

Also, the implementation here is disabling the code fix for files that appear in more than one project, whether or not the file is a linked file or simply included as a normal file in more than one project. The code fix will still behave incorrectly for the following cases:

* A file is included as a link into only one project in the solution
* A file is part of a shared project, and that shared project is only referenced by one project in the solution

:thought_balloon: Perhaps it's better to go with a configuration approach where users can disable a code fix if it's known to be problematic in their solution.

pdelvo:
Do we still want this PR?

vweijsters:
I'm still undecided on this, as the issues are still valid. There is no progress on the roslyn issue yet.

It will probably be possible to create a workaround where we use either VisualStudioWorkspace or EnvDTE to make the code fix work properly for Visual Studio only. This will improve the usability for the majority of users, but it will render (part of) the codefix useless outside of Visual Studio.

@sharwell what is your opinion?

sharwell:
Such a frustrating case. In some projects, the code fix works wonderfully, while in others it fails completely. :worried:

vweijsters:
Its not really relevant anymore

PR #1715, opened 2015-10-30
Fixed issues with SX1101 analyzer

vweijsters:
Fixes #1707 
Fixes #1706

PR #1719, opened 2015-10-31
Implement SX1101 using the speculative semantic model

### LINKED ISSUE #1706: SX1101 incorrectly reported for local variable that hides field
SX1101 is reported for the use of `this.` in the following code. The code fix results in a compilation error.

```csharp
class ClassName
{
  int items;
  void Method()
  {
    var items = this.items;
  }
}
```

### LINKED ISSUE #1707: SX1101 incorrectly reported for parameter that hides field
SX1101 is reported for the use of `this.` in the following code. The code fix results in a compilation warning (assignment made to same variable) and (more importantly) a change in code behavior.

```csharp
class ClassName
{
  int items;
  void Method(int items)
  {
    this.items = items;
  }
}
```

sharwell:
:memo: This is an alternative to the approach in #1715.

Thanks to @vweijsters for some of the new tests. :+1:

Fixes #1706
Fixes #1707

Also addresses all of the edge cases found in #1715.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SX1101UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SX1101UnitTests.cs
@@ -59,16 +59,20 @@ public class TestClass : BaseTestClass
             this.TestMethod1();
         }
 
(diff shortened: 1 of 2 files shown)


vweijsters:
:+1: Looks good to me

PR #1724, opened 2015-11-04
Enabled SA1627 by default, moved text strings to resources.

### LINKED ISSUE #1692: SA1627 should be enabled by default
This rule was implemented in #1571, but we forgot to enable it by default.

vweijsters:
Fixes #1692 
Also moved the text strings to the resources.


pdelvo:
:+1:

PR #1726, opened 2015-11-04
Added support to SA1611 for operators and conversion operators

### LINKED ISSUE #1678: SA1611 does not handle operator declarations
During a code review today I observed that SA1611 does not handle `SyntaxKind.OperatorDeclaration` or `SyntaxKind.ConversionOperatorDeclaration`. This should be reviewed to determine if it is a bug or not.

vweijsters:
Fixes #1678
Also moved text strings to the resources.


sharwell:
:+1:

PR #1728, opened 2015-11-04
Added support to SA1502 for operators and conversion operators

### LINKED ISSUE #1677: SA1502 does not handle operator declarations
During a code review today I observed that SA1502 does not handle `SyntaxKind.OperatorDeclaration` or `SyntaxKind.ConversionOperatorDeclaration`. This should be reviewed to determine if it is a bug or not.

vweijsters:
Fixes #1677
Also moved texts to resources.


sharwell:
:+1:

PR #1735, opened 2015-11-07
Fixed issue in UsingCodeFixProvider where a file header followed by …

### LINKED ISSUE #1733: Code fix for SA1200 (Using directives must be inside namespace) doesn't work well with conditional compilation directives
The code fix for the following code results in code that won't compile:

```csharp
// <copyright file="Program.cs" company="PlaceholderCompany">
// Copyright (c) PlaceholderCompany. All rights reserved.
// </copyright>

#if DEBUG
using Fish;
#else
using Fish.Face;
#endif
using System.Text;
using System;

namespace StyleCopBugRepro
{
    class Program
    {
        {
            Int32 q;
            Haddock h;
            StringBuilder sb;
        }
    }
}

namespace Fish
{
    public class Haddock { }

    namespace Face
    {
        public class Haddock { }
    }
}
```

The using block is rearranged as:

```
using Fish;
#else
using Fish.Face;
#endif
using System;
using System.Text;
```

We did hid something similar in real code, though the code fix *did* handle 99.9% of the project correctly and this is a rather unusual case.

vweijsters:
…directive trivia would result in loss of the directive trivia.

Fixes #1733

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.cs
@@ -154,7 +154,8 @@ namespace StyleCop.Analyzers.OrderingRules
             }
 
             var newFirstToken = newSyntaxRoot.GetFirstToken();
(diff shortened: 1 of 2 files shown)


sharwell:
:+1:

PR #1737, opened 2015-11-07
Updating SA1402 to favor a class with the same name as the file.

### LINKED ISSUE #953: SA1402 (FileMayOnlyContainASingleClass) should prefer to keep the type matching the file name
Currently SA1402 reports a warning on all but the first top-level type declaration in a file. However, if a file contains a top-level type whose name matches the name of the file, that type should be the one without a warning reported, regardless of its position within the file.

### Example

**TypeName.cs**

```
namespace NamespaceName
{
    public interface ITypeName { }
    public class TypeName : ITypeName { }
}
```

#### Expected result

Warning SA1402 is reported on `ITypeName`, but not reported on `TypeName`.

#### Actual result

Warning SA1402 is reported on `TypeName`, but not reported on `ITypeName`.

### Impact

Since the code fix for SA1402 uses the diagnostics to determine which types need to move to new files, reporting a diagnostic on a type which should not move results in a violation of SA1649 (FileHeaderFileNameDocumentationMustMatchTypeName) appearing after the code fix is performed.

dcooke28:
Fixes #953

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1402FileMayOnlyContainASingleClass.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1402FileMayOnlyContainASingleClass.cs
@@ -5,6 +5,8 @@ namespace StyleCop.Analyzers.MaintainabilityRules
 {
     using System;
     using System.Collections.Immutable;
+    using System.IO;
+    using System.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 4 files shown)


sharwell:
The basic idea looks fine, but I think we can make a few improvements (I'll add these to the diff).

:bulb: Currently you have Git configured to use **dcooke28** as your name. I recommend you change this to **FirstName LastName** to match the majority of other contributors in the project history.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1402FileMayOnlyContainASingleClass.cs
```diff
@@ -60,37 +62,35 @@ private static void HandleCompilationStart(CompilationStartAnalysisContext conte
...
             var syntaxRoot = context.Tree.GetRoot(context.CancellationToken);
-
             var descentNodes = syntaxRoot.DescendantNodes(descendIntoChildren: node => node != null && !node.IsKind(SyntaxKind.ClassDeclaration));
+            var classNodes = from descentNode in descentNodes
+                             where descentNode.IsKind(SyntaxKind.ClassDeclaration)
+                             select descentNode as ClassDeclarationSyntax;
 
-            string foundClassName = null;
-            bool isPartialClass = false;
+            var preferredClassNode = classNodes.FirstOrDefault(n => n.Identifier.Text == Path.GetFileNameWithoutExtension(context.Tree.FilePath)) ?? classNodes.FirstOrDefault();
```
sharwell: :thought_balloon: The logic for expected file names is actually contained in SA1649. With some (hopefully minor) refactoring, we could reuse that logic here since it accounts for the preferred file naming convention.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1402FileMayOnlyContainASingleClass.cs
```diff
@@ -60,37 +62,35 @@ private static void HandleCompilationStart(CompilationStartAnalysisContext conte
...
 
-            foreach (var node in descentNodes)
+            if (preferredClassNode != null)
```
sharwell: I would prefer to use the following to avoid unnecessary block indentation:

```csharp
if (preferredClassNode == null)
{
  return;
}
```

sharwell:
:memo: I've created a couple of issues related to this diagnostic (#1756, #1760). #1760 in particular may impact this pull request if it is accepted (it would simplify the implementation).

dcooke28:
Do you have any recommendations on where the logic contained in SA1649 should be shared from?  Perhaps NamedTypeHelper?

sharwell:
:+1:

PR #1740, opened 2015-11-08
Fix for 1507

### LINKED ISSUE #1507: SA1202 thrown an exception of type KeyNotFoundException
I am not sure when exactly did that message show up but I've deleted a C# class file before the message and then rebuild.
untitled
Message text:
```Message	AD0001	The User Diagnostic Analyzer 'StyleCop.Analyzers.OrderingRules.SA1202ElementsMustBeOrderedByAccess' threw an exception of type 'System.Collections.Generic.KeyNotFoundException' with message 'The given key was not present in the dictionary.'.```
Restarting Visual Studio of course fixed the problem.

### PR DESCRIPTION
**Pull request by bexxx**
Need to skip all non handled syntax kinds early to avoid looking them up in a dictionary and thus causing a key not found exception.
 
Fixes #1507

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1202UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1202UnitTests.cs
@@ -929,6 +929,43 @@ public class TestClass : TestInterface
 {
     private string Test;
     public string
+    private string
+}
+";
+
+            // We don't care about the syntax errors.
+            var expected = new[]
+            {
+                 new DiagnosticResult
+                 {
+                     Id = "CS1585",
+                     Message = "Member modifier 'public' must precede the member type and name",
+                     Severity = DiagnosticSeverity.Error,
+                     Locations = new[] { new DiagnosticResultLocation("Test0.cs", 5, 5) }
+                 },
+                 new DiagnosticResult
+                 {
+                     Id = "CS1519",
+                     Message = "Invalid token '}' in class, struct, or interface member declaration",
+                     Severity = DiagnosticSeverity.Error,
+                     Locations = new[] { new DiagnosticResultLocation("Test0.cs", 6, 1) }
+                 }
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verifies that the analyzer will properly handle incomplete members.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task Issue150RegressionAsync()
+        {
+            string testCode = @"public class OuterType
+{
+    private string Test;
+    private string
     public string
 }
 ";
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: StyleCop.Analyzers/StyleCop.Analyzers.Test/OrderingRules/SA1202UnitTests.cs [hunk 1]
```diff
@@ -955,6 +955,43 @@ public string
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Verifies that the analyzer will properly handle incomplete members.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task Issue150RegressionAsync()
```
vweijsters: :thought_balloon: I think `Issue1507RegressionAsync` is intended here

vweijsters:
:+1: Looks good to me

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
```diff
@@ -122,6 +122,14 @@ private static void HandleMemberList(SyntaxNodeAnalysisContext context, SyntaxLi
             {
                 var currentSyntaxKind = member.Kind();
                 currentSyntaxKind = currentSyntaxKind == SyntaxKind.EventFieldDeclaration ? SyntaxKind.EventDeclaration : currentSyntaxKind;
+
+                // if the syntaxkind of this member (e.g. SyntaxKind.IncompleteMember) will not
+                // be handled, skip early.
+                if (!((ICollection<SyntaxKind>)MemberNames.Keys).Contains(currentSyntaxKind))
```
sharwell: :exclamation: This can be simplified to the following condition:

```csharp
if (!MemberNames.ContainsKey(currentSyntaxKind))
```

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
```diff
@@ -122,6 +122,14 @@ private static void HandleMemberList(SyntaxNodeAnalysisContext context, SyntaxLi
             {
                 var currentSyntaxKind = member.Kind();
                 currentSyntaxKind = currentSyntaxKind == SyntaxKind.EventFieldDeclaration ? SyntaxKind.EventDeclaration : currentSyntaxKind;
+
+                // if the syntaxkind of this member (e.g. SyntaxKind.IncompleteMember) will not
```
sharwell: :bulb: Consider changing `syntaxkind` to `SyntaxKind`, as the former triggers a spelling violation.

sharwell:
This is basically ready to merge. I verified that the code fix you provided does fail with an AD0001 when the corresponding change to the analyzer is omitted.

:bulb: I highly recommend adding the email address you used for these commits to your GitHub profile. It will not be shared, but it will allow you to associate these commits with your username here. Currently if you view commit 40fe029 the author says Ralf Beckers but you can't click the name to jump to your profile.

bexxx:
addressed all feedback.

pdelvo:
:+1:

PR #1741, opened 2015-11-08
Diagnostic order needs to consider both location and ID

### LINKED ISSUE #1734: DiagnosticVerifier.VerifyDiagnosticResults should cope with non-deterministic Roslyn behavior
Currently `DiagnosticVerifier.VerifyDiagnosticResults` does check the order in which expected diagnostics occur. This occasionally leads to failed tests, because the order in which diagnostics are produced by Roslyn is not (always) deterministic at the moment.

Also see dotnet/roslyn#372 and dotnet/roslyn#1506 for details

sharwell:
Fixes #1734


vweijsters:
:+1: Looks good to me

PR #1742, opened 2015-11-08
Work around perf problems caused by AdditionalText.GetText

### LINKED ISSUE #1736: Investigate workarounds for thread pool explosion
A bug in `AdditionalFile.GetText` (fixed in dotnet/roslyn#6596) can cause thread pool explosions which result in long delays (UI hangs) when analyzers are enabled in Visual Studio. We currently use this method in `SettingsHelper` to read the **stylecop.json** file. As a workaround prior to this bug being fixed, we should investigate the use of an alternative approach (e.g. `File.ReadAllText` based on `AdditionalFile.Path`).

sharwell:
Fixes #1736

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
@@ -5,6 +5,7 @@ namespace StyleCop.Analyzers
 {
     using System.Collections.Immutable;
     using System.IO;
+    using System.Threading;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.Diagnostics;
     using Newtonsoft.Json;
@@ -21,23 +22,25 @@ namespace StyleCop.Analyzers
         /// Gets the StyleCop settings.
         /// </summary>
         /// <param name="context">The context that will be used to determine the StyleCop settings.</param>
+        /// <param name="cancellationToken">The cancellation token that the operation will observe.</param>
         /// <returns>A <see cref="StyleCopSettings"/> instance that represents the StyleCop settings for the given context.</returns>
-        internal static StyleCopSettings GetStyleCopSettings(this SyntaxTreeAnalysisContext context)
+        internal static StyleCopSettings GetStyleCopSettings(this SyntaxTreeAnalysisContext context, CancellationToken cancellationToken)
         {
-            return context.Options.GetStyleCopSettings();
+            return context.Options.GetStyleCopSettings(cancellationToken);
         }
 
         /// <summary>
         /// Gets the StyleCop settings.
         /// </summary>
         /// <param name="options">The analyzer options that will be used to determine the StyleCop settings.</param>
+        /// <param name="cancellationToken">The cancellation token that the operation will observe.</param>
         /// <returns>A <see cref="StyleCopSettings"/> instance that represents the StyleCop settings for the given context.</returns>
-        internal static StyleCopSettings GetStyleCopSettings(this AnalyzerOptions options)
+        internal static StyleCopSettings GetStyleCopSettings(this AnalyzerOptions options, CancellationToken cancellationToken)
         {
-            return GetStyleCopSettings(options != null ? options.AdditionalFiles : ImmutableArray.Create<AdditionalText>());
+            return GetStyleCopSettings(options != null ? options.AdditionalFiles : ImmutableArray.Create<AdditionalText>(), cancellationToken);
         }
 
-        private static StyleCopSettings GetStyleCopSettings(ImmutableArray<AdditionalText> additionalFiles)
+        private static StyleCopSettings GetStyleCopSettings(ImmutableArray<AdditionalText> additionalFiles, CancellationToken cancellationToken)
         {
             try
             {
@@ -45,7 +48,8 @@ namespace StyleCop.Analyzers
                 {
                     if (Path.GetFileName(additionalFile.Path).ToLowerInvariant() == SettingsFileName)
                     {
-                        var root = JsonConvert.DeserializeObject<SettingsFile>(additionalFile.GetText().ToString());
+                        string additionalTextContent = additionalFile.GetText(cancellationToken).ToString();
+                        var root = JsonConvert.DeserializeObject<SettingsFile>(additionalTextContent);
                         return root.Settings;
                     }
                 }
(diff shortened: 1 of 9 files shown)


sharwell:
@CyrusNajmabadi @jasonmalinowski @Pilchie @srivatsn Can one of you verify that this workaround will bypass the problematic code when running within Visual Studio?

jasonmalinowski:
It does avoid the problematic call. It does break scenarios where you have the file open in Visual Studio, since you'll still be reading the file from disk. It might have other fun side effects as well.

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
```diff
@@ -57,5 +66,51 @@ private static StyleCopSettings GetStyleCopSettings(ImmutableArray<AdditionalTex
...
+            }
+
+            try
+            {
+                if (fileExists(additionalText.Path))
+                {
+                    return fileReadAllText(additionalText.Path);
+                }
+            }
+            catch (IOException)
```
jasonmalinowski: This can throw [all sorts of exceptions](https://msdn.microsoft.com/en-us/library/ms143368%28v=vs.110%29.aspx#Anchor_1). You'll want to catch them all.

sharwell: Between the call to `Exists` and catching `IOException`, the only one remaining is unauthorized access exception. This really isn't expected for a file contained in the project.

sharwell:
Interesting. I hadn't thought about the case of files opened in the editor. Considering the impact of the bug we're trying to avoid, would you recommend another approach?

sharwell:
Another possibility is implementing a per-compilation cache for this value. Currently we read the value in 6 methods registered with `RegisterCompilationStartAction`. In the solution where I experienced the most problems, only two diagnostics associated with these analyzers were enabled. Since we have 22 C# projects, it seems this would mean up to 44 calls were made to `GetText` after a change was made to source code in the solution. I'm not sure how much we would gain to only have 1 call per compilation. How long does it take for the threads created by this call to be reclaimed by the thread pool?

jasonmalinowski:
Where are you reading this today that isn't per-compilation?

sharwell:
> Where are you reading this today that isn't per-compilation?

Each read location occurs only once per compilation, but unlike the way we identify generated files the result is not shared across the various analyzers which need to access settings. By sharing the result, we reduce the number of calls to GetText from once *per analyzer* (which uses settings) per compilation to once per compilation altogether.

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
pdelvo: I would just catch exceptions and dont bother with Exists. If a file is deleted between the exists and the read all text call this will crash.

sharwell: The deletion case (`FileNotFoundException`) is already covered by `IOException`.

Some workspaces don't place file content on disk at all. The call to `File.Exists` is used to check that the file with a given path actually does exist on disk before trying to read it. If another error occurs during the read which we truly hadn't thought of I don't mind getting an AD0001.

jasonmalinowski:
Oh, that would be a very good optimization to make in general. :smile: In general, the fewer parallel calls we have here, the better off you'll be not hitting this perf issue.

> How long does it take for the threads created by this call to be reclaimed by the thread pool?

So the problem is when we're already on the thread pool, we're spawning additional work to the thread pool and then synchronously waiting on it. If it happens that the thread pool isn't already saturated this will complete almost immediately. If it is saturated, then it could block for a few milliseconds while another thread pool job completes, or it could take any number seconds if we've live-locked the thread pool. It's very non-deterministic. :frowning:

sharwell:
@jasonmalinowski Would this be "better"?

```csharp

private static async Task<SourceText> GetAdditionalTextAsync(
    AdditionalText additionalText,
    CancellationToken cancellationToken)
{
    return await Task.Factory.StartNew(
        () => additionalText.GetText(cancellationToken),
        cancellationToken,
        TaskCreationOptions.LongRunning,
        TaskScheduler.Default).ConfigureAwait(false);
}
```

sharwell:
I suppose I could always just use reflection to implement the same thing that fixed the original code...

```csharp
private static SourceText GetText(AdditionalText additionalText, CancellationToken cancellationToken)
{
    object document = GetField(additionalText, "_document");
    if (document != null)
    {
        object textSource = GetField(document, "textSource");
        if (textSource != null)
        {
            object textAndVersion = CallMethod(textSource, "GetValue", cancellationToken);
            if (textAndVersion != null)
            {
                SourceText text = GetProperty(textAndVersion, "Text");
                if (text != null)
                {
                    return text;
                }
            }
        }
    }

    return additionalText.GetText(cancellationToken);
}
```

jasonmalinowski:
@sharwell: yes, to your Task.Factory.StartNew() problem -- although the LongRunning would mean you'd prevent starving the thread pool, you still probably can't consume that API in any meaningful way since you don't have asynchrony higher up.

For reflection, obviously there's all the usual dangers of that. I would encourage in that case to do a version check because it's *very* possible we could change the fields and their meaning internally and break you again.

sharwell:
I updated the code to use the reflection approach which bypasses the problematic method but still uses Roslyn (as opposed to the file system APIs). Do you know what version of Roslyn will contain the fixed code?

jasonmalinowski:
We already have the fix in master, so anything with an assembly version of 1.2 or later will be fine.

sharwell:
Thank you so much for the help @jasonmalinowski!

This should now be ready for review.

vweijsters:
:+1: Looks good to me

jasonmalinowski:
And from the Roslyn API (ab)use perspective, :+1:.

CyrusNajmabadi:
Hey @sharwell Did you this work out for you?

sharwell:
@CyrusNajmabadi I haven't deployed this build to the environment which was having the major performance problems, and probably won't be able to for another month since the code base is in a stabilization period.

CyrusNajmabadi:
Ok.  Let us know (ideally sooner vs later)!  It will help us with making some API decisions we're considering.  Thanks!

PR #1743, opened 2015-11-09
Fix that SA1649 does not support simple names

### LINKED ISSUE #1730: SA1649 (File name must match first type name) for generic types shouldn't require generic specification
For the type `Fish<T>`, legacy StyleCop permits either "Fish{T}.cs" *or* "Fish.cs" as the file name.

Our analyzer allows the former, prefers "Fish`1" but does not accept plain old "Fish.cs". I like the legacy behaviour, since I prefer to omit the qualification in the file name unless there is a non-generic version of the type too.

Apologies if this has already been discussed, but I couldn't find evidence of that.

pdelvo:
Fixes #1730.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1649FileNameMustMatchTypeName.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1649FileNameMustMatchTypeName.cs
@@ -98,6 +98,12 @@ namespace StyleCop.Analyzers.DocumentationRules
 
                 if (string.Compare(fileName, expectedFileName, StringComparison.OrdinalIgnoreCase) != 0)
                 {
+                    if (this.fileNamingConvention == FileNamingConvention.StyleCop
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1649FileNameMustMatchTypeName.cs
```diff
@@ -84,22 +84,22 @@ public void HandleSyntaxTreeAction(SyntaxTreeAnalysisContext context)
                 }
 
                 var fileName = Path.GetFileName(context.Tree.FilePath);
-                string expectedFileName;
+                string[] expectedFileNames;
```
sharwell: :question: Since this is in an analyzer (as opposed to a code fix), can we implement this without the additional array and closure allocations?

pdelvo: I changed that

vweijsters:
:+1: Looks good to me

PR #1751, opened 2015-11-11
Improved indentation handling of IndentationHelper.GetIndentationSteps

### LINKED ISSUE #1036: Review handling of mixed indentation in IndentationHelper.GetIndentationSteps
In the current implementation, `GetIndentationSteps` will return 5 for code indented with the following string:

```
" \t \t \t \t"
```

It's more likely that the user expected this to return 4.

In addition, we should carefully consider the impact of `IndentationSize` and `TabSize` on this code in the case where they differ.

(Originally found during review of #999, but it came up previously in #666.)

### LINKED ISSUE #1037: Review handling of partial indentation in IndentationHelper.GetIndentationSteps
Currently, the implementation of `GetIndentationSteps` truncates any partial indentation. We should carefully consider the impact of each of the following possible ways to handle partial indentation, and use the one that users are most likely to expect.

1. Truncate (for indentation size of 4, an indent of 0-3 spaces is treated as not an indent)
2. Ceiling (for indentation size of 4, and indent of 1-4 spaces is treated as a single indent)
3. Preserve (the code which currently uses `GetIndentationSteps` is updated to use `GetIndentationSize` instead, which always preserves the exact spacing of the input)

:memo: Found during review of #999.

vweijsters:
Fixes #1036
Fixes #1037

For #1037 I've implemented a fourth option: rounding. For indentation size 4 this means: 0-1 no indent, 2-3 single indent. This is my personal preference, as it is forgiving when there is one space to many / little. I'll be happy to adapt it if wanted.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/IndentationHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/IndentationHelper.cs
@@ -5,6 +5,7 @@ namespace StyleCop.Analyzers.Helpers
 {
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
+    using StyleCop.Analyzers.Helpers.ObjectPools;
 
     /// <summary>
     /// Provides helper methods to work with indentation.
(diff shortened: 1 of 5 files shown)


sharwell:
This seems fine.

I do wonder if we should instead simply use the previous indentation string unaltered, rather than trying to rewrite it using the currently configured indentation characters. But that is a separate issue.

PR #1752, opened 2015-11-11
Improved initializer handling of SA1513

### LINKED ISSUE #1713: SA1513 and SA1009 Conflict and code fixes fight between themselves
In this program : 

```csharp
using System;

class Program
{
    private class Foo : IDisposable
    {
        public int Bar { get; set; }

        public void Dispose()
        {
        }
    }

    {
        foreach (
            var x in
            new[]
            {
                1,
                2,
                3
            })
        {
        }

        using (
            new Foo
            {
                Bar = 1
            })
        {
        }

        Console.WriteLine(
            new[]
            {
                1,
                2,
                3
            });
    }
}
```

Both the `foreach` and `using` trigger SA1513 at the end, but it isn't solvable, applying the codefix will produce SA1009.

Also applying SA1513 code fix once doesn't work, it add a line break between `}` and `)` but no new line, applying it a second time fixes it but SA1009 is still there.

vweijsters:
Fixes #1713


sharwell:
:+1:

PR #1754, opened 2015-11-11
Added complex element initializer support to SA1500

### LINKED ISSUE #1679: SA1500 does not handle complex element initializer expressions
During a code review today I observed that SA1500 does not handle `SyntaxKind.ComplexElementInitializerExpression`. This should be reviewed to determine if it is a bug or not.

vweijsters:
Fixes #1679


sharwell:
:+1:

PR #1755, opened 2015-11-11
SA1001 no longer accepts commas as first token on a line

### LINKED ISSUE #1663: SA1001 should not allow a comma at the beginning of the line
As discussed in #1655, SA1001 should not allow a comma at the beginning of the line. The documentation needs to be updated to reflect this, and new test cases added for code containing commas which is not a parameter or argument list (which are covered by SA1113).

vweijsters:
Fixes #1663


sharwell:
:+1:

Issue #1761, opened 2015-11-12
Implement SA1135: Using directives must be qualified


0x084E:
In Visual Studio 2015 if you add a using directive via quick actions inside of a namespace the namespace in the using statement is cut down if a part of the included namespace is also part of the outer namepsace.

E.g.: You Are In N1.N2 and you include N1.N2.N3 it is cut down to N3
`namepsace N1.N2 { using N3; }`
You are in N1.N2 and include N1.N4.N5 it is cut down to N4.N5
`namepsace N1.N2 { using N4.N5; }`

I think this is problematic since you do not know what the namespace actually is. In the second example I would think from the context that N4 is contained in N1.N2 but it is not. What's your opinion?

sharwell:
I've been back and forth on this issue. I think it does make sense though, and I also think most existing code using StyleCop Classic would already follow this rule.

:+1:

vweijsters:
:+1: I prefer the explicit clarity without me having to puzzle at what point the common part starts.

0x084E:
I've mentioned one more problem: Consider the namepsaces `System.Xml``System.Xml.Linq` and `System.Linq`. If you want to import a type from `System.Xml.Linq` or `System.Linq` into `System.Xml` the analyzer is smart enough to detect that importing one of them as `Linq` would result in a not resolvable using and imports it qualified.
If either of `System.Xml.Linq` or `System.Linq` doesn't exist at the time of importing it is imported as `Linq` (of course) but if you add it after importing your code won't compile anymore...

vweijsters:
Grabbing this. 

My proposal for the name is: `SA1135 Using directives must be fully qualified`

sharwell:
@vweijsters There is already a pull request open for this (#1762)

0x084E:
@vweijsters You can undertake it if you want. Maybe someone with more knowledge about roslyn can do it better. I just did it myself because I really wanted to have it.

vweijsters:
I missed the pull request :blush: 

@0x084E no point in me redoing the work. I'll have a look at the pull request.

icnocop:
For reference, @sharwell has since merged https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1979 to work around the issue reported in https://github.com/dotnet/roslyn/issues/7446.

sharwell:
@tdworshak Can you file a new bug for that? 👍

tdworshak:
[...]

PR #1762, opened 2015-11-12
Proposal: New rule  'SA1135UsingDirectivesMustBeQualified'

0x084E:
See #1761
Test is failing because HandleNamespaceDeclaration in SA1135UsingDirectivesMustBeQualified is called twice for the same namespace (and therefore all violations are reported twice), I don't know why...


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/ReadabilityResources.resx
```diff
@@ -477,6 +477,18 @@
...
     <value>Attributes must not share line</value>
   </data>
+  <data name="SA1135CodeFix" xml:space="preserve">
+    <value>Qualify using directive</value>
+  </data>
+  <data name="SA1135Description" xml:space="preserve">
+    <value>All using directives must be qualified.</value>
+  </data>
+  <data name="SA1135MessageFormat" xml:space="preserve">
+    <value>The using directive '{0}' must be qualified to '{1}'</value>
```
sharwell: :question: How about the following?

> Using directive for namespace '{0}' must be qualified

:memo: In this message, `{0}` is the value you referred to as `{1}` in the original message.

sharwell:
:exclamation: This will definitely need tests for globally-qualified namespaces (those starting with `global::` or another alias).

Marked as blocked pending acceptance of #1761.

sharwell:
> Test is failing because HandleNamespaceDeclaration in SA1135UsingDirectivesMustBeQualified is called twice for the same namespace (and therefore all violations are reported twice), I don't know why...

I recommend filing a bug report in the dotnet/roslyn repository with a repro case. When this occurs, your tests will need two identical `DiagnosticResult` elements in the expected results array.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <inheritdoc/>
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterCompilationStartAction(HandleCompilationStart);
+        }
+
+        private static void HandleCompilationStart(CompilationStartAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionHonorExclusions(HandleNamespaceDeclaration, SyntaxKind.NamespaceDeclaration);
```
sharwell: :exclamation: Since the analysis is limited to one using directive at a time, register your analysis for `SyntaxKind.UsingDirective` instead of `SyntaxKind.NamespaceDeclaration`.

0x084E:
>This will definitely need tests for globally-qualified namespaces (those starting with global:: or another alias).

I'm going to add them. (I didn't in first place because currently the analyzer skips using directives with and alias and/or a static keyword.)

>I recommend filing a bug report in the dotnet/roslyn repository with a repro case. When this occurs, your tests will need two identical `DiagnosticResult` elements in the expected results array.

OK, will do. Didn't know this is a roslyn issue.

pdelvo:
I already reported the bug that HandleNamespaceDeclaration is called multiple times and it is fixed for 1.1: https://github.com/dotnet/roslyn/issues/4745

0x084E:
> This will definitely need tests for globally-qualified namespaces

Good one, these were actually reported by the analyser. I'm know skipping usings containing `::`. If anybody knows a better way of checking for globally-qualified namespaces please tell me.

The analyzer now reports aliased namepsaces and types (e.g. `R = Reflection` and `RA = Reflection.Assembly`) also.

I've added more test but after chaning the message I get an error
`System.FormatException : Index (zero based) must be greater than or equal to zero and less than the size of the argument list.`
I don't know what I'm doing wrong...

0x084E:
If aliased types should be reported also (as it is in the latest commit) the message
`Using directive for namespace '{0}' must be qualified`
is missleading since the reported value isn't a namespace.

sharwell:
@0x084E Interesting (and valid) point. What about two messages?

> Type '{0}' must be fully qualified in using directive

> Namespace '{0}' must be fully qualified in using directive

Our code includes some examples of cases where one diagnostic ID can have multiple messages - an example is SA1633.

:memo: When two messages are used instead of making Type/Namespace an additional variable, it is easier to provide localized messages in the future.

0x084E:
Now using two different messages for namespaces and aliased types (and finally passing all tests :)).

0x084E:
I've noticed a problem since merging the latest changes. The following raises a warning

```csharp
namespace System.Collection
{
    using IO;
}
```

but this does not

```csharp
namespace System.Collections
{
    using IO;
}
```
(notice the trailling 's' in Collections)  

The analyzer doesn't even seem to be called in the second case.
Has anyone a hint what is going on here?

0x084E:
@vweijsters did you have the time to take a look? The problems are roslyn related and will be most likely worked around with #7446, too.

icnocop:
For reference, @sharwell has since merged https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1979 to work around the issue reported in https://github.com/dotnet/roslyn/issues/7446.

vweijsters:
Superseeded by #2592

PR #1765, opened 2015-11-12
Implemented generated file filters

### LINKED ISSUE #1430: C# files in the obj/ folder should be considered Auto-Generated.
This is standard practice for auto-generated code.

### LINKED ISSUE #1699: Add more ways to exclude files from Analysis
In our project using legacy StyleCop, we put an `//<unstyled/>` comment on top of the file, I know there's an `<autogenerated>` comment but 'unstyled' would be more fitting.

Addidtionally we used this GeneratedFileFIlters to exclude things like EF Migrations. It looks like this:
````xml
 <CollectionProperty Name="GeneratedFileFilters">
   <Value>\.g\.cs$</Value>
   <Value>\.generated\.cs$</Value>
   <Value>\.g\.i\.cs$</Value>
   <Value>\.xaml\.cs$</Value>
   <Value>\.aspx\.cs$</Value>
   <Value>Form\.cs$</Value>
   <Value>Mixin\.cs$</Value>
   <Value>SpecFlow_\d*\.cs$</Value>
   <Value>SpecFlow_\d*\.feature\.cs$</Value>
   <Value>Generation\.cs$</Value>
   <Value>Designer\.cs$</Value>
   <Value>(\d{15})_(.*)\.cs$</Value>
 </CollectionProperty>
```
Exclusions of generated files using a custom filter is quite an essential feature for large projects.

### LINKED ISSUE #1710: Proposal: Disable SA1649 for file names that contain a period
Given a class "ClassName" and a filename of "ClassName.Generated.cs", I think this should be valid for the rule. I believe it's a common practice for naming code generated files.

vweijsters:
Fixes #1430
Fixes #1699 
Fixes #1710 

I think I got everything covered, but please review.
As default filters I used the same defaults as StyleCop classic, with `*.designer.cs` added.


LePastis:
Hello,
We are currently looking to switch to StyleCop StyleCop.Analyzer in our business. However, this feature greatly missing in the current version.
When do you validate this pull request

sharwell:
@LePastis I'm hoping to not need to handle this directly inside our analyzers. The Roslyn team is actively working on this functionality for the 1.2 release, and anything we do here would either our maintenance burden, be confusing to users, and/or break functionality for users after Roslyn 1.2 is released. See dotnet/roslyn#7000 and its multiple child issues for more information.

vweijsters:
I'm marking this code review, as dotnet/roslyn#7000 has been pushed back to 1.3

sharwell:
Starting with 1.1 Beta 1, generated files are handled entirely by Roslyn itself, so this change is no longer required.

PR #1046, opened 2015-07-25
Renamed 'curly bracket' to 'curly brace'

EdwinEngelen:
Fixes #989

PR #1769, opened 2015-11-15
Replace the term "curly bracket" with "curly brace"

### LINKED ISSUE #989: Rename 'bracket' to 'brace' when referring to { and }
StyleCop generally refers to `{` and `}` as "curly brackets". However, Roslyn in general refers to these as "curly braces". In my experience (which is admittedly biased by working in parsing for so long), braces seems to be the more widely used term.

I propose changing all uses (code and documentation) of 'bracket' to 'brace' when referring to curly braces.

sharwell:
Fixes #989
Supersedes #1046


StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1013UnitTests.cs
```diff
@@ -134,11 +134,11 @@ public class TestClass
         }
 
         /// <summary>
-        /// Verifies that the analyzer will properly handle closing curly brackets in nested curly brackets.
+        /// Verifies that the analyzer will properly handle closing curly braces in nested curly brackets.
```
vweijsters: you missed the 2nd curly brackets usage here

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/LayoutRules/SA1500CodeFixProvider.cs
```diff
@@ -201,7 +201,7 @@ private static bool IsValidFollowingToken(SyntaxToken nextToken)
 
         private static int DetermineIndentationSteps(IndentationOptions indentationOptions, SyntaxToken token)
         {
-            // For a closing curly bracket use the indentation of the corresponding opening curly bracket
+            // For a closing curly brace use the indentation of the corresponding opening curly bracket
```
vweijsters: You missed the 2nd curly bracket reference here.

vweijsters:
:question: Wouldn't it make sense to use only brace, instead of curly brace, since Roslyn is also using only brace?

sharwell:
@vweijsters Updated!

vweijsters:
:+1: Looks good!

PR #1784, opened 2015-11-16
Fixes null reference exception in SA1129

### LINKED ISSUE #1668: SA1129DoNotUseDefaultValueTypeConstructor throws System.NullReferenceException
Got this AD0001 message while (heavily) editing source code in VS2015 using beta 14.

> The User Diagnostic Analyzer 'StyleCop.Analyzers.ReadabilityRules.SA1129DoNotUseDefaultValueTypeConstructor' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.

I haven't been able to reproduce it yet.

vweijsters:
Fixes #1668

Fixed the only location in `HandleObjectCreationExpression` where a null pointer exception could theoretically occur. Any suggestions on how to reproduce this in a test case are welcome.


sharwell:
:+1:

PR #1790, opened 2015-11-18
Replace custom MSBuild targets with improved documentation

### LINKED ISSUE #1788: Remove custom build task in favor of documentation
I think we should revert #1581 since it still results in various problems like #1684 and #1720. We originally did this and ran into problems where users did not realize manual steps were needed after applying the code fix in order for stylecop.json settings to take effect. This can be resolved by adding a comment at the top of the generated stylecop.json file:

```
// ACTION REQUIRED: This file was automatically added to your project, but it
// will not take effect until additional steps are taken to enable it. See the
// following page for additional information:
//
// https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/EnableConfiguration.md
```

:memo: This is not technically valid JSON syntax, but it is supported by both the Visual Studio JSON editor and Newtonsoft.Json library we use for reading the file.

sharwell:
Fixes #1788


vweijsters:
:+1: Looks good, time to get the release out :smile:

PR #1792, opened 2015-11-18
SA1003 now ignores pointer dereference

### LINKED ISSUE #1776: Code triggers both SA1003 and SA1023
See #1766

Issue D5: Related to SA1003 and SA1023

The following code would only trigger SA1023 in classic StyleCop. The current analyzer triggers both SA1023 and SA1003. (This is unsafe code.)

```
int* value = null;
int blah = * value;
```

vweijsters:
Fixes #1776


sharwell:
:+1:

PR #1796, opened 2015-11-20
Unwrap the TargetInvocationException for Reflection calls

### LINKED ISSUE #1795: SettingsHelper.GetText needs to rethrow properly
Currently when an `OperationCanceledException` is thrown while reading the text of **stylecop.json**, it is wrapped inside of a `TargetInvocationException` due to the way we use reflection to invoke `GetText`. This can result in an AD0001 being thrown by any of the analyzers which read from this file.

The `TargetInvocationException` needs to be unwrapped, and the inner exception rethrown.

sharwell:
Fixes #1795


pdelvo:
:+1: How hard would be it be to test this?

sharwell:
> How hard would be it be to test this?

Would be very hard to test in the unit tests, not counting the two `throw;` lines which are unreachable at runtime.

pdelvo:
:+1:

PR #1797, opened 2015-11-20
Run code fixes sequentially for unit testing in the debugger

sharwell:
Previously it was difficult to debug code fixes because multiple threads would be stepping through the same code at the same time.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/CodeFixVerifier.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/CodeFixVerifier.cs
@@ -6,6 +6,7 @@ namespace TestHelper
     using System;
     using System.Collections.Generic;
     using System.Collections.Immutable;
+    using System.Diagnostics;
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/CodeFixVerifier.cs
```diff
@@ -87,13 +88,37 @@ public bool UseTabs
...
+                {
+                    await t3;
+                }
+
                 var t4 = this.VerifyFixInternalAsync(LanguageNames.CSharp, this.GetCSharpDiagnosticAnalyzers().ToImmutableArray(), this.GetCSharpCodeFixProvider(), oldSource, batchNewSource ?? newSource, codeFixIndex, allowNewCompilerDiagnostics, maxNumberOfIterations, GetFixAllAnalyzerSolutionAsync, cancellationToken).ConfigureAwait(false);
-                await t1;
-                await t2;
-                await t3;
-                await t4;
+                if (Debugger.IsAttached)
```
pdelvo: Is the last change really necessary? awaiting a finished task should be a no-op.

sharwell: It's not necessary. I just thought it might be easier to read if the pattern was applied consistently.

pdelvo:
:+1:

PR #1802, opened 2015-11-21
Add information about known changes

### LINKED ISSUE #359: Suggestion: Known Changes Document

### LINKED ISSUE #1773: Documentation bug with SA1642
See #1766 

Issue D1: Related to SA1642.

If there is a class with a private empty constructor, the documentation text in classic StyleCop requires it to begin with "Prevents a default instance of the <see cref="whatever" /> class from being constructed." The current rule requires "Initializes a new instance...".

Sam's reply:

The behavior is by-design, see #413. The documentation does not accurately reflect this, so a documentation bug should be created.

sharwell:
Fixes #359
Fixes #1773

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ The severity of individual rules may be configured using [rule set files](https:
 in Visual Studio 2015. **Settings.StyleCop** is not supported, but a **stylecop.json** file may be used to customize the
 behavior of certain rules. See [Configuration.md](documentation/Configuration.md) for more information.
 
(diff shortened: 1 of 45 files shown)


sharwell:
@GregReddick I would particularly value your feedback on this pull request. :smile:

vweijsters:
:+1:

sharwell:
@GregReddick If you find anything I missed, we can of course update this. :smile:

PR #1805, opened 2015-11-22
StyleCopSettings cache

### LINKED ISSUE #1803: Cache StyleCopSettings for the Compilation
With the recent work to add options to additional analyzers, it's becoming more important to cache the `StyleCopSettings` instance so we aren't deserializing it multiple times per compilation.

sharwell:
Fixes #1803

:memo: To minimize the size of the diffs and make them easier to review, I left the `Analyzer` classes in place, but made them static. I may submit a future review to remove these classes and put the methods in the parent type.


vweijsters:
:+1: Looks good to me. I would prefer to see the Analyzer classes removed.

PR #1812, opened 2015-11-23
Implement the 'elementOrder' configuration property

### LINKED ISSUE #1745: Convert ElementOrderCodeFixProvider use of IsAnalyzerSuppressed to configuration
Currently the behavior of `ElementOrderCodeFixProvider` varies based on whether or not various rules are suppressed (SA1201, SA1202, SA1203, SA1204, SA1214, SA1215). The code fix should not rely on this type of configuration; the behavior should instead be exposed as explicit options in **stylecop.json**.

sharwell:
Fixes #1745 

:exclamation: This pull request merges SA1214 and SA1215 into a single rule (SA1214). The Known Changes documentation and the documentation for each of these rules was updated to reflect the change.

:thought_balloon: Additional tests for non-standard configurations will be required. The current tests cover a couple of common configurations (including extensive testing of the default configuration). I would prefer to get the new implementation out earlier for testing in the wild and add the additional tests separately.


vweijsters:
I had a quick look at the documentation only for now.

Overall it looks good. I got one remark:
- The documentation is incomplete with respect to describing what order is implied by the kind and accessibility settings. It can be reconstructed by looking at the various impacted rules, but it would be nice to have it all in the same place.

sharwell:
@vweijsters I updated this to improve documentation and messaging.

oatkins:
This seems like a step in the right direction to me. Legacy SC defines many ordering rules, but it really isn't useful to pick and choose which ones to enable because they all have implicit knowledge of each other. Amalgamating some of the rules is therefore very reasonable.

I'm not sure that I'll use the config option to customize the traits that determine the correct ordering, but I do appreciate that this makes the rules much more explicit and easier to understand. We are also moving away from the legacy "my way or the highway" approach; a lot of people will appreciate that.

Looking at this on my phone, so I've not studied the code; just the documentation. But a wholehearted +1 on that basis.

sharwell:
@oatkins I believe we will have the ability to expand on this customization in the future to cover the "kind" and "accessibility" traits in more detail. But I agree, the most important aspect right at this point is clarity regarding the relative order of the rules.

I'll wait for @vweijsters to finish the implementation review before merging.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/MemberOrderHelper.cs
```diff
@@ -113,63 +127,42 @@ private enum ModifierFlags
...
         {
-            var flags = ModifierFlags.None;
-            if (checks.Const && syntax.Any(SyntaxKind.ConstKeyword))
+            SyntaxKind type = member.Kind();
+
+            AccessLevel accessibility;
+            if ((type == SyntaxKind.ConstructorDeclaration && modifiers.Any(SyntaxKind.StaticKeyword))
+                || (type == SyntaxKind.MethodDeclaration && ((MethodDeclarationSyntax)member).ExplicitInterfaceSpecifier != null)
+                || (type == SyntaxKind.PropertyDeclaration && ((PropertyDeclarationSyntax)member).ExplicitInterfaceSpecifier != null)
+                || (type == SyntaxKind.IndexerDeclaration && ((IndexerDeclarationSyntax)member).ExplicitInterfaceSpecifier != null))
```
vweijsters: I think that `SyntaxKind.EventDeclaration` should also be addressed, as events can also be implement explicitly.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214ReadonlyElementsMustAppearBeforeNonReadonlyElements.cs
```diff
@@ -0,0 +1,150 @@
...
+            var elementOrder = settings.OrderingRules.ElementOrder;
+            int readonlyIndex = elementOrder.IndexOf(OrderingTrait.Readonly);
+            if (readonlyIndex < 0)
+            {
+                return;
+            }
+
+            var typeDeclaration = (TypeDeclarationSyntax)context.Node;
+
+            FieldDeclarationSyntax previousMember = null;
```
vweijsters: :thought_balloon: Consider renaming this to previousField to improve readability

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214ReadonlyElementsMustAppearBeforeNonReadonlyElements.cs
```diff
@@ -0,0 +1,150 @@
...
+                    case OrderingTrait.Accessibility:
+                        if (previousAccessLevel != currentAccessLevel)
+                        {
+                            compareReadonly = false;
+                        }
+
+                        continue;
+
+                    case OrderingTrait.Constant:
+                        if (previousFieldConst || currentFieldConst)
```
vweijsters: :bug: This will skip read-only checks for two consecutive constant fields. The operator should be `!=` or `^`.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214ReadonlyElementsMustAppearBeforeNonReadonlyElements.cs
```diff
@@ -0,0 +1,150 @@
...
+                    previousMember = field;
+                    previousFieldConst = currentFieldConst;
+                    previousFieldStatic = currentFieldStatic;
+                    previousFieldReadonly = currentFieldReadonly;
+                    previousAccessLevel = currentAccessLevel;
+                    continue;
+                }
+
+                bool compareReadonly = true;
+                for (int j = 0; j < readonlyIndex; j++)
```
vweijsters: :thought_balloon: consider adding `compareReadonly` to the condition check, to stop parsing the element order as soon as the outcome is known.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214ReadonlyElementsMustAppearBeforeNonReadonlyElements.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                    case OrderingTrait.Readonly:
+                    default:
+                        continue;
+                    }
+                }
+
+                if (compareReadonly)
+                {
+                    if ((currentFieldReadonly || currentFieldConst) && !previousFieldReadonly && !previousFieldConst)
```
vweijsters: :exclamation: I think that constants should not be considered here. They are handled by SA1203 exclusively.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1204StaticElementsMustAppearBeforeInstanceElements.cs
```diff
@@ -79,70 +62,126 @@ private static void HandleCompilationStart(CompilationStartAnalysisContext conte
...
+            }
+
             var typeDeclaration = (TypeDeclarationSyntax)context.Node;
 
-            HandleMemberList(context, typeDeclaration.Members, AccessLevel.Private);
+            HandleMemberList(context, elementOrder, staticIndex, typeDeclaration.Members, AccessLevel.Private);
         }
 
-        private static void HandleMemberList(SyntaxNodeAnalysisContext context, SyntaxList<MemberDeclarationSyntax> members, AccessLevel defaultAccessLevel)
+        private static void HandleMemberList(SyntaxNodeAnalysisContext context, ImmutableArray<OrderingTrait> elementOrder, int staticIndex, SyntaxList<MemberDeclarationSyntax> members, AccessLevel defaultAccessLevel)
```
vweijsters: :thought_balloon: The `defaultAccessLevel` parameter can be removed, as it is not used anymore.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
```diff
@@ -125,39 +145,79 @@ private static void HandleMemberList(SyntaxNodeAnalysisContext context, SyntaxLi
...
-                {
-                    currentAccessLevel = AccessLevelHelper.GetAccessLevel(modifiers);
-                    currentAccessLevel = currentAccessLevel == AccessLevel.NotSpecified ? defaultAccessLevel : currentAccessLevel;
-                }
+                AccessLevel currentAccessLevel = MemberOrderHelper.GetAccessLevelForOrdering(member, modifiers);
 
-                if (previousAccessLevel != AccessLevel.NotSpecified
-                    && currentSyntaxKind == previousSyntaxKind
-                    && currentAccessLevel > previousAccessLevel)
+                if (previousMember != null && previousAccessLevel != AccessLevel.NotSpecified)
```
vweijsters: :thought_balloon: No point in doing both checks. For the first iteration, the check is always true, after the first iteration previousMember will never be null again, so the second check is unnecessary.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
```diff
@@ -125,39 +145,79 @@ private static void HandleMemberList(SyntaxNodeAnalysisContext context, SyntaxLi
...
+                            {
+                                compareAccessLevel = false;
+                            }
+
+                            continue;
+
+                        case OrderingTrait.Constant:
+                            // Only fields may be marked const
+                            bool previousIsConst = previousMember.IsKind(SyntaxKind.FieldDeclaration) && previousMember.GetModifiers().Any(SyntaxKind.ConstKeyword);
+                            bool currentIsConst = member.IsKind(SyntaxKind.FieldDeclaration) && modifiers.Any(SyntaxKind.ConstKeyword);
```
vweijsters: :thought_balloon: It should be possible to cache the `previousIsConst` value, eliminating the need for an extra `previousMember.GetModifiers().Any(SyntaxKind.ConstKeyword)` call for each field.

vweijsters: :bug: if a const field is followed by anything but a field, the compare access level check is disabled.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
```diff
@@ -125,39 +145,79 @@ private static void HandleMemberList(SyntaxNodeAnalysisContext context, SyntaxLi
...
+                            {
+                                compareAccessLevel = false;
+                            }
+
+                            continue;
+
+                        case OrderingTrait.Readonly:
+                            // Only fields may be marked readonly
+                            bool previousIsReadonly = previousMember.IsKind(SyntaxKind.FieldDeclaration) && previousMember.GetModifiers().Any(SyntaxKind.ReadOnlyKeyword);
+                            bool currentIsReadonly = member.IsKind(SyntaxKind.FieldDeclaration) && modifiers.Any(SyntaxKind.ReadOnlyKeyword);
```
vweijsters: :exclamation: See remarks for `OrderingTrait.Constant` above. They apply here as well

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1201ElementsMustAppearInTheCorrectOrder.cs
```diff
@@ -226,6 +248,46 @@ private static void HandleMemberList(SyntaxNodeAnalysisContext context, SyntaxLi
...
+                    default:
+                        continue;
+                    }
+                }
+
+                if (!compareKind)
+                {
+                    continue;
+                }
+
```
vweijsters: :thought_balloon: It would be nice is this could be refactored to use the same  previous / current pattern as SA1202/SA1203/SA1204.

vweijsters:
Finished my review (as you may already have noticed :wink:). There's several issues that should be addressed, but the overall design looks good to me.

@sharwell Big kudos for keeping this well readable / well understandable.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/MemberOrderHelper.cs
sharwell: :arrow_right: Can you file this as a separate bug? For this PR I just reused the code from #1180/#1179. So far it hasn't been reported as a deviation from StyleCop Classic so I'd like to get a bit more discussion to ensure the change is intentional and properly documented.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1214ReadonlyElementsMustAppearBeforeNonReadonlyElements.cs
sharwell: :thought_balloon: Technically, constants are a special kind of `static readonly` field. For the default value of the `elementOrder` property, this doesn't matter. However, it seems like we should treat them as `static readonly` if the `"constants"` ordering is lower precedence or missing from `elementOrder` altogether. That said, currently only SA1214 treats constants as static and readonly. It seems that SA1203 and SA1204 should do so as well.

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1201ElementsMustAppearInTheCorrectOrder.cs
sharwell: I'd rather do this one separately.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/MemberOrderHelper.cs
vweijsters: Done -> #1823

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1201ElementsMustAppearInTheCorrectOrder.cs
vweijsters: no problem

vweijsters:
:+1: Looks good now

martincostello:
@sharwell I've not reviewed the code, but it turns out I accidentally functionally reviewed this earlier by updating a repo of mine to beta017 and using the new rules that were added.

Specifically, the rule found two cases where I have a constant after a field, but I decided to [change the order configuration](https://github.com/martincostello/sqllocaldb/blob/c354786f2e97e194419c7586f4b7d884c38e27fd/src/Tools/StyleCop/stylecop.json#L31-L35) from the default instead instead and leave the code ([1](https://github.com/martincostello/sqllocaldb/blob/master/src/SqlLocalDb/Logger.cs#L24-L32), [2](https://github.com/martincostello/sqllocaldb/blob/master/src/SqlLocalDb/TraceSourceLogger.cs#L24-L32))  as-is.

sharwell:
@martincostello The configuration in your example *is* the default.

martincostello:
@sharwell Turns out I misread [Configuration.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md) and thought that the first JSON example was what the default was, rather than a tweaked version, so I put what I *thought* was the default in explicitly and then changed that to do what I wanted, which actually *was* the default.

sharwell:
It looks like you like to have your configuration be explicit. If that is the case, I recommend pulling your default configuration directly from the [schema](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers/Settings/stylecop.schema.json), which is actually pretty easy since you have a link to the schema as the first line of your configuration file. :smile:

martincostello:
@sharwell I had some caching issues in Visual Studio earlier on where it wouldn't pick up the updated intellisense for the new members, even after I explicitly opened it in the IDE from the schema URL, which is why I was working from the markdown rather than it. Sounds like a good idea for me to do  though!

sharwell:
@martincostello Open up your **stylecop.json** file and right click anywhere in the document. Then select **Reload Schemas**.

martincostello:
Ah...thank you!

sharwell:
That might be a good tip to add to [EnableConfiguration.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/EnableConfiguration.md) if you want to. :wink:

martincostello:
Done - #1850

PR #1813, opened 2015-11-23
SA1623/1624 are now using the AccessLevelHelper

### LINKED ISSUE #1810: Update SA1623 and SA1624 tests to include various accessibilities
Currently the tests for SA1623 and SA1624 only test the public and private accessibility modifiers. The tests should be updated to include others as well.

### LINKED ISSUE #1811: Consider using AccessLevelHelper.GetEffectiveAccessibility for SA1623/SA1624
Currently the implementation of SA1623/SA1624 relies on a locally-implemented method `IsPubliclyVisible`. The helper method `AccessLevelHelper.GetEffectiveAccessibility` implements a more precise and accurate version of this functionality. If possible, the implementation of SA1623/SA1624 should be update to use it.

Marked as a :bug: because the implementation of `IsPubliclyVisible` currently looks for the `private` modifier for classes, when it should probably be looking for the `internal` modifier.

vweijsters:
Fixes #1811
Fixes #1810

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
@@ -245,8 +245,8 @@ namespace StyleCop.Analyzers.Helpers
                 return accessLevel.ToAccessibility();
             }
 
-            BasePropertyDeclarationSyntax propertyDefinition = syntax.FirstAncestorOrSelf<BasePropertyDeclarationSyntax>();
-            return propertyDefinition.GetDeclaredAccessibility(semanticModel, cancellationToken);
+            ISymbol declaredSymbol = semanticModel.GetDeclaredSymbol(syntax, cancellationToken);
+            return declaredSymbol?.DeclaredAccessibility ?? Accessibility.NotApplicable;
         }
 
         internal static Accessibility GetDeclaredAccessibility(this BaseFieldDeclarationSyntax syntax, SemanticModel semanticModel, CancellationToken cancellationToken)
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
```diff
@@ -234,6 +234,21 @@ internal static Accessibility GetDeclaredAccessibility(this BasePropertyDeclarat
...
+            Requires.NotNull(syntax, nameof(syntax));
+            Requires.NotNull(semanticModel, nameof(semanticModel));
+
+            AccessLevel accessLevel = GetAccessLevel(syntax.Modifiers);
+            if (accessLevel != AccessLevel.NotSpecified)
+            {
+                return accessLevel.ToAccessibility();
+            }
+
+            BasePropertyDeclarationSyntax propertyDefinition = syntax.FirstAncestorOrSelf<BasePropertyDeclarationSyntax>();
```
sharwell: :bulb: I would prefer to see `syntax.Parent.Parent` here, which is more efficient than the current check.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/AccessLevelHelper.cs
```diff
@@ -343,6 +358,27 @@ internal static Accessibility GetEffectiveAccessibility(this BasePropertyDeclara
...
+            Requires.NotNull(syntax, nameof(syntax));
+            Requires.NotNull(semanticModel, nameof(semanticModel));
+
+            Accessibility declaredAccessibility = syntax.GetDeclaredAccessibility(semanticModel, cancellationToken);
+            if (declaredAccessibility <= Accessibility.Private)
+            {
+                return declaredAccessibility;
+            }
+
+            BaseTypeDeclarationSyntax enclosingType = syntax.FirstAncestorOrSelf<BaseTypeDeclarationSyntax>();
```
sharwell: :bulb: I would prefer to see `syntax.Parent` here.

sharwell:
:+1:

PR #1817, opened 2015-11-23
Implemented SA1133 (incl. tests + codefix)

### LINKED ISSUE #805: Implement SA1133 (DoNotCombineAttributes)
**Category:** Readability
**Title:** Do not combine attributes
**Description:** C# allows attributes to be combined, such as the following:

```csharp
[Export, Shared]
```

This form makes it difficult to distinguish which attributes are applied to a code element, and which arguments are passed to those attributes. To avoid this situation, only place one attribute within a single set of brackets, such as the following:

```csharp
[Export]
[Shared]
```

**Note:** A rule to ensure only one attribute is placed on a given line will be proposed separately.

vweijsters:
Fixes #805

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
@@ -51,7 +51,7 @@ namespace StyleCop.Analyzers.ReadabilityRules
         private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
         {
             var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
-            var violatingAttribute = (AttributeSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan);
+            var violatingAttribute = (AttributeSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan).Parent;
             var attributeList = (AttributeListSyntax)violatingAttribute.Parent;
             var newAttributeLists = new List<AttributeListSyntax>();
 
@@ -61,7 +61,7 @@ namespace StyleCop.Analyzers.ReadabilityRules
 
             for (var i = 0; i < attributeList.Attributes.Count; i++)
             {
-                var newAttributes = SyntaxFactory.SeparatedList(new[] { attributeList.Attributes[i] });
+                var newAttributes = SyntaxFactory.SingletonSeparatedList(attributeList.Attributes[i]);
                 var newAttributeList = SyntaxFactory.AttributeList(attributeList.Target, newAttributes);
 
                 newAttributeList = (i == 0)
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1133DoNotCombineAttributes.cs
```diff
@@ -32,7 +38,22 @@ internal class SA1133DoNotCombineAttributes : DiagnosticAnalyzer
...
+            context.RegisterSyntaxNodeActionHonorExclusions(HandleAttributeListAction, SyntaxKind.AttributeList);
+        }
+
+        private static void HandleAttributeList(SyntaxNodeAnalysisContext context)
+        {
+            AttributeListSyntax attributeList = (AttributeListSyntax)context.Node;
+
+            if (attributeList.Attributes.Count > 1)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, attributeList.Attributes[1].GetLocation()));
```
sharwell: :bulb: I believe this highlights the entire argument list for the attribute if one is present. I would prefer to see it only mark the (possibly qualified) attribute type name.

sharwell:
:bulb: Consider including a test which includes multiple attribute lists, each of which contains multiple attributes (bonus for trivia placed between them).

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+            var attributeList = (AttributeListSyntax)violatingAttribute.Parent;
+            var newAttributeLists = new List<AttributeListSyntax>();
+
+            var indentationOptions = IndentationOptions.FromDocument(document);
+            var indentationSteps = IndentationHelper.GetIndentationSteps(indentationOptions, attributeList);
+            var indentationTrivia = IndentationHelper.GenerateWhitespaceTrivia(indentationOptions, indentationSteps);
+
+            for (var i = 0; i < attributeList.Attributes.Count; i++)
+            {
+                var newAttributes = SyntaxFactory.SeparatedList(new[] { attributeList.Attributes[i] });
```
Maxwe11: I'm not sure but `SyntaxFactory.SingletonSeparatedList` could be  used ?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+                var newAttributeList = SyntaxFactory.AttributeList(attributeList.Target, newAttributes);
```
pdelvo: You have to make sure you also set the attribute target.

vweijsters: @pdelvo Is that really necessary? I'm not creating new instances, merely adding the existing ones to new lists.

vweijsters:
Processed review comments

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
pdelvo: I overlooked that you actually pass the target. So everything is fine

sharwell:
:+1:

PR #1819, opened 2015-11-24
Allow invocation expressions to span multiple lines

### LINKED ISSUE #1597: Proposal: SA1118 (ParameterMustNotSpanMultipleLines) exception for method invocations
Currently SA1118 documents two exceptions: the first parameter, and any parameter which is an anonymous method.

> To prevent method calls from becoming excessively complicated and unreadable, only the first parameter to a method or indexer call is allowed to span across multiple lines. The exception is an anonymous method passed as a parameter, which is always allowed to span multiple lines.

To improve support for constructing types like `XElement`, I propose adding an additional exception to this rule for method invocations, indexer expressions, and object initializer expressions. Note that the arguments passed to the method, indexer, or constructor must still adhere to this rule.

:memo: This exception would provide us with an easily-documented resolution to #1596.

sharwell:
Fixes #1597

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1118UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1118UnitTests.cs
@@ -91,6 +91,31 @@ class Foo
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


pdelvo:
:+1:

PR #1820, opened 2015-11-24
Add the `newlineAtEndOfFile` configuration property

### LINKED ISSUE #1800: Rule proposal: Code files should (or should not) end with an empty line
StyleCop is all about consistency. Having empty lines at the end of some files and not having it in others break the consistency. That's why I am suggesting this rule. 

**Category:** Layout rules
**Title:** Code files should/shouldn't end with an empty line

Code without an empty line at the end:
```cs
namespace Namespace
{
    public class Class
    {
    }
}
```

Code with an empty line at the end:
```cs
namespace Namespace
{
    public class Class
    {
    }
}


```

This rule is very similar to https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1518.md but 1518 is triggered only when 2 or more end lines are present at the end of the file:

```cs
namespace Namespace
{
    public class Class
    {
    }
}



```

Discussion about ending with an empty line: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

GCC also reports this: http://gcc.gnu.org/ml/gcc/2003-11/msg01568.html

sharwell:
This option allows users to configure the way SA1518 treats newline characters at the end of a file.

Fixes #1800

:memo: The default behavior of SA1518 remains the same as before. However, the location where violations are reported has changed slightly (typically on the line before where they were previously reported) since it greatly simplified the process of making the code fix adhere to the settings file.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/LayoutRules/SA1518CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/LayoutRules/SA1518CodeFixProvider.cs
@@ -5,14 +5,14 @@ namespace StyleCop.Analyzers.LayoutRules
 {
     using System.Collections.Immutable;
     using System.Composition;
-    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
     using Helpers;
     using Microsoft.CodeAnalysis;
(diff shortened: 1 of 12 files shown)


pdelvo:
:+1:

vweijsters:
:+1:

PR #1825, opened 2015-11-26
Fix NullReferenceException in SA1623/SA1624

sharwell:
Expression-bodied properties do not have accessor lists.


oatkins:
Good catch.

Code looks good to me :+1:

PR #1835, opened 2015-11-26
Fix SA1003 code fix's handling of trivia

### LINKED ISSUE #1831: SA1003 fix not working for conditional directive
```cs
class Program
{
    static int Main(string[] args)
    {
#if! NOT
        return 1;
#endif
    }
}
```

When applying the fix at line 5 it adds a space right before 'return'.

sharwell:
Fixes #1831


Maxwe11:
:+1:

PR #1836, opened 2015-11-26
Fix handling of tabs inside of documentation comments

### LINKED ISSUE #1830: SA1027 does not detect tabs in xml documentation
Repro:
```csharp
///[Tab]Foo
class Program { }
```

This should report SA1027. It does not at the moment.

sharwell:
Fixes #1830


pdelvo:
What happens if there is a sequence with mixed tabs/whitespaces? Does this work correctly? It might be good to include a test case.

sharwell:
@pdelvo Updated!

pdelvo:
:+1:

PR #1837, opened 2015-11-26
Fix handling of filenames with varying suffixes

### LINKED ISSUE #1829: SA1649 - File name must match first type name - false positive error
This rule mistakenly fires for files with double postfix names, i.e. 

File "AppEventReceiver.svc.cs"

That contains class:
	public class AppEventReceiver : AppEventReceiverBase

Will fire the rule. It shouldn't and it doesn't in the original stylecop tool. .

Fix should be fairly easy by using the filename "before the first dot" instead of "filename without extension" for matching the class name. 

And thanks for a great tool :-)

sharwell:
Fixes #1829


pdelvo:
:+1:

PR #1839, opened 2015-11-26
Enable SA1133 by default

### LINKED ISSUE #1827: SA1133 should be enabled by default
This rule was implemented in #1817, but we forgot to enable it by default.

sharwell:
Fixes #1827


Maxwe11:
:+1:

PR #1840, opened 2015-11-26
Fix SA1009 behavior inside trivia

### LINKED ISSUE #1834: SA1009 not reported inside conditional directives
I expected SA1009 to be reported in the following, but no spacing diagnostics were reported.

```csharp
class Program
{
    static int Main(string[] args)
    {
#if !(X || NOT )
        return 1;
#endif
    }
}
```

:memo: Discovered while working on #1831.

sharwell:
Fixes #1834


pdelvo:
:+1:

PR #1845, opened 2015-11-26
Fix handling of property accessor visibility

### LINKED ISSUE #1844: SA1623 - Property summary documentation in private class
After updating to Beta 17 I am getting SA1623 reported for the following scenario (note the class is private):

    private class TestClass
    {
        /// <summary>
        /// Gets the name
        /// </summary>
        public string Name { get; private set; }
    }

It thinks that the summary should start with "Gets or sets". The analyzer gets confused because the effective accessibility of the getter/setter is determined to be private due to the fact that is in contained in a private class.
I started trying to fix this but it's rather tricky. It's a low priority for me because we very rarely use private classes.

For what it's worth, I think the following unit test should pass.

        [Theory]
        [InlineData("public", "int", "{ get; set; }", "Gets or sets")]
        [InlineData("public", "int", "{ get; protected set; }", "Gets or sets")]
        [InlineData("public", "int", "{ get; protected internal set; }", "Gets or sets")]
        [InlineData("public", "int", "{ get; internal set; }", "Gets")]
        [InlineData("public", "int", "{ get; private set; }", "Gets")]
        [InlineData("public", "int", "{ get; }", "Gets")]
        [InlineData("public", "int", "{ get; } = 0;", "Gets")]
        [InlineData("public", "int", "=> 0;", "Gets")]
        [InlineData("public", "int", "{ set { } }", "Sets")]
        [InlineData("public", "int", "{ internal get { return 0; } set { } }", "Sets")]
        [InlineData("public", "int", "{ private get { return 0; } set { } }", "Sets")]
        [InlineData("public", "bool", "{ get; set; }", "Gets or sets a value indicating whether")]
        [InlineData("public", "bool", "{ get; }", "Gets a value indicating whether")]
        [InlineData("public", "bool", "{ get; } = true;", "Gets a value indicating whether")]
        [InlineData("public", "bool", "=> true;", "Gets a value indicating whether")]
        [InlineData("public", "bool", "{ get; private set; }", "Gets a value indicating whether")]
        [InlineData("public", "bool", "{ set { } }", "Sets a value indicating whether")]
        [InlineData("public", "bool", "{ private get { return false; } set { } }", "Sets a value indicating whether")]
        [InlineData("protected", "int", "{ get; private set; }", "Gets")]
        [InlineData("protected", "int", "{ private get { return 0; } set { } }", "Sets")]
        [InlineData("protected internal", "int", "{ get; internal set; }", "Gets")]
        [InlineData("protected internal", "int", "{ get; private set; }", "Gets")]
        [InlineData("protected internal", "int", "{ internal get { return 0; } set { } }", "Sets")]
        [InlineData("protected internal", "int", "{ private get { return 0; } set { } }", "Sets")]
        [InlineData("internal", "int", "{ get; private set; }", "Gets")]
        [InlineData("internal", "int", "{ private get { return 0; } set { } }", "Sets")]
        public async Task VerifyDocumentationOfPublicMethodInPrivateClassWillProduceDiagnosticAsync(string accessibility, string type, string accessors, string expectedArgument)
        {
            var testCode = $@"
    public class TestClass
    {{
        private class PrivateTestClass
        {{
            /// <summary>
            /// The first test value.
            /// </summary>
            {accessibility} {type} TestProperty {accessors}
        }}
    }}
    ";

            var fixedTestCode = $@"
    public class TestClass
    {{
        private class PrivateTestClass
        {{
            /// <summary>
            /// {expectedArgument} the first test value.
            /// </summary>
            {accessibility} {type} TestProperty {accessors}
        }}
    }}
    ";

            var expected = this.CSharpDiagnostic(PropertySummaryDocumentationAnalyzer.SA1623Descriptor).WithLocation(9, 11 + accessibility.Length + type.Length).WithArguments(expectedArgument);

            await this.VerifyCSharpDiagnosticAsync(fixedTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
            await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
        }

sharwell:
Fixes #1844 

To make this easier to review in the context of the original bug report, the commits are as follows:

1. First commit uses the originally proposed set of tests from #1844, without changes.
2. Second commit updates the implementation to follow the documented rules for SA1623 and SA1624.
3. Third commit corrects test cases which failed after updating the implementation. This shows the difference between the actual implementation here, the original tests, and issue #1844.


vweijsters:
:+1: Looks good to me

PR #1846, opened 2015-11-27
Improve messaging for SA1205

### LINKED ISSUE #1778: Code triggers SA1202 instead of SA1400
First reported in #1766, Issue D9, with some refinements here.

This is a difference with classic StyleCop. In classic StyleCop the code below would trigger SA1400 and SA1205. In the current analyzer it triggers SA1202 and SA1205. It boils down to the fact that the partial classes without an access modifier may appear before or after the public class, depending on the access modifier assigned. Classic StyleCop suppressed the SA1202 and just reported SA1205 (and SA1400). The current analyzer suppresses SA1400 in favor of SA1205, but classic StyleCop reported both.

I'd be okay with just documenting this functionality difference with classic StyleCop. This is not a problem that will come up frequently, and I think it was really a bug in classic StyleCop that they reported both SA1400 and SA1205. The current functionality is actually more correct than standard StyleCop.

```csharp
namespace RuleViolations
{
    /// <summary>A rule SA1205 fail.</summary>
    public class RuleSA1205Fail
    {
        /// <content>A part 1 class.</content>
        partial class Part1
        {
        }

        /// <summary>A part 2 class.</summary>
        public class Part2
        {
        }
    }
}
```

sharwell:
Closes #1778

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1205PartialElementsMustDeclareAccess.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1205PartialElementsMustDeclareAccess.cs
@@ -57,19 +57,14 @@ namespace StyleCop.Analyzers.OrderingRules
         {
             var typeDeclarationNode = (TypeDeclarationSyntax)context.Node;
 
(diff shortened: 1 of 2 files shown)


pdelvo:
:+1:

PR #1848, opened 2015-11-27
Implement SA1625

### LINKED ISSUE #145: SA1625: ElementDocumentationMustNotBeCopiedAndPasted
http://www.stylecop.com/docs/SA1625.html

pdelvo:
Fixes #145.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
@@ -71,7 +71,7 @@ namespace StyleCop.Analyzers.DocumentationRules
         private const string HelpLink = "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1625.md";
 
         private static readonly DiagnosticDescriptor Descriptor =
-            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, AnalyzerCategory.DocumentationRules, DiagnosticSeverity.Warning, AnalyzerConstants.DisabledNoTests, Description, HelpLink);
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, AnalyzerCategory.DocumentationRules, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
 
         /// <inheritdoc/>
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } =
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
@@ -3,8 +3,12 @@
 
 namespace StyleCop.Analyzers.DocumentationRules
 {
+    using System.Collections.Generic;
     using System.Collections.Immutable;
+    using Helpers;
     using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
```diff
@@ -75,7 +80,40 @@ internal class SA1625ElementDocumentationMustNotBeCopiedAndPasted : DiagnosticAn
...
+        {
+            context.RegisterSyntaxNodeActionHonorExclusions(HandleDocumentationTrivia, SyntaxKind.SingleLineDocumentationCommentTrivia);
+            context.RegisterSyntaxNodeActionHonorExclusions(HandleDocumentationTrivia, SyntaxKind.MultiLineDocumentationCommentTrivia);
+        }
+
+        private static void HandleDocumentationTrivia(SyntaxNodeAnalysisContext context)
+        {
+            DocumentationCommentTriviaSyntax syntax = context.Node as DocumentationCommentTriviaSyntax;
+
+            HashSet<string> documentationTexts = new HashSet<string>();
```
sharwell: :question: Should we disable this rule by default? It looks like it will inevitably be expensive. What's the rate of positive hits on a project like Roslyn or Code Contracts (I'll test this and reply back)?

/cc @vweijsters @AArnott

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
```diff
@@ -75,7 +80,40 @@ internal class SA1625ElementDocumentationMustNotBeCopiedAndPasted : DiagnosticAn
...
+
+            foreach (var content in syntax.Content)
+            {
+                string text = XmlCommentHelper.GetText(content, true).Trim();
```
sharwell: :exclamation: I'm getting a `NullReferenceException` on this line when I run StyleCopTester on StyleCopAnalyzers.sln.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
```diff
@@ -75,7 +80,40 @@ internal class SA1625ElementDocumentationMustNotBeCopiedAndPasted : DiagnosticAn
         /// <inheritdoc/>
         public override void Initialize(AnalysisContext context)
         {
-            // TODO: Implement analysis
+            context.RegisterCompilationStartAction(HandleCompilationStart);
```
sharwell: :bulb: This will need to be updated to match the `CompilationStartAction` cached delegate pattern used in all the other analyzers.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
```diff
@@ -75,7 +80,40 @@ internal class SA1625ElementDocumentationMustNotBeCopiedAndPasted : DiagnosticAn
...
+        }
+
+        private static void HandleCompilationStart(CompilationStartAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionHonorExclusions(HandleDocumentationTrivia, SyntaxKind.SingleLineDocumentationCommentTrivia);
```
sharwell: :bulb: This will need to be updated to match the cached delegate pattern used in all the other analyzers.

1. Cache the delegate in `DocumentationCommentTriviaAction`
1. Cache the immutable array containing both syntax kinds
1. Only call `RegisterSyntaxNodeActionHonorExclusions` once, and pass in the cached delegate and immutable array

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1625ElementDocumentationMustNotBeCopiedAndPasted.cs
sharwell: | Project | Violation count |
| --- | --- |
| StyleCopAnalyzers.sln | 0 |
| Roslyn.sln | 6 |
| CodeContracts.sln | 4 |

AArnott: I haven't studied the PR trends enough to recognize that this is particularly more expensive than average. Is it that we create a HashSet<T> and we're concerned about GC pressure, or is it CPU time that you expect to suffer?
We could mitigate the HashSet<T> GC pressure by creating a thread-safe pool of HashSet<T> objects that your analyzers can 'borrow' from.

pdelvo:
I implemented your feedback

AArnott:
Nice. :smile:

PR #1849, opened 2015-11-27
Update SA1302 code fix to support conflict resolution

### LINKED ISSUE #436: Update SA1302 (InterfaceNamesMustBeginWithI) to handle conflicting names

sharwell:
Fixes #436

Edit: ~~Need to update this to ensure a nested type isn't renamed to the same name as the enclosing type.~~ This is now working.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1302UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1302UnitTests.cs
@@ -185,6 +185,56 @@ public class MyNativeMethods
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
         }
 
+        [Fact]
+        public async Task TestInterfaceDeclarationDoesNotStartWithIWithConflictAsync()
+        {
+            string testCode = @"
+public interface Foo
+{
+}
+
+public interface IFoo { }";
+            string fixedCode = @"
+public interface IFoo1
+{
+}
+
+public interface IFoo { }";
+
+            DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(2, 18);
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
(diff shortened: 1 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1302UnitTests.cs
```diff
@@ -185,6 +185,160 @@ public interface FileOpenDialog111
...
+{
+    int IFoo { get; }
+}";
+
+            DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(2, 18);
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
+        }
```
vweijsters: :question:  Why is the interface renamed to `IFoo1` here? There clearly is no conflict.

vweijsters:
:+1: Looks good to me, besides the question about a testcase.

StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1302UnitTests.cs
sharwell: :arrow_right: It contains a member named `IFoo`. Members are not allowed to have the same name as the containing type, so either we don't use the name `IFoo` for the interface, or we have to rename both the interface and the property.

vweijsters: Members can have the same name as the containing type, there is nothing preventing that. The compiler accepts it, and so does StyleCop Analyzers. Its not desirable, but that should the subject for a new diagnostic imho.

sharwell: Looks like this is partially correct. If the containing type is a `class` or `struct`, you get CS0542. If the containing type is an `interface` or `enum`, you don't get an error.

PR #1850, opened 2015-11-27
Add hint about how to refresh the JSON schema

martincostello:
Adds a hint about how to refresh the schema if it goes out of date in Visual Studio as suggested [here](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1812#issuecomment-160184753).


sharwell:
:+1:

PR #1855, opened 2015-11-27
Implemented SA1134 (incl. tests + code fix)

### LINKED ISSUE #809: Implement SA1134 (AttributesMustNotShareLine)
**Category:** ~~Layout~~ Readability
**Title:** Attribute must be on separate line
**Description:** C# allows attributes to be specified on the same line of code, like:
```C#
[Export] [Shared]
public class Foo
{
}
```
The form makes it more difficult to determine which attributes are present, especially when the attributes also have a (long) list op parameters. Each attribute should therefore be placed on its own line. 
```C#
[Export]
[Shared]
public class Foo
{
}
```

An exception to this rule would be inline attributes for parameters.

vweijsters:
Fixes #809


NikolayIT:
May be a test with attributes that are located above a namespace definition will be a good idea (`[assemly:...]` for example) :)

sharwell:
Also tests for attributes on generic type parameters.

vweijsters:
Updated after CR

PR #1858, opened 2015-11-28
Improve IsValidNewMemberName handling of parameters and locals

### LINKED ISSUE #1315: SA1306CodeFixProvider does not support name conflict resolution
In #1316, the code fix currently used for SA1306 was generalized to `RenameToLowerCaseCodeFixProvider`. Currently this operation does not support name conflict resolution, which could lead to broken code following a fix or fix-all operation.

sharwell:
Fixes #1315


vweijsters:
:+1: Looks good to me. 

There is a possible interference with SX1309(S), but I've created #1861 for that.

PR #1864, opened 2015-11-29
Update SA1312 to check additional local variable declarations

### LINKED ISSUE #1856: SA1312 needs to use a symbol visitor
There are many different ways to declare a local variable. SA1312 needs to use a symbol visitor for `ILocalSymbol` (as opposed to the current syntax visitor) to consistently identify these for analysis.

sharwell:
Fixes #1856 

:memo: Unfortunately it's not possible to just register a symbol action for `ILocalSymbol`, so instead this pull request operates on the additional types of syntax nodes which declare local variables.


vweijsters:
:+1:

PR #1865, opened 2015-11-29
Fix UsingCodeFixProvider handling of aliases

### LINKED ISSUE #1770: UsingCodeFixProvider does not handle aliases correctly
Currently UsingCodeFixProvider can merge two using directives which do not use the same alias. The correct behavior is to allow merging a using directive which does not have an alias with any using directives which have the alias `global::`. Using directives with any other alias can only be merged with other using directives that have exactly the same alias.

sharwell:
This change can result in the output of this code fix containing duplicate
using directives, but this situation does not break the semantics of code
and only produces a build warning.

Fixes #1770


vweijsters:
:+1:

PR #1866, opened 2015-11-29
SA1306 now ignores leading underscores.

### LINKED ISSUE #1861: SA1306 should also check fields starting with a single underscore
The following code should trigger SA1306 in my opinion (with SA1309 disabled and SX1309 enabled).

```csharp
public class TestClass
{
    private int _TestField;
}
```

I would propose that we make the underscore for fields a property and that we adapt the different analyzers and code fixes to take this setting into account properly. I think with the correct set of properties we can wrap up SA1309, SX1309 and SX1309S into a single analyzer.

vweijsters:
Fixes #1861


NikolayIT:
I would add these tests: :)

* `_`
* `___`
* `___someVar`
* `someVar_`

vweijsters:
Added additional test cases.

vweijsters:
Fixed failing test.

sharwell:
:+1:

PR #1880, opened 2015-12-01
Fix #1878 (SA1133CodeFixProvider crash with System.InvalidCastException)

### LINKED ISSUE #1878: SA1133CodeFixProvider crash
Given the following code :

```csharp
namespace Stylecop_rc1_bug_repro
{
    using System;

    internal class Program
    {
        [Foo, Bar]
        private static void Main(string[] args)
        {
        }
    }

    internal class FooAttribute : Attribute
    {
    }

    internal class BarAttribute : Attribute
    {
    }
}
```

SA1133 is raised on `Bar` usage but activating the CodeFix throw an exception :

```text
Exception thrown: 'System.InvalidCastException' in StyleCop.Analyzers.CodeFixes.dll

Additional information: Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.AttributeListSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.AttributeSyntax'.

   at StyleCop.Analyzers.ReadabilityRules.SA1133CodeFixProvider.<GetTransformedDocumentAsync>d__5.MoveNext()
```

(*VS2015 update 1, StyleCop 1.0 RC1*)

NikolayIT:
Fix #1878

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
@@ -51,8 +51,18 @@ namespace StyleCop.Analyzers.ReadabilityRules
         private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
         {
             var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
-            var violatingAttribute = (AttributeSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan).Parent;
-            var attributeList = (AttributeListSyntax)violatingAttribute.Parent;
+            var nodeInSourceSpan = syntaxRoot.FindNode(diagnostic.Location.SourceSpan);
+            AttributeListSyntax attributeList;
+            if (nodeInSourceSpan.Parent is AttributeListSyntax)
+            {
+                attributeList = (AttributeListSyntax)nodeInSourceSpan.Parent;
+            }
+            else
+            {
+                var violatingAttribute = (AttributeSyntax)nodeInSourceSpan.Parent;
+                attributeList = (AttributeListSyntax)violatingAttribute.Parent;
+            }
+
             var newAttributeLists = new List<AttributeListSyntax>();
 
             var indentationOptions = IndentationOptions.FromDocument(document);
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
@@ -18,6 +18,10 @@ namespace StyleCop.Analyzers.ReadabilityRules
     /// <summary>
     /// Implements a code fix for <see cref="SA1133DoNotCombineAttributes"/>.
     /// </summary>
+    /// <remarks>
+    /// The SA1133 code fix adds the new lines to make sure that it doesn't immediately introduces a SA1134 after code fixing,
+    /// but it will not / should not attempt to fix any preexisting SA1134 cases.
+    /// </remarks>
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
```diff
@@ -51,8 +51,18 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
         private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
         {
             var syntaxRoot = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
-            var violatingAttribute = (AttributeSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan).Parent;
-            var attributeList = (AttributeListSyntax)violatingAttribute.Parent;
+            var nodeInSourceSpan = syntaxRoot.FindNode(diagnostic.Location.SourceSpan);
```
sharwell: We've run into problems in the past where error nodes in the tree caused `FindNode` to return unexpected nodes. Normally this is solved by setting `getInnermostNodeForTie: true`, but I'm not sure that would behave properly here. We probably want to have a third case handled where `nodeInSourceSpan.Parent` is neither an `AttributeListSyntax` nor an `AttributeSyntax`, and just return.

NikolayIT:
@sharwell added default case which just returns the document object

Noryoko:
Could you add these tests?

```
[Foo]
[Bar, Car]

[Foo, Bar]
[Car]

[Foo, Bar, Car]

[Foo]
[Bar, Car]
[Ear]
```

NikolayIT:
@Noryoko unit tests added.

NikolayIT:
@Noryoko, @vweijsters, do you think the following tests should also pass?

    [InlineData("[Foo][Bar, Car]", "[Foo]\r\n[Bar]\r\n[Car]", 2, 12)]
    [InlineData("[Foo, Bar][Car]", "[Foo]\r\n[Bar]\r\n[Car]", 2, 7)]
    [InlineData("[Foo][Bar, Car][Ear]", "[Foo]\r\n[Bar]\r\n[Car]\r\n[Ear]", 2, 12)]

With the current code they are failing.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1133CodeFixProvider.cs
vweijsters: :exclamation:  A better solution would be to use `getInnermostNodeForTie: true` and then call `FirstAncestorOrSelf<AttributeListSyntax>()` on the found node. That call should never fail, as the analyzer works on `AttributeSyntaxList`.

Noryoko:
I think so. The fix is already splitting the attributes onto their own lines. Would make the end result look consistent.

vweijsters:
I don't think they should pass, as they are partly in the domain of SA1134. The SA1133 code fix adds the new lines to make sure that it doesn't immediately introduces a SA1134 after code fixing, but it will not / should not attempt to fix any preexisting SA1134 cases.

NikolayIT:
OK, then this pull request is ready for approval.

Noryoko:
I'll change my vote to agree with @vweijsters.

vweijsters:
@NikolayIT I would like to see my remark on `SA1133CodeFixProvider.cs` to be addressed before I'm ok with this pull request.

NikolayIT:
@vweijsters is that what you meant?

vweijsters:
No, I meant:
> :exclamation: A better solution would be to use getInnermostNodeForTie: true and then call FirstAncestorOrSelf<AttributeListSyntax>() on the found node. That call should never fail, as the analyzer works on AttributeSyntaxList.

NikolayIT:
Sorry, didn't see it. Is now OK? @vweijsters

vweijsters:
Totally :wink: 

:+1: Looks good to me

Noryoko:
:+1: Looks good to me

vweijsters:
@sharwell Could you merge this into master as well?

sharwell:
@vweijsters yes I'll merge it tonight

PR #1887, opened 2015-12-02
Fix for bug in SA1115 (Parameter must begin on next line fix) #1886

### LINKED ISSUE #1886: SA1115 (Parameter must follow comma) generates diagnostic when parameter list contains #pragma directives
The following code generates a SA1115 diagnostic.

    public void Baz()
    {
        Bar(
    #pragma warning disable CS4014
            1,
    #pragma warning restore CS4014
            2);
    }

It shouldn't.

(The suppression here doesn't make sense, but it's common to use this particular suppression when intentionally not awaiting an async method.)

oatkins:
Fixes #1886.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1115ParameterMustFollowComma.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1115ParameterMustFollowComma.cs
@@ -270,18 +270,32 @@ namespace StyleCop.Analyzers.ReadabilityRules
                 return;
             }
 
-            var previousLine = argumentListSyntax.Arguments[0].GetLineSpan().EndLinePosition.Line;
+            var previousArgumentLine = argumentListSyntax.Arguments[0].GetLineSpan().EndLinePosition.Line;
             for (int i = 1; i < argumentListSyntax.Arguments.Count; i++)
             {
                 var currentArgument = argumentListSyntax.Arguments[i];
-                var lineSpan = currentArgument.GetLineSpan();
-                var currentLine = lineSpan.StartLinePosition.Line;
-                if (currentLine - previousLine > 1)
+                int currentArgumentStartLine;
+                int currentArgumentEndLine;
+
+                if (currentArgument.HasLeadingTrivia && currentArgument.GetLeadingTrivia().All(trivia => IsValidTrivia(trivia)))
+                {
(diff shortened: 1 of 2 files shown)


NikolayIT:
:+1: looks good to me

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1115ParameterMustFollowComma.cs
```diff
@@ -327,6 +341,7 @@ private static bool IsValidTrivia(SyntaxTrivia trivia)
             case SyntaxKind.EndIfDirectiveTrivia:
             case SyntaxKind.DisabledTextTrivia:
             case SyntaxKind.WhitespaceTrivia:
+            case SyntaxKind.PragmaWarningDirectiveTrivia:
```
sharwell: :question: Is there any directive trivia we wouldn't want to consider here?

```csharp
if (trivia.IsDirective)
{
    return true;
}

switch (trivia.Kind())
{
...
}
```

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1115ParameterMustFollowComma.cs
```diff
@@ -319,12 +333,13 @@ private static void AnalyzeParameterList(SyntaxNodeAnalysisContext context, Base
 
         private static bool IsValidTrivia(SyntaxTrivia trivia)
         {
+            if (trivia.IsDirective)
+            {
+                return true;
+            }
```
vweijsters: This check is probably not correct, as it will also mark `#region` / `#endregion` as valid, but I think they are not.

vweijsters: It would be nice to have a test case for that as well.

sharwell: :thought_balloon:  Those would still be reported as SA1123. I don't see any need to make a special consideration for it here as well.

vweijsters:
I think this one is blocked until a decision has been made on how to proceeded with the other issues that I mentioned in #1886.

vweijsters:
:+1:

PR #1902, opened 2015-12-04
SA1133 now longer considers attributes on parameters and type parameters

### LINKED ISSUE #1882: SA1133 on P/Invoke method parameters
Sample code:
```
#pragma warning disable SA1600
#pragma warning disable SA1649
#pragma warning disable SA1652

namespace SA1133Crash
{
    using System;
    using System.Runtime.InteropServices;

    internal static class NativeMethods
    {
        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
        public static extern IntPtr CreateJobObject([In, Optional] IntPtr security_attributes, string lpName);
    }
}
```

Of particular concern is this `[In, Optional] IntPtr security_attributes,`.

1. Is this a valid violation of SA1133? How is `[In] [Optional]` any better than `[In, Optional]`? I agree for attributes on classes, methods and properties that they should be separate and on separate lines, but not neccesarily for parameters.

2. Attempting to use the code fix throws this exception, which is likely a duplicate of #1878:

```
System.InvalidCastException occurred
Message: Exception thrown: 'System.InvalidCastException' in StyleCop.Analyzers.CodeFixes.dll
Additional information: Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.AttributeListSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.AttributeSyntax'.
```

Using VS2015 Update 1 (RTM), StyleCop.Analyzers 1.0.0-RC

vweijsters:
Fixes #1882

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1133UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1133UnitTests.cs
@@ -276,6 +276,36 @@ public class Ear : Attribute
             await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Verifies that attribute list with multiple attributes for (generic) parameters will not produce diagnostics.
+        /// Regression test for #1882
+        /// </summary>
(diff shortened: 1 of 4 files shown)


sharwell:
:+1:

pdelvo:
:+1:

PR #1925, opened 2015-12-06
Implement SA1314 (TypeParameterNamesMustBeginWithT)

### LINKED ISSUE #739: Implement SA1314: TypeParameterNamesMustBeginWithT
I'd like to propose new naming rule for generic parameters. Name should start with "T", second character (if it exists) should be upper case letter or digit. AFAIK Whole BCL follows this convention. 

Few valid names: "T", "T1", "TKey", "TValue" 

Few invalid names: "t", "Tkey", "Key" 

This rule was originally proposed by Andy Reeves here:
http://stylecop.codeplex.com/workitem/6891

nbarbettini:
Fixes #739. Includes code fix, tests, and documentation.

Let me know if there's any test scenarios I didn't think of!


sharwell:
:exclamation: naming rules have IDs with the form SA13xx.

vweijsters:
Type parameters would be preferred for me

nbarbettini:
I wasn't sure myself; I'd be totally fine with changing it to "type parameters" across the board.

nbarbettini:
Fixed both the ID and the rule name. :+1:

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/RenameHelper.cs
```diff
@@ -52,6 +52,20 @@ internal static class RenameHelper
 
             var containingSymbol = symbol.ContainingSymbol;
 
+            if (symbol.Kind == SymbolKind.TypeParameter)
```
pdelvo: We cant rename the type parameter if there is already a type with the same name

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/SA1314CodeFixProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+                    diagnostic);
+            }
+
+            return SpecializedTasks.CompletedTask;
+        }
+
+        private static async Task<Solution> CreateChangedSolutionAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        {
+            var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+            var token = root.FindToken(diagnostic.Location.SourceSpan.Start);
```
pdelvo: Do we also want to make the first letter after T capital?

StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1314UnitTests.cs
```diff
@@ -0,0 +1,221 @@
...
+        var key = typeof(TKey);
+    }
+}
+";
+
+            await this.VerifyCSharpFixAsync(testCode, fixedCode).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestGenericParameterStartsWithLowerTAsync()
```
pdelvo: Do we want to be smarter in this case?

pdelvo:
Maybe we should include a test to verify that this also works for type parameters in method signatures

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/RenameHelper.cs
nbarbettini: You mean (for example) if there was already a class named `TKey`?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/SA1314CodeFixProvider.cs
nbarbettini: I followed the example of SA1302, which simply adds an I and doesn't touch the rest of the token. This makes sense, though; I could implement this change if others agree.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/RenameHelper.cs
pdelvo: Exactly

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/SA1314CodeFixProvider.cs
sharwell: I don't mind just adding the `T` prefix.

nbarbettini:
@pdelvo Added a test case for type parameters in method signatures, and also refactored the logic in `RenameHelper` to handle conflicts with other types.

PR #1954, opened 2015-12-12
Fix SA1606 analyzer crashes with incomplete member

### LINKED ISSUE #1937: RC2: NullReferenceException in SA1606
After upgrading to RC2 from Beta15, I'm seeing the following exception occur frequently:

```
Analyzer 'StyleCop.Analyzers.DocumentationRules.SA1606ElementDocumentationMustHaveSummaryText' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null.
Parameter name: s'
```

If there's anything I can do to capture & provide additional information, let me know

Noryoko:
- [x] Add regression test for #1937
 - [x] Fixes #1937


vweijsters:
:+1: Looks good to me

sharwell:
Merged into **master** instead of **stabilization**.

PR #1957, opened 2015-12-12
Fix crash when running SA1623 code fix

### LINKED ISSUE #1943: SA1623 code fix throws exception for empty summary
The code fix for SA1623 throws an exception when attempting to fix the following code:

```csharp
public class ClassName
{
    /// <summary></summary>
    public int Property
    {
        get;
    }
}
```

NikolayIT:
Fix #1943

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1623UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1623UnitTests.cs
@@ -152,6 +152,39 @@ public class TestClass
             await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Verifies that empty summary tag does not throw an exception.
+        /// Regression test for #1943
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task EmptySummaryTagShouldNotThrowAnExceptionAsync()
+        {
+            var testCode = @"public class ClassName
+{
+    /// <summary></summary>
+    public int Property
+    {
+        get;
+    }
+}";
+
+            var fixedTestCode = @"public class ClassName
+{
+    /// <summary>Gets Property</summary>
+    public int Property
+    {
+        get;
+    }
+}";
+
+            var expected = this.CSharpDiagnostic(PropertySummaryDocumentationAnalyzer.SA1623Descriptor).WithLocation(4, 16).WithArguments("Gets");
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
+        }
+
         /// <inheritdoc/>
         protected override CodeFixProvider GetCSharpCodeFixProvider()
         {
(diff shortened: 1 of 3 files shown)


NikolayIT:
We have to discuss what should the default text of the code fix for SA1623 be.

Currently it is always `<summary>Gets the value</summary>`.

I am having a suggestion to replace it with `<summary>Gets PropertyName</summary>`.

What to you think?

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1623UnitTests.cs
```diff
@@ -152,6 +152,39 @@ private class PrivateTestClass
...
+    /// <summary></summary>
+    public int Property
+    {
+        get;
+    }
+}";
+
+            var fixedTestCode = @"public class ClassName
+{
+    /// <summary>Gets the value</summary>
```
sharwell: :exclamation: This suppresses the warning but does not actually fix the documentation problem. For other similar situations, we intentionally do not provide a code fix, and I think we should continue to follow that model here.

NikolayIT: OK, tomorrow I will remove the default summary and will leave it empty.

NikolayIT:
@sharwell, done.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1623UnitTests.cs
```diff
@@ -152,6 +152,33 @@ private class PrivateTestClass
...
+}";
+
+            var expected = this.CSharpDiagnostic(PropertySummaryDocumentationAnalyzer.SA1623Descriptor).WithLocation(4, 16).WithArguments("Gets");
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+
+            // Intentionally do not provide a code fix: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1957#discussion_r47703520
+            //// await this.VerifyCSharpDiagnosticAsync(fixedTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+
+            await this.VerifyCSharpFixAsync(testCode, testCode).ConfigureAwait(false);
```
vweijsters: :thought_balloon: This should probably use:
```csharp
var offeredFixes = await this.GetOfferedCSharpFixesAsync(testCode).ConfigureAwait(false);
Assert.Empty(offeredFixes);
```

vweijsters:
Currently a code fix is offered that will do nothing. I would prefer to see it changed so that the code fix is not offered at all. It will be more clear for the user that the diagnostic cannot be fixed by us.

sharwell:
> Currently a code fix is offered that will do nothing. I would prefer to see it changed so that the code fix is not offered at all. It will be more clear for the user that the diagnostic cannot be fixed by us.

We could probably do this efficiently by adding a property to the reported diagnostic in the case where the summary is empty. That way the code fix can simply skip over diagnostics that have the property without performing any additional analysis.

NikolayIT:
So what should I do in order to match your CR? :smile:

vweijsters:
@NikolayIT Have a look at the SA1003 analyzer and code fix. They implement a similar mechanism. 

In this case you can simplify stuff a bit. The presence of the property in the set can be used to check if the code fix should be offered or not.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1623UnitTests.cs
sharwell: Interesting, I didn't even know we had that. :smile:

vweijsters:
:+1:

sharwell:
:+1:

PR #1959, opened 2015-12-13
Remove inheritdoc analysis from SA1606

### LINKED ISSUE #1944: SA1606 inheritdoc analysis does not consider overrides
The following code should report SA1606, but no warning is reported.

```csharp
/// <summary>
/// </summary>
/// <inheritdoc/>
public string Property => "P";
```

sharwell:
Fixes #1944


vweijsters:
:+1: Looks good to me

PR #1960, opened 2015-12-13
Do not report SA1648 for inheritdoc elements with a cref attribute

### LINKED ISSUE #1948: SA1648 should always allow <inheritdoc/> with a cref attribute
Based on the documentation for [use of `<inheritdoc/>`](http://ewsoftware.github.io/XMLCommentsGuide/html/86453FFB-B978-4A2A-9EB5-70E118CA8073.htm), SA1648 should never be reported for the use of `<inheritdoc cref="..."/>`.

:memo: The target of the `cref` attribute does not need to be checked because the compiler will report CS1574 if it is not correct.

sharwell:
Fixes #1948


vweijsters:
:+1: Looks good to me

PR #1962, opened 2015-12-14
SA1303 should not report on enum members

### LINKED ISSUE #1936: RC2: SA1300 and SA1303 raised for enum values starting with a lower case letter
Consider the following code:

```
public enum SpecialFile
{
    iTunesMetadata
}
```

This raises the following warnings:
1. SA1300	Element 'iTunesMetadata' must begin with an uppercase letter
2. SA1303	Const field names must begin with upper-case letter.

Essentially, duplicating the same warning. In my case, this now requires two source code warnings. One ought to be enough.

nbarbettini:
Closes #1936


sharwell:
:+1:

Issue #1614, opened 2015-10-11
Reference data for churn testing


sharwell:
We should compile reference results from running the analyzers and batch code fixes over large reference projects to detect subtle behavior changes over time. 

We can expect some changes as the code fixes are improved, but we want to make sure these changes don't negatively impact the overall experience in edge cases.

PR #1968, opened 2015-12-17
Fix StyleCopTester not running disabled rules even with /id

sharwell:
:memo: This change only affects StyleCopTester, and is a prerequisite for #1614. It will not require a new RC release and will not reset the release time frame.


pdelvo:
:+1:

PR #1969, opened 2015-12-18
Output test results

sharwell:
Add the `/log` argument to StyleCopTester.

Add a script which analyzes a particular repository using both Roslyn 1.0.0 and Roslyn 1.1.0.


vweijsters:
:question: This seems identical to #1970. Am I missing something?

sharwell:
Superseded by #1970.

PR #1970, opened 2015-12-18
Churn testing

### LINKED ISSUE #1614: Reference data for churn testing

sharwell:
Fixes #1614

### Tested versions of Roslyn:

* 1.0.0
* 1.1.0
* 1.1.1

:memo: Assembly binding redirection was used to run analysis with different versions of Roslyn without recompiling the project.

### Tested projects:

* DartVS/DartVS@6f54d1d2: DanTup.DartVS.sln
* JamesNK/Newtonsoft.Json@48786adc: Src\Newtonsoft.Json.Portable.sln

Projects are good candidates when they have the following characteristics:

* No code generation is performed during the build. A simple clone and NuGet restore is all that's required to load and analyze the solution with 
* The code base contains a wide variety of various style violations.
* The project is open source on GitHub, and is expected to remain in the same location and not have the history rewritten anytime soon. Note that new commits to the project do not affect results because we always check out a specific commit in the history.

### Outputs:

* **{Project}-{Version}.txt:** *All* reported diagnostics for the specified project and Roslyn version
* **{Project}-{Version}-Unique.txt:** The reported diagnostics for the specified project and Roslyn version, with duplicates removed. A diagnostic is considered a duplicate if the same diagnostic has already been reported for the same project. This means the same diagnostic can appear in the output as long as the file on disk was included in multiple projects.

### Usage notes:

* The **-Unique** outputs provide the best comparison between different versions of Roslyn, since this matches the behavior of the reporting infrastructure used by builds.
* Comparing the two outputs for the same project and Roslyn version is helpful for verifying that Roslyn's diagnostic driver is not resulting in the same analysis calls occurring multiple times. This happened with Roslyn 1.0 for each of the tested projects, but so far has not been observed at all for Roslyn 1.1+.

### Limitations:

* Would be nice to fix:
  * No automation is provided to filter out/highlight differences between the various Roslyn versions. Better get your diff tools ready for some large files...
  * No summaries are provided regarding the specific analyzers, files, and number of diagnostics affected by changes across Roslyn versions.
  * The lack of documentation output (SA1652) is likely [negatively] affecting the results for other documentation-related diagnostics which are disabled under these conditions. I'm considering adding SHFB to the set of tested projects provided it meets the other conditions above and does output documentation files.
* Maybe fix in the future?
  * We aren't injecting settings files at this time, so the behavior is only tested in the context of the default settings.
  * Code fix behavior is not tested.
  * We are not testing to ensure the multiple runs with the same version of Roslyn produce the same results (failing this case would indicate race conditions in the analyzer driver). However, I did not notice any differences in this aspect under simple copy/paste checking locally.


sharwell:
@mavasani: You will definitely want to run this script. We're seeing a variety of subtle differences between Roslyn 1.0.0 and 1.1.0, as well as between Roslyn 1.0.0/1.1.0 and 1.1.1. I'm only in a position to make users aware of this, but hopefully you are better positioned to identify problems and get them corrected for future Roslyn releases.

@giggio: You may find this helpful for your project.

sharwell:
Sample data from commit c6c3b0fd:

[SampleData.zip](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/files/66609/SampleData.zip)

vweijsters:
:+1: Looks good to me. Impressive piece of work :smile:

Issue #1973, opened 2015-12-18
Improve SA1129 code fix for CancellationToken


sharwell:
The SA1129 code fix currently replaces `new CancellationToken()` with `default(CancellationToken)`. An improved code fix would use `CancellationToken.None` instead.

nbarbettini:
Grabbing this.

PR #1976, opened 2015-12-21
Add special cases to SA1129 code fix

### LINKED ISSUE #1972: Improve SA1129 code fix for enum types
Currently the SA1129 code fix replaces `new TaskState()` with `default(TaskState)`. An improved code fix for enum types would be the following:

* If a single member `Member` in the enum declaration `T` has the value 0, replace `new T()` with `T.Member`
* Otherwise, replace `new T()` with `default(T)` (i.e. fall back to the current behavior)

nbarbettini:
Fixes #1972 and #1973

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
@@ -144,7 +144,7 @@ namespace StyleCop.Analyzers.ReadabilityRules
                 return false;
             }
 
-            foundMemberName = foundMembers.Single().Name;
+            foundMemberName = foundMembers[0].Name;
             return true;
         }
 
@@ -155,10 +155,12 @@ namespace StyleCop.Analyzers.ReadabilityRules
         /// <param name="memberName">The member name.</param>
         /// <returns>A new member access expression.</returns>
         private static SyntaxNode ConstructMemberAccessSyntax(TypeSyntax typeSyntax, string memberName)
-            => SyntaxFactory.MemberAccessExpression(
+        {
+            return SyntaxFactory.MemberAccessExpression(
                 SyntaxKind.SimpleMemberAccessExpression,
                 typeSyntax,
                 SyntaxFactory.IdentifierName(memberName));
+        }
 
         private class FixAll : DocumentBasedFixAllProvider
         {
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
@@ -97,7 +97,7 @@ namespace StyleCop.Analyzers.ReadabilityRules
             {
                 return false;
             }
-            
+
             return true;
         }
 
@@ -131,11 +131,11 @@ namespace StyleCop.Analyzers.ReadabilityRules
                 }
 
                 var syntaxRoot = await document.GetSyntaxRootAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
-                var semanticModel = await document.GetSemanticModelAsync().ConfigureAwait(false);
+                var semanticModel = await document.GetSemanticModelAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
 
                 var nodes = diagnostics.Select(diagnostic => syntaxRoot.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true));
 
-                return syntaxRoot.ReplaceNodes(nodes, (originalNode, rewrittenNode) => GetReplacementNode(rewrittenNode, semanticModel, CancellationToken.None));
+                return syntaxRoot.ReplaceNodes(nodes, (originalNode, rewrittenNode) => GetReplacementNode(rewrittenNode, semanticModel, fixAllContext.CancellationToken));
             }
         }
     }
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
@@ -450,6 +450,7 @@ public class TestClass
         [InlineData("Member = 0")]
         [InlineData("Member = 0, Another = 1")]
         [InlineData("Another = 1, Member = 0")]
+        [InlineData("Member, Another")]
         public async Task VerifyEnumMemberReplacementBehaviorAsync(string declarationBody)
         {
             var testCode = $@"public class TestClass
@@ -491,6 +492,8 @@ public class TestClass
         [Theory]
         [InlineData("Member = 1")]
         [InlineData("Member = 1, Another = 2")]
+        [InlineData("FooMember = 0, BarMember = 0")]
+        [InlineData("FooMember, BarMember = 0")]
         public async Task VerifyEnumMemberDefaultBehaviorAsync(string declarationBody)
         {
             var testCode = $@"public class TestClass
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
@@ -279,7 +279,7 @@ public class TestClass
         }
 
         /// <summary>
-        /// Verifies that the codefix will preserve trivia surrounding <c>new CancellationToken()</c>
+        /// Verifies that the codefix will preserve trivia surrounding <c>new CancellationToken()</c>.
         /// </summary>
         /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
         [Fact]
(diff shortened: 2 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
```diff
@@ -58,12 +58,32 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            }
+
+            return replacement
                 .WithLeadingTrivia(newExpression.GetLeadingTrivia())
                 .WithTrailingTrivia(newExpression.GetTrailingTrivia());
         }
 
+        private static SyntaxNode GetCancellationTokenNoneSyntax()
+        {
+            return SyntaxFactory.MemberAccessExpression(
```
sharwell: This needs to start as fully-qualified and then be simplified (using `Simplifier`) rather than assuming `CancellationToken` is in scope.

nbarbettini: Ok, that makes sense.

nbarbettini:
@sharwell I used your second suggestion and reused the `NameSyntax` from the constructor. How does this look?

sharwell:
This seems to take care of it. I'll wait for @vweijsters or @pdelvo to second that before merging. :smile:

Noryoko:
Should this support aliased cancellation tokens?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
```diff
@@ -58,12 +58,32 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
         private static SyntaxNode GetReplacementNode(SyntaxNode node)
         {
             var newExpression = (ObjectCreationExpressionSyntax)node;
+            var nameSyntax = newExpression.Type as NameSyntax;
+            var identifierName = NameSyntaxHelpers.ToNormalizedString(nameSyntax);
 
-            return SyntaxFactory.DefaultExpression(newExpression.Type)
+            SyntaxNode replacement = null;
+            if (identifierName.EndsWith(nameof(CancellationToken), System.StringComparison.OrdinalIgnoreCase))
```
sharwell: :exclamation: I overlooked this part originally. This check needs to check the *symbol*, not the syntax.

vweijsters: The `EndsWith` check is also a problem when there is something like `new MyOwnCancellationToken()`. It should be a verify specific check on `CancellationToken`.

nbarbettini:
@sharwell @vweijsters My latest push should address the issues you pointed out: the code fix now compares the symbol, not the syntax, and takes the fully qualified name (including namespace) into account. I've added a new unit test to verify that somebody adding a goofy `struct CancellationToken` won't trick it. :smile:

nbarbettini:
@Noryoko Not sure what you mean - can you give an example?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
```diff
@@ -81,10 +131,11 @@ private class FixAll : DocumentBasedFixAllProvider
                 }
 
                 var syntaxRoot = await document.GetSyntaxRootAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
+                var semanticModel = await document.GetSemanticModelAsync().ConfigureAwait(false);
```
vweijsters: This should use `fixAllContext.CancellationToken`.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
```diff
@@ -81,10 +131,11 @@ private class FixAll : DocumentBasedFixAllProvider
...
 
                 var nodes = diagnostics.Select(diagnostic => syntaxRoot.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true));
 
-                return syntaxRoot.ReplaceNodes(nodes, (originalNode, rewrittenNode) => GetReplacementNode(rewrittenNode));
+                return syntaxRoot.ReplaceNodes(nodes, (originalNode, rewrittenNode) => GetReplacementNode(rewrittenNode, semanticModel, CancellationToken.None));
```
vweijsters: This should use `fixAllContext.CancellationToken`

Noryoko:
@nbarbettini Here's an example.
```
using SystemToken = System.Threading.CancellationToken;

public class TestClass
{
    private SystemToken cancellantionToken = new SystemToken();
}
```

nbarbettini:
@Noryoko I'll leave that up to @sharwell and @vweijsters. I'm not sure if it's worth supporting that edge case.

sharwell:
@nbarbettini Now that you are using a semantic check, that case should actually be supported with no additional work. Maybe just add a unit test to verify.

nbarbettini:
@sharwell Hey, you're right! :smile: Done.

nbarbettini:
Since they were similar, I expanded this PR to include the code fix for #1972 as well. :+1:

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
```diff
@@ -49,21 +50,116 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                .OfType<IFieldSymbol>()
+                .Where(fs => fs.ConstantValue.Equals(0))
+                .ToList();
+
+            if (foundMembers.Count != 1)
+            {
+                return false;
+            }
+
+            foundMemberName = foundMembers.Single().Name;
```
vweijsters: :thought_balloon:  Since `foundMembers` is already a list, `foundMembers[0].Name` will work more efficiently.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
```diff
@@ -49,21 +50,116 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
+        /// Gets a qualified member access expression for the given <paramref name="typeSyntax"/>.
+        /// </summary>
+        /// <param name="typeSyntax">The type syntax from the original constructor.</param>
+        /// <param name="memberName">The member name.</param>
+        /// <returns>A new member access expression.</returns>
+        private static SyntaxNode ConstructMemberAccessSyntax(TypeSyntax typeSyntax, string memberName)
+            => SyntaxFactory.MemberAccessExpression(
+                SyntaxKind.SimpleMemberAccessExpression,
+                typeSyntax,
+                SyntaxFactory.IdentifierName(memberName));
```
vweijsters: :question: I'm not particularly found of this syntax.
@sharwell @pdelvo How do you think above the construction above?

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
```diff
@@ -237,6 +237,292 @@ public T TestMethod2<T>()
...
+            DiagnosticResult[] expected =
+            {
+                this.CSharpDiagnostic().WithLocation(5, 18),
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
+        }
+
```
vweijsters: :question: Could you please also add a test case for an `enum` with multiple `0` value members?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
nbarbettini: @vweijsters The expression-bodied method syntax?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
nbarbettini: Good catch, thanks!

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1129UnitTests.cs
nbarbettini: Done

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
vweijsters: Yes, I think it reduces readability, but that is my personal opinion. That's why I asked @sharwell and @pdelvo for their opinions.

nbarbettini: I was on the fence in this case. Usually I think it improves readability, but with the large block here I wasn't sure.

pdelvo: I personally wouldn't use the arrow syntax if the expression spans multiple lines. But I don't think it is a big deal.

vweijsters:
:+1: Looks good to me now

PR #1979, opened 2015-12-21
Work around dotnet/roslyn#7446

### LINKED ISSUE #1939: StyleCop Violation SA12** not displayed in Visual Studio CodeFiles
I have found some strange problem with some violations not displayed inside of visual studio code files.

The StyleCop Violation SA1200 and SA1210 are listed as warnings after building the project (in visual studio or with msbuild), but not displayed in the CodeFile in Visual Studio.
Also when i open the CodeFile, where the issue is listed, the issue dissapears from warning list, until i build the project again.

I am using StyleCopAnalyzers RC2, Visual Studio 2015 Update 1 (and Resharper 10.0.1).
I have also tryed to suspend Resharper to ensure it doesn't conflict with Stylecop, but thes didn't help.

sharwell:
Fixes #1939 and related problems with in-IDE diagnostic reporting.


sharwell:
@mavasani I'm assigning this to you hoping you can do a code review. :smile: 

I'm interested in the following two things:

1. Does this properly implement the workaround you described?
2. We will create an RC 3 release to get some testing before releasing 1.0.0. Normally this would reset our RC testing timeline, thus delaying 1.0.0 stable from January 6 to January 21. Are you sufficiently confident in this change that we can keep the schedule for January 6, provided this is the only change going into RC 3?

mavasani:
@sharwell The change looks good to me. Did you do a quick sanity to verify IDE behavior after this change?

sharwell:
I haven't had the time yet, but I definitely will before merging it.

mavasani:
Okay, you probably just want to ensure that intellisense reports the same set of diagnostics as build for the rules where you were seeing differences reported by StyleCopTester.

sharwell:
Looks like it fixed it!

pdelvo:
:+1:

sharwell:
@mavasani I'm seeing this analyzer's compilation start method get called just over 1200 times when loading StyleCopAnalyzers.sln, over the course of about 5 minutes. If I try to run Fix All in Project before this completes, the operation takes a long time because it waits for them to finish. If I run Fix All in Project after the 1200 complete, it's very fast. However, making a change in a document starts a new sequence of these events, so making a change in a document means you can't run Fix All in Project for the next ~5 minutes.

sharwell:
I used SA1028 (which works with and without this workaround) to compare the performance difference. I'm seeing about a 15:1 performance hit with this change on an operation that already took 20 seconds.

mavasani:
@sharwell That's an unfortunate regression that's crept into Update1. IDE background analysis creates a new CompilationWithAnalyzers for each document being analyzed as holding it in memory was causing VM regressions. Unfortunately, FixAll code path for computing diagnostics wasn't tuned to re-use the same CompilationWithAnalyzers instance for the entire project. It does re-use the cached diagnostics from background analysis - hence the behavior of it being fast when all project diagnostics have been computed in the background, and slow otherwise.

As a workaround for Update1, can you try to change your fix all provider to not use FixAllContext.GetXXXDiagnosticsAsync methods and instead do the following:

1. Create a new `CompilationWithAnalyzers` instance with the specific analyzer (you'll need to create a static map from analyzer IDs to analyzer type)
2. Invoke `CompilationWithAnalyzers.GetAnalyzerDiagnostics()` to get all project diagnostics for that analyzer
3. Document scope: Either re-use the same GetAnalyzerDiagnostics method and filter diagnostics to document OR
  1. For analyzers with only Syntax tree action: invoke `CompilationWithAnalyzers.GetAnalyzerSyntaxDiagnostics(SyntaxTree)`
  2. For rest of analyzers: invoke Invoke `CompilationWithAnalyzers.Compilation.GetDeclarationDiagnostics()` and then `CompilationWithAnalyzers.GetAnalyzerSemanticDiagnostics(SemanticModel)`

sharwell:
Alright I have testing results:

### Setup:

1. Uninstall StyleCop.Analyzers package from projects in StyleCopAnalyzers.sln
2. Replace with regex `\}(?=\r?$)` &rarr; (`}` followed by two space characters) in \*.cs
3. Run StyleCop.Analyzers.Vsix in Release configuration without debugging
4. Load StyleCopAnalyzers.sln and immediately fix all SA1028 violations in solution

### Results

| Code base | Behavior | Elapsed time (minutes) |
| --- | --- | --- |
| Current **stabilization** branch | IDE crash | 6:55 |
| #1983 applied to previous | Success | 0:30 |
| #1983 and #1979 applied | Success | 6:36 |

I also have a second performance tweak for more aggressive caching of generated document analysis which I'm testing now.

sharwell:
@mavasani I added results for `GetAnalyzerDiagnosticsAsync` to the table.

sharwell:
@mavasani The use of `GetAnalyzerDiagnosticsAsync` in the code fix resolves the performance problems with Fix All, whether or not #1979 is applied (which is good news). However, #1979 appears to dramatically reduce the speed at which the Error List is updated. Do you have any ideas on how to resolve that?

mavasani:
@sharwell Without #1979, you will end up with intermittently missing analyzer diagnostics in Update1. Do you think you can try workaround 2 mentioned here: https://github.com/dotnet/roslyn/issues/7446#issuecomment-164510894? It should be easily implemented for syntax node analyzers, but would be non-trivial for symbol analyzers (if you have any) as it needs computing all symbols declared in file (you can use http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/CSharpDeclarationComputer.cs,6dc18cc1267a8e01,references as a reference).

sharwell:
I think that workaround will be more work than it's worth. Our analyzer registrations are already non-trivial, and we already have to change them all after Update 2 to leverage the new per-compilation state features. Perhaps we just provide release notes that say stick with RTM or wait for Update 2.

sharwell:
I made several usability improvements to this improvements:

* Rename `Roslyn7446WorkaroundAnalyzer` to `SA0000Roslyn7446Workaround`
* Allow users to configure the behavior of **SA0000**
* Improve the title and description of **SA0000**
* Disable the rule by default for improved performance

:memo: The help link would give a 404 error, but since no diagnostics are ever reported by this analyzer the user will never be able to click the link.

StyleCop.Analyzers/StyleCop.Analyzers/SpecialRules/SA0000Roslyn7446Workaround.cs
```diff
@@ -0,0 +1,59 @@
...
+    internal sealed class SA0000Roslyn7446Workaround : DiagnosticAnalyzer
+    {
+        /// <summary>
+        /// The ID for diagnostics produced by the <see cref="SA0000Roslyn7446Workaround"/> analyzer.
+        /// </summary>
+        public const string DiagnosticId = "SA0000";
+        private static readonly LocalizableString Title = new LocalizableResourceString(nameof(SpecialResources.SA0000Title), SpecialResources.ResourceManager, typeof(SpecialResources));
+        private static readonly LocalizableString MessageFormat = new LocalizableResourceString(nameof(SpecialResources.SA0000MessageFormat), SpecialResources.ResourceManager, typeof(SpecialResources));
+        private static readonly LocalizableString Description = new LocalizableResourceString(nameof(SpecialResources.SA0000Description), SpecialResources.ResourceManager, typeof(SpecialResources));
+        private static readonly string HelpLink = "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA0000Roslyn7446Workaround.md";
```
pdelvo: Wouldn't it better to link to the issue or to this PR?

pdelvo:
:+1:

StyleCop.Analyzers/StyleCop.Analyzers/SpecialRules/SA0000Roslyn7446Workaround.cs
sharwell: We have a unit test that checks the help link for each diagnostic.

sharwell: I added a proper documentation page for this special diagnostic.

PR #1981, opened 2015-12-21
SA1115 will now allow comments between parameters

### LINKED ISSUE #1921: SA1115 should allow comments between parameters.
The following code should not produce SA1115, but it currently does:

```csharp
    public void MyTest(int v1, int v2, int v3)
    {
    }

    public void TestMethod()
    {
        this.MyTest(
            1,
            /* comment */
            2,
            /* comment */
            3);
    }
```

The same construction should probably also be allowed for `////` comments. `//` comments will not work because they require an empty line before them and that will (and shall) trigger SA1115.

vweijsters:
Fixes #1921


Noryoko:
:+1:

PR #1983, opened 2015-12-22
Improve DocumentBasedFixAllProvider performance for multiple documents

### LINKED ISSUE #1974: Reduce memory usage of the batch fixer
Currently the Fix All In Project|Solution operations are unusable for even moderate numbers of violations in large projects, except for cases where we implemented a custom fix all provider. We should look into ways to reduce the memory consumption of the batch fixer in an effort to avoid this pain point.

sharwell:
Previously, custom fix all providers called `FixAllContext.GetDocumentDiagnosticsAsync` for every document in the fix all scope. This change first replaces those calls with a single call to `FixAllContextHelpers.GetDocumentDiagnosticsToFixAsync`, which was extracted from `CustomBatchFixAllProvider`. It goes on to instead operate on a `CompilationWithAnalyzers` following the commentary in #1979. The result is improved performance when fixing violations in large projects, especially when only a few documents in the project contained violations.

Fixes #1974.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/FixAllContextHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/FixAllContextHelper.cs
@@ -126,6 +126,44 @@ namespace StyleCop.Analyzers.Helpers
             return ImmutableDictionary<Project, ImmutableArray<Diagnostic>>.Empty;
         }
 
+        public static async Task<ImmutableArray<Diagnostic>> GetAllDiagnosticsAsync(Compilation compilation, CompilationWithAnalyzers compilationWithAnalyzers, IEnumerable<Document> documents, bool includeCompilerDiagnostics, CancellationToken cancellationToken)
+        {
+            if (GetAnalyzerSyntaxDiagnosticsAsync == null || GetAnalyzerSemanticDiagnosticsAsync == null)
+            {
+                // In everything except Roslyn 1.1, we use GetAllDiagnosticsAsync and return the result.
+                var allDiagnostics = await compilationWithAnalyzers.GetAllDiagnosticsAsync().ConfigureAwait(false);
+                return allDiagnostics;
+            }
+
+            compilationWithAnalyzers.Compilation.GetDeclarationDiagnostics(cancellationToken);
+
+            // Note that the following loop to obtain syntax and semantic diagnostics for each document cannot operate
+            // on parallel due to our use of a single CompilationWithAnalyzers instance. Also note that the following
+            // code is not sufficient for cases where analyzers register compilation end actions. However, this project
+            // does not currently contain any such analyzers.
+            var diagnostics = ImmutableArray<Diagnostic>.Empty;
+            foreach (var document in documents)
+            {
+                var syntaxTree = await document.GetSyntaxTreeAsync(cancellationToken).ConfigureAwait(false);
+                var syntaxDiagnostics = await GetAnalyzerSyntaxDiagnosticsAsync(compilationWithAnalyzers, syntaxTree, cancellationToken).ConfigureAwait(false);
+                diagnostics = diagnostics.AddRange(syntaxDiagnostics);
+
+                var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+                var semanticDiagnostics = await GetAnalyzerSemanticDiagnosticsAsync(compilationWithAnalyzers, semanticModel, default(TextSpan?), cancellationToken).ConfigureAwait(false);
+                diagnostics = diagnostics.AddRange(semanticDiagnostics);
+            }
+
+            if (includeCompilerDiagnostics)
+            {
+                // This is the special handling for cases where code fixes operate on warnings produced by the C#
+                // compiler, as opposed to being created by specific analyzers.
+                var compilerDiagnostics = compilation.GetDiagnostics(cancellationToken);
+                diagnostics = diagnostics.AddRange(compilerDiagnostics);
+            }
+
+            return diagnostics;
+        }
+
         private static ImmutableDictionary<string, ImmutableArray<Type>> GetAllAnalyzers()
         {
             Assembly assembly = typeof(NoCodeFixAttribute).GetTypeInfo().Assembly;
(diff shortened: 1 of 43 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/FixAllContextHelper.cs
```diff
@@ -0,0 +1,228 @@
...
+        private static async Task<ImmutableArray<Diagnostic>> GetAllDiagnosticsAsync(FixAllContext fixAllContext, Project project, ImmutableArray<DiagnosticAnalyzer> analyzers)
+        {
+            bool includeCompilerDiagnostics = fixAllContext.DiagnosticIds.Any(x => x.StartsWith("CS", StringComparison.Ordinal));
+            var compilation = await project.GetCompilationAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
+
+            var diagnostics = ImmutableArray<Diagnostic>.Empty;
+            if (analyzers.Any())
+            {
+                var compilationWithAnalyzers = compilation.WithAnalyzers(analyzers, project.AnalyzerOptions, fixAllContext.CancellationToken);
+                diagnostics = await compilationWithAnalyzers.GetAnalyzerDiagnosticsAsync().ConfigureAwait(false);
```
sharwell: @mavasani I'm finding that when I use `CompilationWithAnalyzers` here along with #1979, even though all diagnostics are reported in the errors pane in Visual Studio not all diagnostics end up corrected by the Fix All in Project and Fix All in Solution operations. Is there some way I can update this method so we have the performance of `CompilationWithAnalyzers.GetAnalyzerDiagnosticsAsync`, but also get all diagnostics?

mavasani: GetAnalyzerDiagnostics seems to be affected by the same bug you pointed out in your churn testing. You may try to replicate what IDE diagnostic service does sequentially - for each document invoke GetAnalyzerSyntaxDiagnosticsAsync and GetAnalyzerSemanticDiagnosticsAsync. If you also have any compilation end analyzers, then also invoke GetAnalyzerCompilationDiagnosticsAsync at the end. This won't be as fast as GetAnalyzerDiagnosticsAsync() but should be faster then current performance of FixAllContext.GetAllDiagnosticsAsync.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/FixAllContextHelper.cs
```diff
@@ -0,0 +1,267 @@
...
+                        diagnostics = diagnostics.AddRange(syntaxDiagnostics);
+
+                        var semanticModel = await document.GetSemanticModelAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
+                        var semanticDiagnostics = await GetAnalyzerSemanticDiagnosticsAsync(compilationWithAnalyzers, semanticModel, default(TextSpan?), fixAllContext.CancellationToken).ConfigureAwait(false);
+                        diagnostics = diagnostics.AddRange(semanticDiagnostics);
+                    }
+                }
+                else
+                {
+                    diagnostics = await compilationWithAnalyzers.GetAnalyzerDiagnosticsAsync().ConfigureAwait(false);
```
sharwell: @mavasani This code is in effect for Roslyn 1.0 and Roslyn 1.2+. Originally this method was written to work around performance problems in Roslyn 1.1. For the others, is there any reason to continue using `CompilationWithAnalyzers`, or should I go back to using `FixAllContext.GetAllDiagnosticsAsync`?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/FixAllContextHelper.cs
```diff
@@ -0,0 +1,267 @@
...
+                var compilationWithAnalyzers = compilation.WithAnalyzers(analyzers, project.AnalyzerOptions, fixAllContext.CancellationToken);
+                if (GetAnalyzerSyntaxDiagnosticsAsync != null && GetAnalyzerSemanticDiagnosticsAsync != null)
+                {
+                    // This whole block is workaround code for issues with Roslyn 1.1...
+                    compilationWithAnalyzers.Compilation.GetDeclarationDiagnostics(fixAllContext.CancellationToken);
+
+                    var tasks = new List<Task<ImmutableArray<Diagnostic>>>();
+                    foreach (var document in project.Documents)
+                    {
+                        var syntaxTree = await document.GetSyntaxTreeAsync(fixAllContext.CancellationToken).ConfigureAwait(false);
```
sharwell: :memo: The processing of documents cannot run in parallel, because a single `CompilationWithAnalyzers` instance is used for each of the documents. We are using a single instance to ensure that all diagnostics are identified following a *single* call to `GetDeclarationDiagnostics`, as opposed to calling it once per document.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/FixAllContextHelper.cs
mavasani: I would go back to `FixAllContext.GetAllDiagnosticsAsync` - it will give better performance compared to `CompilationWithAnalyzers` when background analysis has already computed and cached the diagnostics.
Can you also let us know how to run your fixall perf tests? We would surely want to validate that the issues in `FixAllContext.GetAllDiagnosticsAsync` are fully addressed in 1.2

sharwell: The performance tests can be run using the 1.0.0 RC 2 release using the steps described in the Setup section of [#1979 (comment)](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1979#issuecomment-166656002).

sharwell:
With the latest change, the churn tests indicate that aside from true duplicates in 1.0.0, all three tested versions of Roslyn produce **identical outputs**.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Helpers/FixAllContextHelper.cs
```diff
@@ -0,0 +1,307 @@
...
+                // In everything except Roslyn 1.1, we use GetAllDiagnosticsAsync and return the result.
+                var allDiagnostics = await compilationWithAnalyzers.GetAllDiagnosticsAsync().ConfigureAwait(false);
+                return allDiagnostics;
+            }
+
+            compilationWithAnalyzers.Compilation.GetDeclarationDiagnostics(cancellationToken);
+
+            // Note that the following loop to obtain syntax and semantic diagnostics for each document cannot operate
+            // on parallel due to our use of a single CompilationWithAnalyzers instance.
+            var diagnostics = ImmutableArray<Diagnostic>.Empty;
```
vweijsters: :thought_balloon: using an `ImmutableArray<Diagnostic>.Builder` or a `List<Diagnostic>` would improve the performance when adding diagnostics. There is no need to have the diagnostics in the `ImmutableArray<Diagnostic>` type yet, as they are never passed to another method.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1412FixAllProvider.cs
```diff
@@ -3,77 +3,111 @@
...
+                    title,
+                    cancellationToken => this.GetDocumentFixesAsync(fixAllContext.WithCancellationToken(cancellationToken)),
+                    nameof(SA1412FixAllProvider));
                 break;
 
             case FixAllScope.Project:
-                newSolution = await GetProjectFixesAsync(fixAllContext, fixAllContext.Project).ConfigureAwait(false);
+                fixAction = CodeAction.Create(
+                    title,
+                    cancellationToken => this.GetProjectFixesAsync(fixAllContext.WithCancellationToken(cancellationToken), fixAllContext.Project),
```
vweijsters: :question: Why does this need to pass the project? Its already in the `fixAllContext` as far as I can determine.

sharwell: In this case, it doesn't. I also realized the methods can be `static`.

vweijsters:
:+1: Looks good to me, with two small remarks.

pdelvo:
:+1:

Issue #1879, opened 2015-12-01
SA1133 batch code fix does half the work


vbfox:
Given the following code (Using the `JetBrains.Annotations` NuGet package):

```csharp
using JetBrains.Annotations;

namespace Stylecop_rc1_bug_repro
{
    class Foo
    {
        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo1{ get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo2{ get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo3 { get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo4{ get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo5{ get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo6{ get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo7{ get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo8{ get; set; }

    }
}
```

Applying SA1133 code fix on document scope give :

```csharp
using JetBrains.Annotations;

namespace Stylecop_rc1_bug_repro
{
    class Foo
    {
        [CanBeNull]
        [UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo1 { get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo2 { get; set; }

        [CanBeNull]
        [UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo3 { get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo4 { get; set; }

        [CanBeNull]
        [UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo5 { get; set; }

        [CanBeNull, UsedImplicitly(ImplicitUseKindFlags.Assign)]
        public string Foo6 { get; set; }

        [CanBeNull]
        [UsedImplicitly(ImplicitUseKindFlags.Assign)]
[...]

PR #1989, opened 2015-12-23
Increase strictness of Fix All iteration count tests

sharwell:
* Check iterations as a fixed number instead of a bound
* Assume "Fix All" operations take one iteration unless specified

This change helps ensure situations like #1879 do not go unnoticed.


vweijsters:
:+1:

PR #2002, opened 2015-12-26
Remove duplicate filtering

### LINKED ISSUE #2000: SA1211 fails to preserve necessary using directive
When the code fix for the SA1211 reported in the following code is executed, the `Func1` using declaration is eliminated from the result, causing the code to no longer compile.

```csharp
namespace NamespaceName
{
    using Func2 = System.Func<int>;
    using Func1 = System.Func<int>;

    internal delegate void DelegateName(Func1 arg1, Func2 arg2);
}
```

sharwell:
* [x] Add regression test for #2000 
* [x] Fixes #2000


vweijsters:
:+1: Looks good to me

pdelvo:
:+1:

PR #2003, opened 2015-12-27
Adding a comment about the use of DispatcherSynchronizationContext

hmemcpy:
I'm reading this code, and it took me a few moments to understand what this code did. I've added a comment, pointing to the original post by Stephen Toub, explaining how to leverage the WPF dispatcher for this purpose.

### DIFF
--- a/StyleCop.Analyzers/StyleCopTester/Program.cs
+++ b/StyleCop.Analyzers/StyleCopTester/Program.cs
@@ -39,6 +39,10 @@ namespace StyleCopTester
                     cts.Cancel();
                 };
 
+            // The following allows continuations to be processed by a single thread.
+            // Since Console apps do not have a SyncronizationContext, we're leveraging the built-in support
+            // in WPF to pump the messages via the Dispatcher.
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -39,6 +39,10 @@ private static void Main(string[] args)
                     cts.Cancel();
                 };
 
+            // The following allows continuations to be processed by a single thread.
```
sharwell: :thought_balloon: We use `ConfigureAwait(false)` pretty much everywhere, so the comment about using a single thread is misleading at best, but probably incorrect.

sharwell: :bulb: I think a good approach would be:

1. Remove the first line of this comment.
2. Add a second informational link to https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1362.

sharwell: :arrow_right: At the request of @hmemcpy, I made these corrections during the merge.

PR #2009, opened 2015-12-28
opencover-report.bat to stabilization

ddur:
Runs OpenCover/xUnit/StyleCop.Analyzers.Test
Runs ReportGenerator

When done open .\OpenCover.Reports\index.htm


pdelvo:
I tried this locally. Very handy to be able to create these reports locally :+1:

PR #2013, opened 2015-12-28
Fix handling of Incremental switch in build.ps1

sharwell:
The build script was failing to copy the build output to the intermediate output directory.


vweijsters:
:+1:

PR #2015, opened 2015-12-29
Implement SA1315

### LINKED ISSUE #1949: New rule proposal: ParametersShouldMatchInheritedNames
**Category:** Naming
**Title:** Parameters should match inherited names
**Description:** When a method overrides a method from a base class, or implements an interface method, the parameter names of the overriding method should match the names in the base definition.

This rule is an *enforcement* of the special case added to SA1313 in #1372.

pdelvo:
Fixes #1949.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/RenameToAnyCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/RenameToAnyCodeFixProvider.cs
@@ -1,0 +1,82 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.NamingRules
+{
+    using System.Collections.Immutable;
+    using System.Composition;
+    using System.Threading.Tasks;
+    using Helpers;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.CSharp;
+
+    /// <summary>
+    /// Implements a code fix for diagnostics which are fixed by renaming a symbol to a set of given new names.
+    /// </summary>
+    /// <remarks>
+    /// <para>To fix a violation of this rule, change the name of the symbol so that it is one of the set of new names.</para>
+    /// </remarks>
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(RenameToAnyCodeFixProvider))]
+    [Shared]
+    internal class RenameToAnyCodeFixProvider : CodeFixProvider
+    {
+        private static char[] comma = { ',' };
+
+        /// <inheritdoc/>
+        public override ImmutableArray<string> FixableDiagnosticIds { get; } =
+            ImmutableArray.Create(
+                SA1315ParametersShouldMatchInheritedNames.DiagnosticId);
+
+        /// <inheritdoc/>
+        public override FixAllProvider GetFixAllProvider()
+        {
+            return CustomFixAllProviders.BatchFixer;
+        }
+
+        /// <inheritdoc/>
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var document = context.Document;
+            var root = await document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            foreach (var diagnostic in context.Diagnostics)
+            {
+                var token = root.FindToken(diagnostic.Location.SourceSpan.Start);
+                if (string.IsNullOrEmpty(token.ValueText))
+                {
+                    continue;
+                }
+
+                var originalName = token.ValueText;
+
+                var memberSyntax = RenameHelper.GetParentDeclaration(token);
+
+                SemanticModel semanticModel = await document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+
+                var declaredSymbol = semanticModel.GetDeclaredSymbol(memberSyntax);
+                if (declaredSymbol == null)
+                {
+                    continue;
+                }
+
+                string[] newNames = diagnostic.Properties[SA1315ParametersShouldMatchInheritedNames.PropertyName].Split(comma);
+                foreach (var newName in newNames)
+                {
+                    if (!await RenameHelper.IsValidNewMemberNameAsync(semanticModel, declaredSymbol, newName, context.CancellationToken).ConfigureAwait(false))
+                    {
+                        continue;
+                    }
+
+                    context.RegisterCodeFix(
+                        CodeAction.Create(
+                            string.Format(NamingResources.RenameToCodeFix, newName),
+                            cancellationToken => RenameHelper.RenameSymbolAsync(document, root, token, newName, cancellationToken),
+                            nameof(RenameToAnyCodeFixProvider) + newName),
+                        diagnostic);
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


sharwell:
I think this needs to be SA1315 (see #1925).

documentation/SA1314.md
```diff
@@ -0,0 +1,40 @@
...
+
+:memo: This rule is new for StyleCop Analyzers, and was not present in StyleCop Classic.
+
+## Cause
+
+A violation of this rule occurs when the name of a parameter does not match its inherited name.
+
+## Rule description
+
+A violation of this rule occurs when the name of a parameter does not begin with a lower-case letter.
```
vweijsters: This seems a copy-paste error.

vweijsters:
:thought_balloon: It would be nice to have a few test cases that would cover multiple levels of inheritance. Like:
```csharp
public class TopLevelBaseClass
{
    public void Method(int baseArg);
}

public class IntermediateBaseClass : TopLevelBaseClass
{
}

public class TestClass : IntermediateBaseClass
{
  public override Method(int arg);
}
```

vweijsters:
:question: How would the scenario above work when the `IntermediateBaseClass` redefines the parameter names with an override (like: `public override Method(int renamedArg)`)?

vweijsters:
I think a test is needed when a sub class redefines a method declaration with the `new` keyword (which should not trigger a diagnostic)

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1314ParametersShouldMatchInheritedNames.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        private static void HandleMethod(SymbolAnalysisContext context)
+        {
+            var symbol = context.Symbol as IMethodSymbol;
+
+            if (!symbol.IsOverride && !NamedTypeHelpers.IsImplementingAnInterfaceMember(symbol))
+            {
+                return;
+            }
```
vweijsters: :question: Is it possible that context.Symbol is null because the method declaration is incomplete? -> If so, the statement above needs a null check as well.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1314ParametersShouldMatchInheritedNames.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            var pool = SharedPools.Default<List<IMethodSymbol>>();
+
+            using (var pooledObject = pool.GetPooledObject())
+            {
+                var originalDefinitions = pooledObject.Object;
+                NamedTypeHelpers.GetOriginalDefinitions(originalDefinitions, symbol);
```
vweijsters: :bug: There is a contamination issue here with the usage of the shared pools. The list the retrieved from the pool should be cleared, as this is not done by the `ObjectPool<T>` implementation.

vweijsters: :question: Is there really much benefit in using a shared pool here?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/RenameToAnyCodeFixProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                SemanticModel semanticModel = await document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+
+                var declaredSymbol = semanticModel.GetDeclaredSymbol(memberSyntax);
+                if (declaredSymbol == null)
+                {
+                    continue;
+                }
+
+                string[] newNames = diagnostic.Properties[nameof(newNames)].Split(comma);
```
vweijsters: The `nameof(newNames)` construction here is risky. If an automatic rename is done, the name of the property changes here, but not in the analyzer. It would preferable to declare an internal constant in the analyzer and use that in both places.

vweijsters:
:question: The code fix currently can offer multiple rename options based on the original definitions found. Is this a good way address this? I would expect that the root declaration will determine the name and that there will be no alternatives.

sharwell:
> :question: The code fix currently can offer multiple rename options based on the original definitions found. Is this a good way address this? I would expect that the root declaration will determine the name and that there will be no alternatives.

The following code can produce this:

```csharp
interface IInterface
{
  void Method(string p1);
}

class BaseClass
{
  public abstract void Method(string p2);
}

class Derived : BaseClass, IInterface
{
  public override void Method(string p)
  {
  }
}
```

The parameter `p` needs to be `p1` to match the interface, but `p2` to match the base class. If any code fix is offered in this case, it should be renaming it to `p2`, since the correct way to fix the method with respect to the interface is add a new method:

```csharp
void IInterface.Method(string p1)
{
  this.Method(p1);
}
```

In the case of multiple overloads in a chain of base classes, the analyzer and code fix should only consider the method which would be called using the `base.` syntax. That method could have a `SuppressWarnings` attribute applied to it for this diagnostic, and we don't want all methods which override it to have to repeat that diagnostic.

pdelvo:
What if there are multiple interfaces with parameter names that do not match?

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1314ParametersShouldMatchInheritedNames.cs
pdelvo: I don't think so. It should work like it does with syntax node actions. I register for a method symbol and I should get one.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1314ParametersShouldMatchInheritedNames.cs
pdelvo: lists are in general larger objects because of the underlying array. This is why I'm using a pool in the analyzer code because it is run on every keystroke possibly often.

lgolding:
@pdelvo @sharwell Would you consider also submitting this code to the dotnet/roslyn-analyzers repo as the implementation of rule CA1725, ParameterNamesShouldMatchBaseDeclaration? This is issue dotnet/roslyn-analyzers#457.

@srivatsn FYI

vweijsters:
@lgolding @srivatsn If this is still wanted, I'll be happy to start working on it this week. I checked with @pdelvo and he is OK with it.

lgolding:
Hi Vincent,

We'd be delighted if you would contribute your work on this analyzer SA1315 to the roslyn-analyzers repo as rule [CA1725](https://github.com/dotnet/roslyn-analyzers/issues/457).

The roslyn-analyzers repo includes skeleton code for all the analyzers we haven't yet implemented. The skeleton code for the CA1725 analyzer is [here](https://github.com/dotnet/roslyn-analyzers/blob/e2fe09593060a6b44f0d2a851b6f6b40cbcd2fa7/src/Microsoft.ApiDesignGuidelines.Analyzers/Core/ParameterNamesShouldMatchBaseDeclaration.cs), and the skeleton code for the fixer is [here](https://github.com/dotnet/roslyn-analyzers/blob/e2fe09593060a6b44f0d2a851b6f6b40cbcd2fa7/src/Microsoft.ApiDesignGuidelines.Analyzers/Core/ParameterNamesShouldMatchBaseDeclaration.Fixer.cs).

If you search the roslyn-analyzers repo for "CA1725" you'll see that for both the analyzer and the fixer, there is a base class, and then there are derived classes for C# and VB versions. In your case, you have written a symbol analyzer, which works for both languages, so you would simply remove the C# and VB analyzers, declare the "core" analyzer to handle both C# and VB, and paste your implementation in there. I see that your fixer works at the syntax level, so you would paste your implementation into the C# version. You are welcome to contribute a VB fixer, but you certainly don't have to. At the moment, we are concentrating on analyzers and deferring work on fixers until we have a sufficient set of analyzers completed.

Feel free to email me with any questions (email in my profile).

Best regards,
Larry

pdelvo:
Do we still want this analyzer in stylecop analyzers?

vweijsters:
Good question, I have no real preference.
But if we do want to keep it, we need to line this one up with the one in roslyn-analyzers. I made several small modifications there.

pdelvo:
I close this for now. @vweijsters maybe you can create a new PR with the changes you made

PR #2017, opened 2015-12-30
Add /apply option to write /fixall changes back to disk

sharwell:
This update to StyleCopTester allows it to be used to automate the correction of specific style violations in a solution.

Currently only one type of diagnostic can be fixed at a time, but it would be easy to create a script that invokes the tool multiple times to cover a set of diagnostics.

### DIFF
--- a/StyleCop.Analyzers/StyleCopTester/Program.cs
+++ b/StyleCop.Analyzers/StyleCopTester/Program.cs
@@ -102,7 +102,9 @@ namespace StyleCopTester
 
                 stopwatch.Restart();
 
-                var diagnostics = await GetAnalyzerDiagnosticsAsync(solution, solutionPath, analyzers, cancellationToken).ConfigureAwait(true);
+                bool force = args.Contains("/force");
+
+                var diagnostics = await GetAnalyzerDiagnosticsAsync(solution, solutionPath, analyzers, force, cancellationToken).ConfigureAwait(true);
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCopTester/Program.cs
```diff
@@ -195,7 +212,24 @@ private static void WriteDiagnosticResults(ImmutableArray<Tuple<ProjectId, Diagn
                 {
                     stopwatch.Restart();
                     Console.WriteLine($"Calculating fix for {fix.CodeFixEquivalenceKey} using {fix.FixAllProvider} for {fix.NumberOfDiagnostics} instances.");
-                    await fix.GetOperationsAsync(cancellationToken).ConfigureAwait(true);
+                    var operations = await fix.GetOperationsAsync(cancellationToken).ConfigureAwait(true);
+                    if (applyChanges)
+                    {
+                        var applyOperations = operations.OfType<ApplyChangesOperation>().ToList();
+                        if (applyOperations.Count > 1)
```
pdelvo: @sharwell Do you know a case where this can happen?

pdelvo:
:+1:

StyleCop.Analyzers/StyleCopTester/Program.cs
sharwell: I do not. Presumably if this message ever appears someone will report it.

oatkins:
Excellent! This will be really helpful. :+1:

PR #2018, opened 2015-12-31
Fix exclusion tests

sharwell:
Previously most of our generated code exclusion tests were not running in most cases since we were passing an array containing a null element to the method which tests it.


sharwell:
According to codecov, the difference here is:

Before this change: 23 tests included automatic generated code testing.
After this change: 2743 tests include automatic generated code testing.

vweijsters:
:+1:

PR #2019, opened 2016-01-01
Set newlineAtEndOfFile to require

sharwell:
* Enable SA0000 as a hidden diagnostic. This ensures all diagnostics are reported for files which are open in the editor with Visual Studio 2015 Update 1 without the performance penalty this analyzer would incur on files which are not open for editing.
* Require a newline at the end of files, and fix all violations.


vweijsters:
:+1:

PR #2024, opened 2016-01-05
Implement SA1137 (ElementsShouldHaveTheSameIndentation)

### LINKED ISSUE #2023: Implement SA1137: Elements should have the same indentation
**Category:** Readability
**Title:** Elements should have the same indentation
**Description:** When two elements _A_ and _B_ have the same syntactical parent and each start on their own line, the indentation of each element should be the same.

The following collections of syntax nodes are considered:
- Members of a `CompilationUnitSyntax`
- Members of a `NamespaceDeclarationSyntax`
- Statements of a `BlockSyntax`
- Expressions of an initializer
- Declarators
- Parameters and arguments
- Attribute lists
- Attributes within an attribute list

The following syntax nodes need to be treated specially for this:
- `CaseSwitchLabel`/`DefaultSwitchLabel` (labels in the same switch statement are compared to each other, separately from the statements in the switch statement)
- `LabeledStatement` (labels in the same block are compared to each other, and the child statement of the labeled statement is compared to other non-labeled statements in the block)

sharwell:
Fixes #2023

I'm currently working on tests:
- [ ] Compilation unit
- [x] Namespace declaration
- [x] Base type declaration
- Type declaration
  - [x] Constraint clauses
  - [x] Members
- [x] Enum declaration
- ~~Enum member declaration~~ (enum member attributes now handled at the enum declaration level)
- ~~Base field declaration~~ (field attributes now handled at the type declaration level)
- ~~Base method declaration~~ (method attributes now handled at the type declaration level)
- [x] Method declaration
- ~~Base property declaration~~ (property attributes now handled at the type declaration level)
- [x] Accessor list
- ~~Accessor declaration~~ (accessor attributes now handled at the accessor list level)
- [x] Variable declaration
- [x] Type parameter list
- ~~Type parameter~~ (type parameter attributes now handled at the type parameter list level)
- [x] Base parameter list
- ~~Parameter~~ (parameter attributes now handled at the parameter list level)
- [x] Base argument list
- [x] Attribute list
- [x] Attribute argument list
- [x] Block
- [x] Switch statement
- [x] Initializer expression
- [x] Anonymous object creation expression

I also still need to implement the following cases:
- [x] Anonymous object creation expression
  
  ``` csharp
  new
  {
      X = 0,
      Y = 0,
      Z = 0,
  }
  ```


sharwell:
Blocked on approval of #2023.

sharwell:
At this point I've hit 100% coverage over the main parts of the analyzer. If someone wants to see additional tests, please let me know what you're thinking of. :smile:

Noryoko:
Could you add some XML comments? It will look strange if they start aligned with the element and then are no longer so after the move. Having a rule that checks for the alignment of comments with the element they are for would be a good idea, I think. I'll create a proposal for that idea.

sharwell:
> Could you add some XML comments?

I'll add a couple tests to demonstrate the current behavior.

> It will look strange if they start aligned with the element and then are no longer so after the move. Having a rule that checks for the alignment of comments with the element they are for would be a good idea, I think. I'll create a proposal for that idea.

It will look strange. There's no need to create a separate rule for this. We can treat it the same way we're currently treating attribute lists. It would make sense to implement it separately though.

Noryoko:
Sounds good to me.

sharwell:
This is blocked on #2038 and #2039.

sharwell:
No longer blocked, and updated for review! :smile:

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
```diff
@@ -0,0 +1,432 @@
...
+            }
+
+            // Try to reorder the list so the first item is not an attribute list. This element will establish the
+            // expected indentation for the entire collection.
+            int desiredFirst = elements.FindIndex(x => !x.IsKind(SyntaxKind.AttributeList));
+            if (desiredFirst > 0)
+            {
+                T newFirstElement = elements[desiredFirst];
+                elements = elements.RemoveAt(desiredFirst).Insert(0, newFirstElement);
+            }
```
vweijsters: :question: Would this work correctly if the code would violate SA1134 and has an attribute on the same line as another element?
For example:
```csharp
  [return: MarshallAs(LPStr)] string MyMethod()
    {
    // implementation
  }
```

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
```diff
@@ -0,0 +1,432 @@
...
+
+            bool first = true;
+            string expectedIndentation = null;
+            foreach (T element in elements)
+            {
+                SyntaxToken firstToken = GetFirstTokenForAnalysis(element);
+                SyntaxTrivia indentationTrivia = firstToken.LeadingTrivia.LastOrDefault();
+                string indentation = indentationTrivia.IsKind(SyntaxKind.WhitespaceTrivia) ? indentationTrivia.ToString() : string.Empty;
```
vweijsters: :exclamation:  This will mess up stuff if the first statement has the following format:
```csharp
namespace TestNamespace
{
  public class TestClass
  {
    /* obsolete */ public void OldMethod()
    {
    }

    public void NewMethod()
    {
    }
  }
}
```

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
```diff
@@ -0,0 +1,432 @@
...
+                    expectedIndentation = indentation;
+                    first = false;
+                    continue;
+                }
+
+                if (string.Equals(expectedIndentation, indentation, StringComparison.Ordinal))
+                {
+                    // This handles the case where elements are indented properly
+                    continue;
+                }
```
vweijsters: :question: What is the expected behavior if a mixed set of tabs and spaces is used?
I would expected that `<TAB><TAB><SPC> ` would be equal to `<TAB><SPC><SPC><SPC>` if indentation and tab size are both 2, otherwise it may be hard to explain the difference to users.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
sharwell: "Correctly" is vague in this situation. The following tests cover the current behavior:

* `TestDocumentationCommentBehaviorAsync` (`Enum3`)
* `TestEnumDeclarationAsync` (`Enum3`)
* `TestPropertyAccessorListAsync` (`Property3`)
* `TestIndexerAccessorListAsync` (`IContainer3`)
* `TestEventAccessorListAsync` (`Event3`)

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
sharwell: Yes it will. We need to modify `IsFirstInLine` to have an additional argument regarding whether to allow non-whitespace trivia to precede the token in question.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
sharwell: The code expects indentation of elements to match exactly.

vweijsters: Fair point, but users may not see the problem in the editor and they may therefore get annoyed. The very least we should do is document this behavior in the cause section of the documentation.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
vweijsters: Just checked those and I'm not happy with that behavior. I think we should use the indentation of the attribute list if it is the first item on the line and it is on the same line as the element that it belongs to.
If it is not on the same line as the element, the current behavior is correct.

vweijsters:
If the indentation of an element with documentation changes, the indentation of the documentation should change to lineup as well I think.
:question: Do you want to address that now or leave it as future improvement?

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
sharwell: The behavior you describe is what happens in #2034 when you have SA1138 enabled (in cases where absolute indentation is supported, which includes all locations where attributes can appear), SA1138 takes precedence over SA1137.

With that in mind, do you think we still need to change this?

sharwell:
> If the indentation of an element with documentation changes, the indentation of the documentation should change to lineup as well I think.

There are two issues here:

1. The indentation of documentation is not checked, so if the elements line up you won't get a warning regardless of how the documentation lines are indented.
2. The code fix only ever applies changes to a single line. This is really annoying with multi-line elements, but also really difficult to get right. It's a little better with SA1138 implemented since indentation is based on absolute expectations in many cases.

I was planning to leave both of these for future work since are more of a "it could do more than it does" than a "it's breaking my code".

vweijsters:
:question: What is the purpose of `TypeName1` in the `TestAttributeListAsync` test case? I don't see any differences between the test code and fixed test code for it.

vweijsters:
> I was planning to leave both of these for future work since are more of a "it could do more than it does" than a "it's breaking my code".

I'm OK with that. It will be hard to implement indeed.

vweijsters:
Done with my review.

sharwell:
> :question: What is the purpose of `TypeName1` in the `TestAttributeListAsync` test case? I don't see any differences between the test code and fixed test code for it.

Making sure it doesn't crash when you have 0 or 1 element to check.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
vweijsters: I'll get back on it after I've reviewed that one.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
sharwell: :arrow_right: Fixed.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
sharwell: If possible, I'd like to wait until we deal with the following items before changing this:

1. Not checking comments and trivia (like directives)
2. Sub-optimal code fix behavior for multi-line elements
3. Issue #2034 (SA1138)

sharwell:
@Noryoko I started the implementation of adjusting indentation of an entire element instead of just one line in #2034.

Noryoko:
Awesome! :smile:

PR #2031, opened 2016-01-07
Rename SA1652 to SA0001 and improve messaging

### LINKED ISSUE #2030: Proposal: Make SA1652 a "special" rule (SA0xxx)
I propose the following changes to SA1652 (Enable XML documentation output).

1. Move the rule to the "Special" category (presumably SA0001 or SA0002).
1. Change the title to "XML comment analysis disabled".
1. Change the message format to "XML comment analysis is disabled due to project configuration".
1. Change the analyzer to operate as a compilation end analyzer, which reports a single diagnostic for the project instead of one diagnostic per file.

Renumbering the rule will cause it to appear again for users who previously disabled SA1652, not realizing that it was alerting them to a problem that affected multiple rules.

sharwell:
Fixes #2030


vweijsters:
:+1: Looks good to me

Issue #2028, opened 2016-01-06
Analyzer doesn't read stylecop.json.


DirtyPiece:
Hello, sorry for bringing this one up again as I know it's been addressed a few times, but I'm having issues getting stylecop.json to be recognized by my project. Here's what I've tried:

1. Adding the file in quick actions and fixing the csproj manually by using the steps at https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/EnableConfiguration.md.

2. Changing the Build Action to be None, Additional Items, and Content; each time unloading and reloading the project. Unforunately, I haven't had success with any of them. I did notice that when I use Additional Items, the Quick Actions no longer suggest an option for me to generate a stylecop.json file for the project. With the other Build Actions, it gives me this suggestion.

3. Moving the `<AdditionalItems />` element into its own `<ItemGroup />`.

In all cases I still get a quick action that suggests the fix which is to replace my header which looks like this:
```
// =======================================================================
// <copyright file="AnalyticsDataProvider.cs" company="LAMA, LLC">
// Copyright 2015 Lama, LLC
// </copyright>
// =======================================================================
```

To this:
```
// =======================================================================
// <copyright file="AnalyticsDataProvider.cs" company="PlaceholderCompany">
// Copyright (c) PlaceholderCompany. All rights reserved.
// </copyright>
// =======================================================================
```

Here's the contents of my stylecop.json file:
```
{
  "$schema": "https://raw.githubusercontent.com/DotNetAnalyzers/StyleCopAnalyzers/master/StyleCop.Analyzers/StyleCop.Analyzers/Settings/stylecop.schema.json"
  "settings": {
    "documentationRules": {
      "companyName": "LAMA, LLC",
      "copyrightText": "Copyright 2015 {companyName}. All rights reserved."
    }
  }
}
```

I'm using Visual Studio Community 2015, version 4.6.00081 and RC3 of the analyzer. Any ideas are appreciated, thanks!

sharwell:
Can you try `AdditionalFiles` instead of `AdditionalItems`?

When you double click on **stylecop.json** for your project in Solution Explorer, does it open the correct file?

DirtyPiece:
Hello and thanks for the suggestion. Apologies, I meant to say AdditionalFiles in my original post. Here's what I've put in the csproj:

```
<ItemGroup>
    <AdditionalFiles Include="stylecop.json" />
  </ItemGroup>
```

[...]

PR #2032, opened 2016-01-08
Implement SA0002 (InvalidSettingsFile)

### LINKED ISSUE #2029: Implement SA0002: Invalid settings file
It would be nice to use a compilation end event to report a single instance of this violation if we are unable to deserialize **stylecop.json**. This would catch (at least) the following situations:

1. JSON syntax errors (e.g. #2028)
2. Semantic errors, such as using an array instead of a string or an unknown value for an enumerated type

This would *not* report a diagnostic if the file is not [configured correctly](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/EnableConfiguration.md).

sharwell:
Fixes #2029

Screenshot of a report when the syntax error matches the one described in #2028:

image

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Settings/SettingsFileCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Settings/SettingsFileCodeFixProvider.cs
@@ -23,8 +23,7 @@ namespace StyleCop.Analyzers.Settings
     [Shared]
     internal class SettingsFileCodeFixProvider : CodeFixProvider
     {
-        private const string StyleCopSettingsFileName = "stylecop.json";
-        private const string DefaultSettingsFileContent = @"{
+        internal const string DefaultSettingsFileContent = @"{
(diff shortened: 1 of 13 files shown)


vweijsters:
:question: Would it be possible to suppress the call stack dump, as it does not really provide any useful information?

sharwell:
@vweijsters It has the line and position numbers within the file where the problem occurred. It might not be exact, but it should be close. Also keep the following in mind:

* The description is hidden by default. You have to expand the line in the error list to see them.
* It's possible the stack trace could save time if/when issues are reported. In those cases, it's much easier to copy one that's provided than to explain to users how to attach the debugger and find out.

vweijsters:
@sharwell The line and position numbers are in the exception message, so there is no need to have the stack trace for that. Besides that, the stack trace is probably not really useful for debugging, as it will only be dumped for `JsonException`.

vweijsters:
:+1: Looks good otherwise

sharwell:
Updated to only show this:

image

vweijsters:
:+1:

PR #2033, opened 2016-01-08
Make SA0000 hidden but enabled by default

sharwell:
This change balances performance with usability of the workaround by ensuring that all diagnostics are reported for files that are open in the editor, without the performance overhead of enabling the workaround for every file in the project.

After SA0000 was made available, I found myself consistently enabling it in this manner via rule set files. However, it may not be obvious to other users that this is the best practice for handling this bug in Visual Studio 2015 Update 1. By changing the default values of the diagnostic to match the ones used in practice, everyone gets the benefit right out of the box. :+1:


vweijsters:
:+1:

PR #2038, opened 2016-01-11
Extract helper class SyntaxKinds and use where appropriate

sharwell:
This change was originally introduced in #2024, but it can be reviewed separately to simplify the reviews.


nbarbettini:
Looks good to me :+1:

PR #2045, opened 2016-01-13
Update SA1027 to support the UseTabs setting

### LINKED ISSUE #2035: Proposal: SA1027 should still reports tabs after non-whitespace with useTabs
~~**Blocked on the acceptance of #2039, which adds the option described here.**~~

#2039 Adds a new `useTabs` option which defaults to `false`. When this value is set to true, SA1027 (TabsMustNotBeUsed) is automatically disabled.

I propose updating the implementation when `useTabs` is true to only disable analysis of tabs used for indentation. In other words, tabs placed after the first non-whitespace character on a line would still be reported.

This is related to the OneTabIndent option described in #730, but not quite the same. This proposal only covers a specific case where the use of tab characters *always* produces strange behavior when the tab width changes.

### LINKED ISSUE #2037: Proposal: SA1027 should use tabs for indentation when useTabs is true
~~**Blocked on the acceptance of #2039, which adds the option described here.**~~

Currently SA1027 checks for tabs and converts them to spaces. When `useTabs` is true, I propose automatically having SA1027 switch to checking for spaces used for indentation, reporting a diagnostic, and automatically converting them to tabs according to the `tabSize` property.

sharwell:
This is currently blocked on the acceptance of proposals #2035 and #2037 (both are implemented by this pull request).

Fixes #2035
Fixes #2037


StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UseTabsUnitTests.cs
```diff
@@ -0,0 +1,259 @@
...
+                "        }\r\n" +
+                "    }\r\n";
+
+            var fixedTestCode =
+                "\tpublic class Foo\r\n" +
+                "\t{\r\n" +
+                "\t\tpublic void Bar()\r\n" +
+                "\t\t{\r\n" +
+                "\t\t\t/*\r\n" +
+                "            Comment     1\r\n" +
```
sharwell: :exclamation: This is not the behavior we want.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1027UseTabsUnitTests.cs
```diff
@@ -0,0 +1,259 @@
...
+            await this.VerifyCSharpFixAsync(testCode, fixedTestCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <inheritdoc/>
+        protected override string GetSettings() =>
+            @"
+{
+    ""settings"": {
+        ""indentation"": {
+            ""useTabs"": true
```
sharwell: :memo: We should also have some tests for cases where `tabSize` is not 4.

sharwell: :arrow_right: Added `SA1027AlternateIndentationUnitTests`.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027UseTabsCorrectly.cs
```diff
@@ -0,0 +1,268 @@
...
+                TextSpan next = sortedTextSpans[nextIndex];
+                if (current.End < next.Start)
+                {
+                    // Increment currentIndex this iteration
+                    currentIndex++;
+
+                    // Only increment nextIndex this iteration if necessary to ensure nextIndex > currentIndex on the
+                    // next iteration. At this point we already incremented currentIndex, but haven't incremented
+                    // nextIndex.
+                    if (currentIndex > nextIndex)
```
vweijsters: :bug: I think this should be `if (currentIndex < nextIndex)`

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027UseTabsCorrectly.cs
```diff
@@ -0,0 +1,268 @@
...
+
+            int lastLineStart = 0;
+            for (int startIndex = 0; startIndex < length; startIndex++)
+            {
+                if (startIndex == nextExcludedSpan.Start)
+                {
+                    startIndex = nextExcludedSpan.End - 1;
+                    excludedSpanIndex++;
+                    nextExcludedSpan = excludedSpanIndex < excludedSpans.Length ? excludedSpans[excludedSpanIndex] : lastExcludedSpan;
+                    continue;
```
vweijsters: :thought_balloon: The continue statement is unnecessary, as it is impossible to have adjacent excluded spans, because they would have been joined by `ReduceTextSpans`. If `startIndex` is set to `nextExcludedSpan.End`, the flow can just continue.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027UseTabsCorrectly.cs
sharwell: I think it's worse than that actually... seems to be missing the following after incrementing `currentIndex`:

```csharp
sortedTextSpans[currentIndex] = next;
```

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1027UseTabsCorrectly.cs
sharwell: Unless the excluded span ends at the end of the file?

vweijsters: Indeed :blush:

vweijsters:
:+1: Looks good to me

adeel41:
Will this be released soon?

sharwell:
@adeel41 It's now on NuGet (1.1.0-beta001). I'll be writing up the release notes this evening.

PR #2051, opened 2016-01-20
Fixed SA1205 issue with nested partial types

### LINKED ISSUE #2040: SA1205 fired wrongly
I have some auto-generated code whose structure looks like that:

```csharp
    internal static partial class TestPartial
    {
        private sealed partial class PartialInner
        {
            public int Do()
            {
                return 2;
            }
        }
    }
```

I get SA1205 (Partial elements must declare an access modifier) when I have specified private access. The proposed fix is to prepend internal before private sealed partial which obviously leads to a compilation error. The issue does not happen if the inner class is internal / public but only when it is private.

The other thing is that styleCop does not care if there is a GeneratedCode attribute or not on the outer or inner class.

vweijsters:
Fixes #2040


pdelvo:
:+1:

Issue #2010, opened 2015-12-28
Non-deterministic fix all behavior in testing


sharwell:
We're seeing some non-deterministic Fix All behavior in the unit tests:

https://ci.appveyor.com/project/sharwell/stylecopanalyzers/build/1.0.2431

PR #2052, opened 2016-01-20
SA1513 will now allow a closing brace to be followed by a four slash …

### LINKED ISSUE #2041: SA1513 Fired when old StyleCop did not complain
I like that in the old StyleCop (4.7) you could put //// comments anywhere in the code and no warning would be raised. With StyleCop.Analyzers rc3 I get a SA1513 with the following code

```csharp
            public int Do(int i)
            {
                if (i > 2)
                {
                    return 1;
                }
                //// else if (i == 2)
                //// {
                ////     return 2;
                //// }

                return 0;
            }
```

whereas I did not get any warning from StyleCop 4.7. I preferred the old way.

vweijsters:
…line comment

Fixes #2041

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
@@ -796,6 +796,36 @@ public class Program
             await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
         }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


vweijsters:
@sharwell Can you check the build failure here as well. It looks to me that `StyleCop.Analyzers.Test.LayoutRules.SA1501UnitTests.TestNoSA1503IfStatementsAsync` is not deterministic.

sharwell:
> Can you check the build failure here as well. It looks to me that StyleCop.Analyzers.Test.LayoutRules.SA1501UnitTests.TestNoSA1503IfStatementsAsync is not deterministic.

#2010

pdelvo:
:+1:

PR #2053, opened 2016-01-21
Improve performance of IsPrecededByWhitespace

pdelvo:
When analyzing performance the IsPrecededByWhitespace was using about ~3.5% total time when running our analyzers on Roslyn.sln.

After this change it is under 0.1%

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
@@ -62,11 +62,11 @@ namespace StyleCop.Analyzers.Helpers
         /// Gets a value indicating whether the <paramref name="token"/> is preceded by a whitespace.
         /// </summary>
         /// <param name="token">The token to process.</param>
-        /// <param name="cancellationToken">A <see cref="CancellationToken"/>.</param>
+        /// <param name="cancellationToken">The cancellation token that the operation will observe.</param>
         /// <returns>true if token is preceded by a whitespace, otherwise false.</returns>
         internal static bool IsPrecededByWhitespace(this SyntaxToken token, CancellationToken cancellationToken)
         {
-            // Perf directly access the text instead of trivia
+            // Perf: Directly access the text instead of the trivia.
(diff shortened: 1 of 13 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
```diff
@@ -64,14 +64,15 @@ internal static bool IsLastInLine(this SyntaxToken token)
...
+
+            if (pos < 0)
             {
-                return triviaList.Last().IsKind(SyntaxKind.WhitespaceTrivia);
+                return false;
             }
 
-            triviaList = token.GetPreviousToken().TrailingTrivia;
-            return triviaList.Count > 0 && triviaList.Last().IsKind(SyntaxKind.WhitespaceTrivia);
+            return char.IsWhiteSpace(token.SyntaxTree.GetText()[pos]);
```
sharwell: :exclamation: `GetText(cancellationToken)`

:question: Should we handle the case where `SyntaxTree` is null (e.g. a token created by `SyntaxFactory.Token` that isn't in the tree yet)?

pdelvo: Fixed

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
```diff
@@ -61,17 +62,19 @@ internal static bool IsLastInLine(this SyntaxToken token)
         /// Gets a value indicating whether the <paramref name="token"/> is preceded by a whitespace.
         /// </summary>
         /// <param name="token">The token to process.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/>.</param>
```
sharwell: :bulb: Description should be:

> The cancellation token that the operation will observe.

pdelvo: fixed

sharwell:
:+1:

PR #2070, opened 2016-02-03
Implemented single pass fix all for SA1651

### LINKED ISSUE #1986: SA1651 fix all takes multiple iterations
The operation tested in `SA1651UnitTests.TestFormattingPreservedAsync` currently takes two iterations of the Fix All operation. Since both violations are present in the original source (i.e. the first iteration does not introduce the violation which is corrected in the second), the Fix All operation should be able to fix both.

vweijsters:
Fixes #1986


sharwell:
:+1:

PR #2075, opened 2016-02-06
Updated KnownChanges.md with entry for SA1210

### LINKED ISSUE #2071: Update known changes for SA1210
As mentioned in #896, the `global::` prefix is used by SA1210 when determining the proper ordering. This appears to be a deviation from StyleCop Classic (@vweijsters can you confirm?), which means it should be documented in **KnownChanges.md**.

vweijsters:
Fixes #2071


sharwell:
:+1:

PR #2076, opened 2016-02-06
Fixed SA1610 to ignore <inheritdoc/> for properties that do not overr…

### LINKED ISSUE #1945: SA1610 inheritdoc analysis does not consider overrides
The following code should report SA1610, but no warning is reported.

```csharp
/// <value>
/// </value>
/// <inheritdoc/>
public string Property => "P";
```

vweijsters:
…ide / implement an interface.

Fixes #1945

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1609UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1609UnitTests.cs
@@ -41,13 +41,20 @@ public class ClassName
         public async Task TestPropertyWithInheritedDocumentationAsync()
         {
             var testCode = @"
+public class BaseClassName
+{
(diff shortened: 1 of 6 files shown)


sharwell:
It looks like I failed to explain the issue correctly. In the example code in #1945, the problem lies in the fact that the `<value>` element will prefer to use one included in the comment itself (if any), and only fall back to using the inherited value if none is present. Our analysis currently assumes that the inline value is ignored if `<inheritdoc/>` is specified.

vweijsters:
:question: That would effectively mean that we can remove any check on `<inheritdoc/>` if the `<value>` element is present, right?

sharwell:
@vweijsters At least for SA1610, yes that is correct.

vweijsters:
@sharwell I've updated the code to reflect the discussion above. Can you have a look and see if this is what you intended?

PR #2077, opened 2016-02-06
Partial types / methods no longer accept <inheritdoc/>

### LINKED ISSUE #1946: SA1607 inheritdoc analysis does not consider overrides
The following code should report SA1607, but no warning is reported.

```csharp
/// <summary>
/// </summary>
/// <inheritdoc/>
public partial class ClassName
{
}
```

In addition, the following code should report SA1607, but no warning is reported.

```csharp
/// <content>
/// </content>
/// <inheritdoc/>
public partial class ClassName
{
}
```

vweijsters:
Fixes #1946


sharwell:
See my comment in #2076. This doesn't address the issue I was trying to describe in #1946, and I actually think we should continue to allow inheritdoc to be used on these elements.

PR #2097, opened 2016-03-07
Fixed static class collisions in SA1101

### LINKED ISSUE #2093: SA1101 misfires when member name coincides with type name
In the below sample (which references `System.Drawing.dll`), the use of the static method `Region.FromHrgn` causes SA1101 to inappropriately report that `this.` should be used in front of the `Region` type. Evidently it is confusing the `Program.Region` member with the `Region` type.

```csharp
using System;
using System.Drawing;

internal class Program
{
    public Region Region => Region.FromHrgn(IntPtr.Zero);
}
```

vweijsters:
Fixes #2093


sharwell:
:+1:

PR #2099, opened 2016-03-10
Add SX Alternative Rule documentation

felickz:
I saw some mention of SX rules in comments and dug to find out what they were... providing documentation to help inform others.


vweijsters:
Thanks for submitting this. It looks good to me :+1:

sharwell:
:memo: In each of these cases, it is likely that the SX rule will eventually be eliminated in favor of configuration settings. But for now this is good to have :+1:

PR #2102, opened 2016-03-13
Update to Roslyn 1.2

sharwell:
* Update to xUnit.net 2.2.0-beta1
* Update to Roslyn 1.2.0-rc
* Replace generated code detection with Roslyn's built-in functionality
* Replace StyleCopSettings cache with Roslyn's built-in functionality
* Enable concurrent analyzer execution
* Remove workarounds for Roslyn 1.0 and/or 1.1


vweijsters:
Looks good so far. What needs to be done before this is ready for merge?

vweijsters:
`version.ps1` should probably be updated to `1.1.0-dev`

sharwell:
> What needs to be done before this is ready for merge?

* It would be good to wait until Visual Studio 2015 Update 2 is available. Otherwise people will not be able to debug this branch inside the IDE.
* I'd like to spend more time testing performance of this branch.

sharwell:
My initial performance testing did not yield the expected results. For the test, I simply ran the following command using dotnet/roslyn@01ba99a61c585a7ff7214d799095f010a017a486.

> StyleCopTester ..\..\..\roslyn\Roslyn.sln

| Commit | Analysis Time (sec) |
| --- | --- |
| 7e69f3ea7021ade6a966fcb8654fb014ecf71d00 | 173 |
| 64d39356d2f4cbdbd2ef7caa65cbe747583c7020 | 244 |
| 44bc1738588364871b46964cc3f9ad7365c6a226 | 251 |

pdelvo:
Thats weird. @mavasani Are we doing something wrong here or is this a regression in roslyn? Thats 45% slower in 1.2 usi g the new APIs compared to 1.0.

mavasani:
We have made a bunch of perf fixes post RC. So I would recommend trying out later bits. Additionally, if you can provide ETL trace with RC bits, we can investigate further.

Did you also measure perf with RC bits but not enabling new features? That might give us more context.

mavasani:
Oh, I see that you did measure with and without new APIs. So seems unrelated to new APIs.

sharwell:
@mavasani Using 1.2.1 did improve things a bit. Haven't run the profiler yet to see where the remaining problem spots are (still *much* slower than 1.0.0 was).

mavasani:
Thanks @sharwell - would also help if you can let me know how to run the test locally (if its via StyleCopTester, I can dig out your old response).

Also adding @basoundr, @Srivatsn and @dotnet/roslyn-analysis as an FYI

sharwell:
@mavasani I'm using StyleCopTester against a specific commit of Roslyn as described above.

mavasani:
@sharwell do you have numbers against 1.1.0 bits of Roslyn

mavasani:
@sharwell I notice that StyleCopTester isn't turning on concurrent execution. Can you change [this](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCopTester/Program.cs#L483) to use the [new overload](http://source.roslyn.io/#Microsoft.CodeAnalysis/DiagnosticAnalyzer/DiagnosticAnalyzerExtensions.cs,28) and ensure that the `ConcurrentAnalysis` is enabled via `CompilationWithAnalyzersOptions`?

sharwell:
@mavasani I enabled concurrent execution in cc7deaf8bd4973e4b774344f18b5fdeb38041b9a, but that just slowed it down from 244 seconds to 254 seconds.

mavasani:
@sharwell Would it be possible to measure the performance against Roslyn 1.1.0 bits? I am guessing this regression happened with the new IDE analyzer driver added in Update1.

I am going to attempt to repro this locally and confirm the root cause. Meanwhile I have also filed https://github.com/dotnet/roslyn/issues/10530 to track the work on our side.

sharwell:
I can get those numbers tonight. Essentially it will be a test like commit 64d3935, but using the 1.1 packages instead of 1.2.0-rc.

sharwell:
@mavasani I ran the test with Roslyn 1.1.1. Some main items to note:

1. Solution load in Roslyn 1.2.1 (`OpenSolutionAsync`) takes 136% longer than it takes in Roslyn 1.1.1 (as in more than doubled).
2. Analysis time with Roslyn 1.1.1 is 460 seconds. So 1.2.1 is much faster than 1.1.1 but still loses out to 1.0.0.

mavasani:
Thanks a lot @sharwell 

1. Can you file a separate issue for OpenSolutionAsync regression?
2. This basically confirms that the regression happened with 1.1.1, but went unnoticed because neither the IDE diagnostic service nor the batch compiler uses this API for diagnostic computation. The background analysis in the IDE probably takes much longer to compute all diagnostics in Update1/Update2 then in RTM - we are restoring our rolling diagnostics perf tests to avoid such regressions in future.

I am working on https://github.com/dotnet/roslyn/issues/10530 which should improve the analyzer execution time for both the IDE analyzer execution and this public API - I'll measure the analysis time with RTM, Update2 and bits with the fix for this issue and update here.

PR #2067, opened 2016-01-31
Improved SA1501 codefix to allow for a single pass fix-all

vweijsters:
Closes #2010 

I've removed the non-determinism in the SA1501 code fix by making it a single pass fix-all.

In order to make that work properly, the analyzer now generates more warnings, so that the code fix can use all the warnings to make a proper fix. I've tried a new concept here of reporting only one warning per code line as warning and reporting the others as hidden, to minimize the impact for the end user.

The code fix was refactored to only use a token replacement strategy, instead of the previous node + token replacement strategy. This was necessary to allow for interaction between fix steps in the fix-all scenario.

PR #2116, opened 2016-04-01
Remove link to contributing guide

accidentaldeveloper:
The contributing guide links to the contributing guide. I think that explains it.


accidentaldeveloper:
I can't imagine why the CI build would fail when checking this change.

vweijsters:
@accidentaldeveloper The build fails because the FixAll code fix for SA1501 is non-deterministic.
(see also #2010 and #2067)

pdelvo:
:+1:

PR #2123, opened 2016-04-10
Include .g.cs files as generated.

BrianSakhai:
Refit (https://github.com/paulcbetts/refit) uses the ".g.cs" file extension to indicate the file is auto-generated. I know the file name detection was simplified until examples could be shown for each. I hope this is an acceptable example.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/GeneratedCodeAnalysisExtensions.cs
@@ -157,7 +157,7 @@ namespace StyleCop.Analyzers
         {
             return Regex.IsMatch(
                 Path.GetFileName(filePath),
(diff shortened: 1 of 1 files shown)


vweijsters:
The change looks good to me and I'm OK with adding this.
This might be obsolete once we merge in the roslyn-1.2 stuff

sharwell:
A similar situation appeared in the past. The solution there was to submit a pull request that updated the code generator to include the proper header in the generated code (GitTools/GitVersion#537), which automatically caused the files to be excluded from analysis in both this project and others which use similar exclusion mechanisms.

sharwell:
For now I'm going to mark this as blocked on #2102. If we merge that pull request prior to releasing version 1.1 Beta 1, then this pull request becomes obsolete as all exclusions handling is performed by Roslyn itself instead of StyleCop Analyzers.

sharwell:
Closing as no longer necessary

PR #2136, opened 2016-05-19
Replaced message parameter of CSharpCompilerError with WithMessage()

vweijsters:
As suggested by @sharwell in #743


pdelvo:
:+1:

PR #2137, opened 2016-05-21
Update DotNetCli.md with RC2 detail

Romanx:
Contains name changes in project.json from RC2 and a clarification on schema issue


pdelvo:
:+1:

PR #2139, opened 2016-05-22
Fixed SA1119CodeFix for preprocessor directives

### LINKED ISSUE #2069: Code fix  for SA1119 doesn't work well for preprocessor directives.
This code:
```
    public class Program
    {
        public static void Main(string[] args)
        {
#if(DEBUG)

#endif

        }
    }
```
got SA1119 warning but after applying a fix this is the new code:
```
    public class Program
    {
        public static void Main(string[] args)
        {
#ifDEBUG

#endif

        }
    }
```
wchich gives a compiler error. It should be `#if DEBUG` not `#ifDEBUG`

vweijsters:
Fixes #2069


sharwell:
What does this code do for the following?

```csharp
#if (DEBUG)
#endif
```

vweijsters:
It inserts once space too many. I had assumed that the separating space would be part of the leading trivia for the open paren, but it isn't. 

It's fixed now.

pdelvo:
:+1:

PR #2140, opened 2016-05-22
Multi-file code fix verification

otac0n:
Adds support for verifying code fixes that affect more than one file.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/FileMayOnlyContainTestBase.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/FileMayOnlyContainTestBase.cs
@@ -3,6 +3,7 @@
 
 namespace StyleCop.Analyzers.Test.MaintainabilityRules
 {
+    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
     using TestHelper;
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/FileMayOnlyContainTestBase.cs
```diff
@@ -32,18 +33,30 @@ public abstract class FileMayOnlyContainTestBase : CodeFixVerifier
...
 
             DiagnosticResult expected = this.CSharpDiagnostic().WithLocation(4, this.Keyword.Length + 2);
 
             await this.VerifyCSharpDiagnosticAsync(testCode.Replace("%1", this.Keyword), expected, CancellationToken.None).ConfigureAwait(false);
-            await this.VerifyCSharpDiagnosticAsync(fixedCode.Replace("%1", this.Keyword), EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+
+            foreach (var code in fixedCode)
+            {
+                await this.VerifyCSharpDiagnosticAsync(code.Replace("%1", this.Keyword), EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            }
```
vweijsters: :thought_balloon: This code construction is repeated a lot. Might be a good idea to add a VerifyNoCSharpDiagnosticsAsync method that will accept either a source file or a list of sources.

vweijsters:
:+1:  Looks good to me, only one small remark.

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/FileMayOnlyContainTestBase.cs
otac0n: There are already some methods in the CodeFixVerifier and DiagnosticVerifier that take arrays of strings. These calls treat the array as multiple files in a single project.

I feel that adding an overload that takes an array of strings, but treats each independently, would confuse the semantics of the verifiers.

otac0n: Turns out, the overload exists.  I'll update.

Issue #1562, opened 2015-09-27
Implement SA1139: Use literal suffix notation instead of casting


berkeleybross:
New rule proposal: Use literal suffix notation instead of casting

_Category_: Readability
_Rule_: When a numeric literal is immediately cast to another numeric type, it should be replaced with a literal suffix.
_Rationale_: Improves readability of code, and (presumably) run time casts.

Bad:

``` CSharp
long foo = (long)50;
```

Good:

``` CSharp
long foo = 50L;
```

See https://msdn.microsoft.com/en-us/library/aa691085(v=vs.71).aspx and https://msdn.microsoft.com/en-us/library/aa664674(v=vs.71).aspx for the specification for integer/real literals.

Things that might need considering:

The literal suffix can be upper and lower case. I would prefer uppercase, (and if you read the integer literal specification above, it recommends this).

This should work on nullable types as well:

``` CSharp
long? foo = (long?)50;
```

Should be fixed to 

``` CSharp
long? foo = 50L;
```

vweijsters:
:+1:

nbarbettini:
:+1:

GregReddick:
:+1: See #1563, though, on the capitalization of the literal specification.

otac0n:
:+1:

sharwell:
:memo: Make sure to not report a warning on the following code:

```
var foo = (long?)50;
```

tmaczynski:
I've decided to make some contributions to DotNetAnalyzers. I've done a draft of analyzer and tests and I'll implement code-fix later.

I've noticed that in some cases (e.g. when a field `long x = (long)1` is declared) this analysis is overlapping with `IDE0004 C# Cast is redundant`. Codefix for IDE0004 deletes cast and relies on an explicit casting. 

I think that SA1139 should be reported in such case and another codefix should be provided (which creates a literal of a desired type) but I wanted to ask about your thoughts before implementing it.

I'll do a pull request when I'm finished with this analyzer, my [first implementation](https://github.com/tmaczynski/StyleCopAnalyzers/compare/8483314b3282b831ae1feba3f4070b0d0e853de1...tmaczynski:f08086ba8ebf699ec86a847e00f800ce40cb075e) is in my forked repository.

tmaczynski:
I was extremely busy recently with my full time job, but I’ve found some time to implement first draft that is worth reviewing and (hopefully) checking in into main repo.

[...]

PR #2148, opened 2016-06-08
implementation of SA1139UseLiteralSuffixNotationInsteadOfCasting 

tmaczynski:
First working and tested implementation of #1562


pdelvo:
I reviewed this PR and this looks mostly fine. I don't like the use of regular expressions. This would also have to be updated for C#7 with its binary literals and literal seperators. I played around with it a little bit and I came up with a solution that does not rely on them. I sent you a PR (tmaczynski/StyleCopAnalyzers#1) so you can have a look.

tmaczynski:
Hi,
Thank you for review, I've merged you changes. I've also added one trivial change to this pull request (4434257).

I really like that you used SemanticModel API and deleted my extension methods (which are redundant).
If you prefer imperative loop to regex, I think that all other changes also make sense - especially that you code works better when C# 7 code is analysed.

tmaczynski:
I've resolved conflicts between tmaczynski:sa1139 topic branch and master.

tmaczynski:
Could you merge my changes or give me more feedback?

pdelvo:
Loogs good to me :+1:

PR #2165, opened 2016-07-05
Updated SA1305 to also handle parameters

### LINKED ISSUE #1928: SA1305 for parameters
```cs
public void foo(bool bDelete)
{
	// this one is caught but not the parameter
	bool bDoSomething;
}
```

bjornhellander:
Fixes #1928

Updated analyzer and known changes. Added tests.


pdelvo:
:+1:

PR #2168, opened 2016-07-14
Fix that SA1115 is reported on multi-line parameters/arguments.

### LINKED ISSUE #2159: SA1115 should allow attributes between parameters
The following code should not produce SA1115, but it currently does on the third and fourth `Required` parameter attributes:

`> Install-Package NConsoler`

```
using NConsoler;

namespace MyConsoleApplication
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Consolery.Run(typeof(Program), args);
        }

        [Action]
        public static void DoWork(
            [Required]
            string param1,
            [Required]
            string param2,
            [Required]
            string param3,
            [Required]
            string param4)
        {
            Console.WriteLine("Hello World!");
        }
    }
}
```

Thank you.

pdelvo:
I also removed some dublicated code.

Fixes #2159.


sharwell:
:+1:

PR #2175, opened 2016-07-18
Update to Roslyn-1.3.2 packages

vweijsters:
Updated the Roslyn NuGet packages to 1.3.2.


sharwell:
:question: What do we gain from this? Users will already be able to use Roslyn 1.3 if they choose to update (e.g. to Visual Studio 2015 Update 3), but as long as we compile against Roslyn 1.2 they'll be able to use Visual Studio 2015 Update 2 as well.

vweijsters:
No particular gain in mind, just force of habit I guess. I'll close the pull request.

PR #2201, opened 2016-09-12
Remove individual contributor name from file headers

sharwell:
@tmaczynski This change removes the contributor comment you included in new files in #2148. My specific concerns here are the following:

1. You are the only contributor mentioned in this manner. Many other users have contributed code, but files they created are not identified in this way.
2. While you did *create* the file, this leaves a big question regarding how to handle *updates* to the file.

So far I've found it easier to include contributors in the following ways:

1. The source control history
2. An explicit mention of people involved with each release on the [release notes](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/releases) pages

:question: Do you have any objection to me removing these lines and proceeding with the existing mechanism for acknowledging contributions?


tmaczynski:
@sharwell not at all, feel free to delete those lines. As I wrote in #1562

> I don’t know whether contributors are listed somewhere. I added comments listing me as a contributor in my pull request, but feel free to remove them if you want.

Right now I know how you recognize contributors, thanks for clarifying this.

sharwell:
@tmaczynski Thanks. I still didn't want to make a change like this without being completely transparent. :smile:

PR #2207, opened 2016-10-02
Remove exclusion of helper types in SA1402

### LINKED ISSUE #1756: Proposal: Remove SA1402 exclusion of "helper types"
Currently SA1402 is written such that it only applies to classes. However, SA1649 applies to the first _type_ in a file. As such, there is a fundamental conflict between rules SA1201, SA1402, and SA1649.

I propose modifying SA1402 to no longer allow a file to contain more than one top-level _type_, regardless of the kind of type. The primary rationale is that a type which is directly related to the top-level type such that it shouldn't be removed from the file should actually be a nested type, which is already supported without problems by the rules mentioned above.

bjornhellander:
Fixes #1756

There are two questions in the issue that were not answered:
- Should SA1649 also be affected by the same "topLevelTypes" setting? I have not made that change, so the configuration of what declarations to consider is now inside the "maintainabilityRules" part of the settings file. If it should affect SA1649 as well, it might belong somewhere else...
- Should the analyzer class be renamed to reflect that it no longer affects only classes? I have made that change.


bjornhellander:
The decrease in coverage seems to be related to a bunch of *.Designer.cs files that are not even included in the coverage report for the current master. Have I messed up or is the report broken? :-)

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1402SettingsConfiguration.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        ConfigureAsTopLevelType,
+
+        /// <summary>
+        /// Provide custom settings that configure the tested type as not being a top level type
+        /// </summary>
+        ConfigureAsNonTopLevelType,
+    }
+
+    [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1649:File name must match first type name.", Justification = "Extension method for enum above.")]
```
vweijsters: ❗️ Instead of suppressing this could be moved to its own file, removing the need to suppress the diagnostic. 

As a general rule this project eats its own dog food, meaning that the source code should be StyleCop.Analyzers compliant without suppressions.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1402CodeFixProvider.cs
```diff
@@ -56,32 +56,14 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
-            else
-            {
-                if (delegateDeclarationSyntax.TypeParameterList?.Parameters.Count > 0)
-                {
-                    extractedDocumentName += "`" + delegateDeclarationSyntax.TypeParameterList.Parameters.Count;
-                }
-            }
-
-            extractedDocumentName += ".cs";
+            var extractedDocumentName = GetExtractedDocumentName(memberDeclarationSyntax) + ".cs";
```
vweijsters: ❗️ This should use `FileNameHelpers.GetConventionalFileName`, to have the same naming conventions as for SA1649, preventing a potential clash. The `GetExtractedDocumentName` method can then be removed.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/NamedTypeHelpers.cs
```diff
@@ -76,6 +77,15 @@ internal static bool IsContainedInNativeMethodsClass(SyntaxNode syntax)
             return false;
         }
 
+        internal static string GetNameOrIdentifier(MemberDeclarationSyntax member)
+        {
+            string name = null;
+            name = name ?? (member as TypeDeclarationSyntax)?.Identifier.Text;
+            name = name ?? (member as EnumDeclarationSyntax)?.Identifier.Text;
+            name = name ?? (member as DelegateDeclarationSyntax)?.Identifier.Text;
+            return name;
```
vweijsters: ❗️ This code can be done more efficiently with a `switch` statement on `member.Kind()`

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1402FileMayOnlyContainASingleType.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        private static IEnumerable<MemberDeclarationSyntax> GetTypeDeclarations(SyntaxNode root, StyleCopSettings settings)
+        {
+            var typeDeclarations = root.DescendantNodes(descendIntoChildren: node => ContainsTypeDeclarations(node)).OfType<MemberDeclarationSyntax>().ToList();
+            var relevantTypeDeclarations = typeDeclarations.Where(x => IsRelevantType(x, settings)).ToList();
+            return relevantTypeDeclarations;
+        }
+
+        private static bool ContainsTypeDeclarations(SyntaxNode node)
```
vweijsters: 💭 Consider renaming this to `ContainsTopLevelTypeDeclarations`, as the current name implies that a class containing subclasses should also return true.

bjornhellander:
@vweijsters Thanks for the feedback! Updated more or less according to your suggestions.

PR #2209, opened 2016-10-09
Re-wrote SA1025CodeFixProvider in the same way as SA1028CodeFixProvider.

### LINKED ISSUE #1398: Rewrite SA1025 code fix to use more efficient text manipulation
The performance of the SA1025 code fix is already good, but it would be _even better_ if it performed the fixes using direct `TextChange` operations in the same way that the SA1028 code fix already does.

bjornhellander:
Fixes #1398.

Ooookej. So, I re-wrote the code fix for SA1025 in the same spirit as SA1028. At least as far as I understood... The problem I'm having is that this did not give me any improvements, actually the opposite.

I tried it with StyleCopTester and a generated solution with lots of files. 138000 violations of SA1025! :-)
With this change, I get around 2000 fixes per second. Without it I get around 50000 per second. If I make the tested solution smaller, the performance difference decreases, but it's still better without the change than with it.

As a comparision, I tried a solution which instead of SA1025 had the same number of SA1028 and again I got around 2000 fixes per second. So, I don't see any performance related reasons for merging this change. Or are we optimizing for the case with just a small number of violations?

Could anyone else try it out and see if you get different behaviour? Or even better, point to where I have made a mistake... :-)

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/SpacingRules/SA1025CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/SpacingRules/SA1025CodeFixProvider.cs
@@ -12,7 +12,7 @@ namespace StyleCop.Analyzers.SpacingRules
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CodeActions;
     using Microsoft.CodeAnalysis.CodeFixes;
-    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.Text;
 
     /// <summary>
     /// Implements a code fix for <see cref="SA1025CodeMustNotContainMultipleWhitespaceInARow"/>.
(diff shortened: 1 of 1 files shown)


sharwell:
@bjornhellander Those results are really interesting. Can you clarify what command line arguments you passed to StyleCopTester for the test?

bjornhellander:
Ah, yes... Documenting that in my comment would have been the smart thing do, right?
Instead, I will do the experiment again :-)

bjornhellander:
These arguments match the execution time from my initial comment:
/id:SA1025 /fixall

sharwell:
Well, that's the right way to measure it. I wonder if the problem lies in our use of `GetRootAsync()` instead of just returning the document with the changed text directly.

sharwell:
@bjornhellander Can you try merging the **faster-fixer** branch from my fork into your local copy and re-running the experiment?

bjornhellander:
OK, here's what I found:
master = 54000 fixes/second
faster-fixer = 8000 fixes/second
my pr = 2000 fixes/second
my pr + faster-fixer = 2000 fixes/second

For comparison, I also measured SA1500 (which uses ReplaceTokens):
master = 9000 fixes/second
faster-fixer = 5000 fixes/second
my pr = 9000 fixes/second

(Merging master into my pr did not have any effect)

bjornhellander:
Hmmm, both our changes replaces text (WithDocumentText/WithChanges). Could ReplaceTrivia and ReplaceTokens simply be faster, because they don't force a complete re-compilation of the document? ReplaceTrivia might not have to do anything except modify whitespace trivia in the syntax tree, whereas WithChanges and WithDocumentText has do redo both scanning and parsing... And ReplaceTokens only needs to redo the parsing...

sharwell:
@bjornhellander Thanks for the detailed investigation. For now I'm going to close the issue as **wontfix**. If something changes in Roslyn in the future to make direct text manipulation the faster option, we can switch to using it at that time.

PR #2225, opened 2016-11-15
Improve performance of batch fixers that operate on text

sharwell:
Previously, DocumentBasedFixAllProvider required implementations to provide
a SyntaxNode. This change eliminates that requirement, improving performance
of low-level batch fixers that manipulate text directly.


sharwell:
Marking **wontfix** based on evaluations by @bjornhellander in #2209.

PR #2229, opened 2016-11-20
Implemented allowConsecutiveUsings option

### LINKED ISSUE #2184: Add setting to relax SA1519 to allow consecutive with braces omitted on all but the last one
#2180 modified the rule to disallow this:

```
using (var a = GetDisposable())
using (var b = GetAnotherDisposable())
{
}
```

This was permitted in classic StyleCop. It would be helpful to have a setting that allows this behaviour to be retained.

vweijsters:
Fixes #2184

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1519BracesMustNotBeOmittedFromMultiLineChildStatement.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1519BracesMustNotBeOmittedFromMultiLineChildStatement.cs
@@ -9,6 +9,7 @@ namespace StyleCop.Analyzers.LayoutRules
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
+    using Settings.ObjectModel;
     using StyleCop.Analyzers.Helpers;
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1519BracesMustNotBeOmittedFromMultiLineChildStatement.cs
```diff
@@ -108,9 +109,15 @@ private static void HandleForEachStatement(SyntaxNodeAnalysisContext context)
             CheckChildStatement(context, forEachStatement.Statement);
         }
 
-        private static void HandleUsingStatement(SyntaxNodeAnalysisContext context)
+        private static void HandleUsingStatement(SyntaxNodeAnalysisContext context, StyleCopSettings settings)
         {
             var usingStatement = (UsingStatementSyntax)context.Node;
+
+            if (settings.LayoutRules.AllowConsecutiveUsings && (usingStatement.Statement is UsingStatementSyntax))
```
pdelvo: Would this allow something like this?
```csharp
using(...)
for(...)
{
// ...
}
```
What is the stylecop classic behavior on this?

documentation/Configuration.md
```diff
@@ -313,6 +314,14 @@ file are handled. The `newlineAtEndOfFile` property supports the following value
 * `"require"`: Files are required to end with a single newline character
 * `"omit"`: Files may not end with a newline character
 
+### Consecutive using statements without braces
```
pdelvo: This should probably specify if using statements are allowed without braces before other similar constructs like `for`

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1519BracesMustNotBeOmittedFromMultiLineChildStatement.cs
vweijsters: That would not be allowed, as the code explicitly checks if the child statement is a using statement.

StyleCop classic also does not allow that construction.

documentation/Configuration.md
vweijsters: I've added a clarification for this

PR #2242, opened 2016-12-06
Implemented SA1628.

### LINKED ISSUE #148: SA1628: DocumentationTextMustBeginWithACapitalLetter

### LINKED ISSUE #149: SA1629: DocumentationTextMustEndWithAPeriod
~~http://www.stylecop.com/docs/SA1629.html~~
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1629.md

dlemstra:
This pull requests implements SA1628 and adds a CodeFixProvider for SA1628 and this will resolve #148. This is my first attempt to write an Analyzer and a CodeFixProvider so please let me know if I can make some improvements. I am not sure if my solution for skipping the CodeFixProvider when something is found inside an "include" is correct but I could not find a better solution. Once this merge request has been reviewed and merged I will start to work on implementing SA1629.

Fixes #148

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1628CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1628CodeFixProvider.cs
@@ -1,0 +1,107 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.DocumentationRules
+{
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Composition;
+    using System.Threading;
+    using System.Threading.Tasks;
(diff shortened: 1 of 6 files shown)


sharwell:
@dlemstra In case you're curious, the following link shows the code coverage for just the lines of code you wrote:
https://codecov.io/gh/DotNetAnalyzers/StyleCopAnalyzers/pull/2242/compare

I notice that `CheckElement` and `CheckIfFirstLetterUppercase` are the two methods missing coverage in areas. If you don't get to it earlier, I'll take a closer look when reviewing this later to propose some additional test cases that cover the edge cases for these methods. 😄

dlemstra:
@sharwell I can add some extra test cases tonight.

sharwell:
@dlemstra SA1628 is marked as "won't fix by design/intent", so it's unlikely we would merge an implementation of it.

dlemstra:
I personally would like to have this rule and SA1629 enabled  in my own projects. This should not be a rule that will be enabled by default though. I think there is no harm in having this in the library but keep it disabled by default. Would you or one of the other members consider re-opening/allowing SA1628?

PR #2243, opened 2016-12-06
Removed remaining traces of old SA0000 rule.

bjornhellander:
I found some remains of the previously deleted rule SA0000... Removed them.


StyleCop.Analyzers/StyleCop.Analyzers.ruleset
```diff
@@ -74,7 +74,6 @@
     <Rule Id="IDE0003" Action="None" />
   </Rules>
   <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
-    <Rule Id="SA0000" Action="Hidden" />
```
sharwell: :memo: It turns out this is actually still in use until #2244 gets merged.

PR #2245, opened 2016-12-06
SA1413 now also checks enum members.

### LINKED ISSUE #2239: Use trailing commas for Enums (similar to SA1413)
Similar to the new introduced maintainability rule SA1413 it would be nice to have such kind of rule for enum members as well (just to be consistent here).
It may also be less work for a reviewer as he can concentrate on the lines that changed but not the one that just got a trailing comma.

dlemstra:
This patch implements the feature that was requests in #2239. 

Fixes #2239

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1413CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1413CodeFixProvider.cs
@@ -3,19 +3,14 @@
 
 namespace StyleCop.Analyzers.MaintainabilityRules
 {
-    using System;
-    using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Composition;
-    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
     using Helpers;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CodeActions;
     using Microsoft.CodeAnalysis.CodeFixes;
-    using Microsoft.CodeAnalysis.CSharp;
-    using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Text;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1413UseTrailingCommasInMultiLineInitializers.cs
```diff
@@ -73,6 +74,23 @@ public override void Initialize(AnalysisContext context)
...
+            if (lastMember == null)
+            {
+                return;
+            }
+
+            var commas = initializer.ChildNodesAndTokens().OfType<SyntaxNodeOrToken>().Where(token => token.IsKind(SyntaxKind.CommaToken));
+            if (initializer.Members.Count() != commas.Count())
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, lastMember.GetLocation()));
+            }
```
vweijsters: The code above should follow the same pattern as the methods below. Specifically you can check `EnumDeclarationSyntax.Members.Count` against `EnumDeclarationSyntax.Members.SeparatorCount`. That will work just as well and it will perform quite a bit better.

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1413UnitTests.cs
```diff
@@ -425,6 +425,98 @@ void Foo()
...
+            DiagnosticResult[] expected =
+            {
+                this.CSharpDiagnostic().WithLocation(3, 5),
+            };
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpDiagnosticAsync(fixedTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+            await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
+        }
+
```
vweijsters: Please add test cases where the element without comma is directly followed by an end of line comment. Like:
```C#
    var testCode = @"enum TestEnum
{
    One = 2 /* test comment */
}
";
```

and verify that it is properly changed to 

```C#
    var testCode = @"enum TestEnum
{
    One = 2, /* test comment */
}
";
```

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1413CodeFixProvider.cs
```diff
@@ -115,5 +121,11 @@ private static SyntaxNode RewriteAnonymousObjectInitializer(AnonymousObjectCreat
             var fixedInitializer = initializer.WithInitializers(newInitializerExpressions);
             return fixedInitializer;
         }
+
+        private static TextChange GetEnumMemberTextChange(Diagnostic diagnostic, SyntaxNode syntaxRoot)
+        {
+            var member = (EnumMemberDeclarationSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan);
+            return new TextChange(diagnostic.Location.SourceSpan, member.ToString() + ",");
+        }
```
vweijsters: 💭  The code above should be the same pattern as the code for `RewriteInitializer` and `RewriteAnonymousObjectInitializer`.
If this code performs better than the existing code, then please also change the other methods to use the same pattern, for consistency.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1413CodeFixProvider.cs
```diff
@@ -115,5 +121,11 @@ private static SyntaxNode RewriteAnonymousObjectInitializer(AnonymousObjectCreat
             var fixedInitializer = initializer.WithInitializers(newInitializerExpressions);
             return fixedInitializer;
         }
+
+        private static TextChange GetEnumMemberTextChange(Diagnostic diagnostic, SyntaxNode syntaxRoot)
+        {
+            var member = (EnumMemberDeclarationSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan);
```
vweijsters: The `syntaxRoot.FindNode` call has already been done in the calling method, please reuse the result of that, as this is usually a quite expensive call.

dlemstra:
Thanks for the quick review. Will apply the changes this weekend.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1413UseTrailingCommasInMultiLineInitializers.cs
dlemstra: Thanks for pointing that out. I was looking for something like this 👍

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1413UnitTests.cs
dlemstra: I changed my unit test and it still passes

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1413CodeFixProvider.cs
dlemstra: I changed the code to only use my method and it still passes all the unit tests. Not sure how to test if it performs better but it does a lot less work: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/2245/commits/f983b5f165190e0083a78a866d3cd908196f6b2f#diff-9ac99be9ffc83f0bfb6aa1415e965bccR49.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/MaintainabilityRules/SA1413CodeFixProvider.cs
dlemstra: See my comment below.

PR #2256, opened 2016-12-21
Added an alternative way to enabling the config

### LINKED ISSUE #2226: Update EnableConfiguration.md to show setting AdditionalFiles through the VS2015 UI
See comment of @StefanKowalewski in #2072

aboo:
The build action can be changed via Properties window. 

Fixes #2226


sharwell:
@aboo I like making users aware of the simpler instructions. How do you feel about moving the new text *before* the existing set of instructions. Instead of "Alternately, ...", you can say "In Visual Studio 2015 Update 3 and newer, ...". Then follow it with something like "In older versions of Visual Studio 2015..." followed by the original instructions.

aboo:
@sharwell great suggestion. I will modify it and then create another pull request.

PR #2269, opened 2017-01-06
Fixed issue with properties that have an accessibility restriction.

dlemstra:
This patch fixes an issue where SA1513 would not report a diagnostic for this:
```C#
public int TestProtected
{
    get
    {
        return 1;
    }
    set
    {
    }
}
```
But would report one for a property that has an accessibility restriction:
```C#
public int TestProtected
{
    get
    {
        return 1;
    }
    protected set
    {
    }
}
```

This patch resolves that issue.


dlemstra:
I will be able to look at this in a couple of weeks. Feel free to add the test yourself if that is too late.

vweijsters:
@dlemstra I send you a pull request for the missing SA1516 unit tests. Could you please merge that into this PR?

dlemstra:
@sharwell The requested unit tests have been added by @vweijsters. I think this can be merged into master now.

PR #2301, opened 2017-02-23
Allow no whitespace on - and + signs that follow opening braces of interpolated strings

### LINKED ISSUE #2300: SA1021 conflicts with SA1012 in string interpolation
The following statement:
```csharp
var geo = DbGeography.FromText($"POINT ({-79.294088} {43.816447})");
```
shows SA1021, as `-79.294088` must be preceded by a space. When I put a space before the minus sign, I get SA1012, as `{` must not be followed by a space.

We need to detect the minus sign inside an interpolated string so that it does not require a preceding space.

ninety7:
Fixes #2300


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
```diff
@@ -57,6 +57,30 @@ internal static bool IsFirstInLine(this SyntaxToken token, bool allowNonWhitespa
...
         /// <summary>
+        /// Gets a value indicating whether the <paramref name="token"/> is an interpolated
+        /// unary expression.
+        /// </summary>
+        /// <param name="token">The token to process.</param>
+        /// <returns>
+        /// <see langword="true"/> if <paramref name="token"/> is an interpolated unary expression;
+        /// otherwise, <see langword="false"/>.
+        /// </returns>
+        internal static bool IsInterpolatedUnaryExpression(this SyntaxToken token)
```
sharwell: :bulb: It would be better to restrict this to `PrefixUnaryExpressionSyntax` instead of all tokens.

sharwell: ➡️ I tried making this change and wasn't thrilled with the result.

sharwell:
Thank you @ninety7! 😄

PR #2313, opened 2017-03-14
- Updated NuGet packages

vweijsters:
- Changed NewtonSoft.Json references to 'netstandard1.0'

Fixes #2290

Issue #2290, opened 2017-02-09
TypeLoadException with .NET Core RC4 and malformed stylecop.json


fubar-coder:
I added StyleCop.Analyzers to my project and managed to manually add a ruleset file to the csproj.

# Creating the project
```
md test
cd test
dotnet new console
```

# The ruleset file
```xml
<?xml version="1.0" encoding="utf-8"?>
<RuleSet Name="WebDAV-Server Ruleset" Description=" Rules for the WebDAV-Server project" ToolsVersion="15.0">
  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
    <Rule Id="SA0001" Action="None" />
    <Rule Id="SA1101" Action="None" />
    <Rule Id="SA1300" Action="None" />
    <Rule Id="SA1309" Action="None" />
    <Rule Id="SA1400" Action="None" />
    <Rule Id="SA1503" Action="None" />
    <Rule Id="SA1600" Action="None" />
    <Rule Id="SA1633" Action="None" />
  </Rules>
</RuleSet>
```

# The project file
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp1.0</TargetFramework>
    <CodeAnalysisRuleSet>test.ruleset</CodeAnalysisRuleSet>
  </PropertyGroup>

  <ItemGroup>
    <AdditionalFiles Include="stylecop.json" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta001">
      <PrivateAssets>All</PrivateAssets>
    </PackageReference>
  </ItemGroup>

</Project>
```

# The stylecop.json file

This file is malformed, because comments aren't allowed in the project.json file.

```json
{
  // ACTION REQUIRED: This file was automatically added to your project, but it
  // will not take effect until additional steps are taken to enable it. See the
  // following page for additional information:
  //
  // https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/EnableConfiguration.md

  "$schema": "https://raw.githubusercontent.com/DotNetAnalyzers/StyleCopAnalyzers/master/StyleCop.Analyzers/StyleCop.Analyzers/Settings/stylecop.schema.json",
  "settings": {
    "documentationRules": {
      "documentInternalElements": false
[...]

PR #2316, opened 2017-03-16
Updated stabilization with Newtonsoft.Json 9.0.1 (target netstandard1.0)

vweijsters:
Fixes TypeLoadExceptions (see #2290) with Newtonsoft.Json in release 1.0.0. Similar to #2313 
This should be sufficient for a 1.0.1 release.


mikes-gh:
This is needed for netcore sdk 1.0.0 compatibility (vs 2017)

sharwell:
@vweijsters Thanks for getting this updated. I need to run through some scenario testing to make sure we don't break anyone (specifically we don't want to break users who haven't updated to 2017 and thus don't need this change).

mikes-gh:
@vweijsters  Just to be clear this not so much a vs 2017 thing in fact it works when building in 2017.
Its to do with the net core sdk 1.0.0 including netcore runtime 1.1.1 that was released at the same time.  Admittedly they were released at the same time but this can be reproduced without vs 2017 using

dotnet build

So for example build servers osx linux which dont have vs 2017 will be affected.

sharwell:
I figured out the problem. In **StyleCop.Analyzers.nuspec** we included the wrong implementation of **Newtonsoft.Json.dll**. Instead of including the one in **net45**, we should be including the one we actually reference in the build, which is located in **portable-net40+sl5+wp80+win8+wpa81**.

PR #2319, opened 2017-03-19
#2308 - SA1008 triggered by typle return types on a method - fixed

### LINKED ISSUE #2308: SA1008 triggered by typle return types on a method (c# version 7)
The following method triggers SA1008

```
public (string primary, string alternate) CalculateMetaphone(string word)
      ~
{
}
```

the only way to avoid the rule is thus:

```
public(string primary, string alternate) CalculateMetaphone(string word)
```

which is clearly incorrect.
if the return type is a tuple, then a space should be inserted.

Nozziel:
#2308 - SA1008 triggered by typle return types on a method (c# version 7) - Fixed

Note: unit tests created but wont work without System.ValueTuple nuget package so set to Skip atm..

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1008UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1008UnitTests.cs
@@ -2032,6 +2032,90 @@ class ClassName
             await this.VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
         }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


vweijsters:
Labeling this `do not merge` until it is clear how / when we are going to use CodeAnalysis 2.0

sharwell:
Thank you for the submission @Nozziel. After a large amount of struggling over how to handle C# 7 in upcoming releases of these analyzers, I finally settled on a pattern which allows us to continue fully supporting C# 6 users. During this work (specifically my next push to pull request #2389), I updated the spacing rules to work with C# 7 and resolved #2308 as part of that. There is still a fair amount of work to do before C# 7 support is fully complete, but I really hope to finish with all the updates soon!

PR #2322, opened 2017-03-21
Remove weird character from start of file

adamralph:


sharwell:
@adamralph The character is a UTF-8 byte order mark. The GitHub change which caused rendering to break has now been corrected (I worked with GitHub support over the weekend to point out the problem and recommend a solution, which has now been deployed).

adamralph:
OK @sharwell np. I know about BOM's but I guess it was something else, since I've never seen them rendered in GitHub before.

adamralph:
BTW - the editor is still rendering them, and allowing them to be deleted:

image

sharwell:
@adamralph Thanks, I sent GitHub an update to the issue with a link to your comment and screenshot.

PR #2327, opened 2017-03-23
Switch to the new project system and CodeAnalysis 2.0

pdelvo:
This switches this project to the new project system. We also reference code analysis 2.0 which allows us to add new C# 7 related features. this was the main motivation behind this. I also included @Nozziel's pull request #2319 in here to have some new C# 7 features to try this with.

The two main analyzer libraries target netstandard1.3 which is the lowest standard that the code analysis libraries work with. The Test library works with netcoreapp1.1 and the tester app with net46.

Todo:
- [x] Use new project system in main libraries
- [x] Switch to netstandard 1.3
- [x] Update to CodeAnalysis 2.0
- [x] Update xunit to an up to date version
- [x] Fix all failing tests
- [x] Fix package creation
- [x] Fix VSIX creation
- [x] Fix merge conflicts
- [x] Fix Code Coverage
- [x] Make sure AppVeyor works
- [x] Clean up test references
- [ ] (Fix all scripts in ./scripts)
- [ ] (Try to make the command line build work with the vsix project by referencing Microsoft.VSSDK.BuildTools)
- [ ] (Try the Churn Testing)
- [ ] Consume our analyzers. This looks like a bug in the project system. It crashes the build when I try to consume the Stylecop.Analyzers package probably because there is a project in the solution with the same name.

@DotNetAnalyzers/stylecopanalyzers Can you take a look at this?


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Directory.Build.targets
```diff
@@ -0,0 +1,21 @@
+<Project>
+
+  <Target Name="AfterBuild">
```
sharwell: I'm not a fan of this approach to creating NuGet packages. The approach used before does a much better job of integrating with the build. Among other things, it does the following:

1. If you change assembly versions and then build again, it will remove old packages from the output directory
1. The **Clean** operation will remove the NuGet package created by the current build, if it exists in the output directory
1. If you delete just the NuGet package from the output directory, it knows to recreate it as an artifact of the build

mikes-gh:
@pdelvo
I downloaded your `NewProjectSystem` branch as I was hoping for netcore sdk 1.0.0/netcore1.1 support.
I built your branch and installed the resulting nuget package in my project.

Building from visual studio 2017 
I get no warnings

Building from the command line `dotnet build`
I got a load of these warnings

```
CSC : warning AD0001: Analyzer 'StyleCop.Analyzers.DocumentationRules.SA1642Cons
tructorSummaryDocumentationMustBeginWithStandardText' threw an exception of type
 'System.TypeLoadException' with message 'Could not load type 'System.Runtime.Se
rialization.Formatters.FormatterAssemblyStyle' from assembly 'mscorlib, Version=
4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.'. [C:\Users\mikes\Re
pos\Reports\src\Reports\Reports.csproj]
```

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Directory.Build.targets
pdelvo: What do you suggest should I do instead? I could make it its own target.

sharwell:
Superseded by #2370, in order to continue supporting Visual Studio 2015 (Roslyn 1.3 / C# 6) users.

PR #2337, opened 2017-04-10
Fix SA1308CodeFixProvider creating empty identifier and add regression test

jamesqo:
`SA1308CodeFixProvider` has a branch to prevent it from creating an empty identifier if the variable name consists solely of the offending prefix. However, if it consists only of multiple prefixes, e.g. `m_s_t_`, then it will try to rename the variable to the empty string.

This PR changes the logic to be more flexible, and adds regression tests. It also converts some existing tests for that analyzer to use `Theory` instead of `Fact`.


jamesqo:
I pulled from the wrong branch, unfortunately. I need to open a new PR.

PR #2338, opened 2017-04-10
Fix SA1308CodeFixProvider creating empty identifier and add regression test

jamesqo:
Had to close #2337 because I submitted from the wrong branch. Sorry about that.

`SA1308CodeFixProvider` has a branch to prevent it from creating an empty identifier if the variable name consists solely of the offending prefix. However, if it consists only of multiple prefixes, e.g. `m_s_t_`, then it will try to rename the variable to the empty string.

This PR changes the logic to be more flexible, and adds regression tests. It also converts some existing tests for that analyzer to use `Theory` instead of `Fact`.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/SA1308CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/SA1308CodeFixProvider.cs
@@ -42,18 +42,12 @@ namespace StyleCop.Analyzers.NamingRules
             {
                 var token = root.FindToken(diagnostic.Location.SourceSpan.Start);
 
-                // The variable name is the full suffix. In this case we cannot generate a valid variable name and thus will not offer a code fix.
-                if (token.ValueText.Length <= 2)
-                {
-                    continue;
-                }
-
-                var numberOfCharsToRemove = 2;
+                var numberOfCharsToRemove = 0;
 
                 // If a variable contains multiple prefixes that would result in this diagnostic,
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/NamingRules/SA1308CodeFixProvider.cs
```diff
@@ -66,16 +60,24 @@ public override FixAllProvider GetFixAllProvider()
                     break;
                 }
 
-                if (!string.IsNullOrEmpty(token.ValueText))
+                if (numberOfCharsToRemove == 0)
+                {
+                    continue;
```
sharwell: ❓ I noticed from the code coverage report that this `continue` statement isn't hit by the current tests. Is it possible to reach this `if` statement and still have `numberOfCharsToRemove` be 0?

jamesqo: @sharwell I think I added this to maintain previous behavior if the name was 1 char, or 2 chars and not one of the prefixes; without this check, we would no longer hit the `continue` below where we had before. However, if either of those were the case, then clearly the analyzer wouldn't have reported this in the first place. I'll remove this `if` block.

sharwell: Awesome. BTW, I asked because I was using this pull request to demonstrate to someone how code coverage tools can be used to ask more specific questions during code review.

PR #2351, opened 2017-04-30
Stabilization - new JSON parser

vweijsters:
After fruitlessly experimenting with different binaries of Newtonsoft.Json to make it work in .NET core, I went the other way and integrated a different JSON parser ([LightJson](https://github.com/MarcosLopezC/LightJson)) into our code as source files, to get rid of the nasty dependencies. I reformatted the source files to meet the StyleCop Analyzers rules.

This parser is a bit more strict, but it works in .NET core builds.

Also updated the xunit visual studio runner, as the old one doesn't work for me anymore since installing VS2017.


sharwell:
Interesting approach for sure. I should be able to do some experimenting in the pretty near future on this.

dlemstra:
I like this approach and looking at what has changed I don't think this a bad direction.

mikes-gh:
Does this unblock 

`dotnet build`

at the command line.  i.e. Msbuild xplat with no errors

dlemstra:
@mikes-gh Yup this would resolve that issue but @sharwell is looking for an alternative solution: https://gitter.im/DotNetAnalyzers/StyleCopAnalyzers?at=590e173b4746d29d2499ea72

sharwell:
@DotNetAnalyzers/members Can I get a second review of this?

mikes-gh:
@sharwell 
It works :+1: 
Thank you

JSkimming:
@vweijsters @sharwell Nice work 👍. Builds in travis now work when StyleCop is enabled, see below:

https://travis-ci.org/JSkimming/abioc/builds/238371077

PR #2364, opened 2017-05-28
Updated SA1513 to fire inside lambdas

### LINKED ISSUE #2062: SA1513 (ClosingBraceMustBeFollowedByBlankLine) not fired inside lambda
SA1513 (ClosingBraceMustBeFollowedByBlankLine) not fired inside lambda. Tested under StyleCopAnalyzers 1.0.0.

**Example**

SA1513 should be found on the closing parenthesis line before the second 'if'

``` csharp
Task.Run(() =>
{
    if (true)
    {
        Console.WriteLine();
    }
    if (true)
    {
        Console.WriteLine();
    }
});
```

bjornhellander:
Fixes #2062.


bjornhellander:
The problem was the use of private method IsPartOf to ignore certain cases, like the closing brace of a lambda in an argument list. This accidently ignored everything inside the lambda body as well.

I added #2365 to review remaining uses of IsPartOf.

PR #2366, opened 2017-05-28
Codecov-exe

larzw:
Changed from using python uploader to codecov-exe uploader.


larzw:
There is a new [codecov-exe](https://github.com/codecov/codecov-exe) uploader (C# code base) for Codecov. It does not have any dependencies (such as python).

sharwell:
Rather than use `choco install`, let's add the package to our repository [packages.config](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/.nuget/packages.config) so it gets cached by AppVeyor.

larzw:
Will do!

appveyor.yml
```diff
@@ -10,11 +10,8 @@ build:
   project: StyleCopAnalyzers.sln
   verbosity: minimal
 test_script:
-- .\packages\OpenCover.4.6.247-rc\tools\OpenCover.Console.exe -register:user -target:"%xunit20%\xunit.console.x86.exe" -targetargs:"C:\projects\stylecopanalyzers\StyleCop.Analyzers\StyleCop.Analyzers.Test\bin\Debug\StyleCop.Analyzers.Test.dll -noshadow -appveyor" -returntargetcode -filter:"+[StyleCop*]*" -excludebyattribute:*.ExcludeFromCodeCoverage* -excludebyfile:*\*Designer.cs -hideskipped:All -output:.\StyleCopAnalyzers_coverage.xml
```
sharwell: ❗️ Removing this line would disable our main unit tests.

larzw:
My bad, I accidentally deleted that line when I merged in the latest from master. I added that back in and switched to using nuget instead of choco. Thanks for you help @sharwell

Issue #2360, opened 2017-05-25
SA1303 suggest to rename "_1var" to "1var"


sm-g:
"1var" is invalid name for compiler

bjornhellander:
@sm-g, could you check again that you actually get SA1303 for this? I get SA1309 instead. Just so we don't have more than one problem here... If you do get SA1303, which version are you using?

bjornhellander:
I can fix the problem with the code fix for now, until we have sorted out which diagnostic is actually triggered.

sm-g:
Now I got SA1312 in same place.

bjornhellander:
Me saying "I get SA1309 instead" is really stupid, since you didn't tell us what your code looks like :-)
Could you provide more information about your code and in what situations you see SA1303 and SA1312?

PR #2367, opened 2017-05-28
Updated SA1309CodeFixProvider to not provide a code fix if the resulting identifier would be illegal

bjornhellander:
Fixes incorrect code fix from #2360.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1309UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1309UnitTests.cs
@@ -149,6 +149,25 @@ namespace StyleCop.Analyzers.Test.NamingRules
             await this.VerifyCSharpFixAsync(testCode, fixedCode).ConfigureAwait(false);
         }
 
+        [Fact]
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1309UnitTests.cs
```diff
@@ -132,6 +132,23 @@ public class Foo
             await this.VerifyCSharpFixAsync(testCode, fixedCode).ConfigureAwait(false);
         }
 
+        // This a regression test for issue #2360.
+        [Fact]
+        public async Task VerifyThatAFieldStartingWithAnUnderscoreAndADigitIsNotAffectedByCodeFixAsync()
```
sharwell: ❓ Can you add a test verifying that a field `_int` is correctly refactored to `@int`? I want to make sure that `IsValidIdentifier` isn't preventing us from using an escaped identifier.

bjornhellander: Good point. Will do.

PR #2370, opened 2017-05-29
Support C# 7 via light-up

sharwell:


sharwell:
@vweijsters @pdelvo This could benefit from additional tests, but I believe I have it in a state where we can start working in parallel using it to fix some of the outstanding C# 7 problems.

roterdam:
Is there a released version of this change?

sharwell:
@roterdam This PR allows us to support C# 7, but doesn't actually update any of the rules to use it. We're still working on updating the rules themselves but I'm hoping to release the first beta that includes this work soon.

roterdam:
thank you, eagerly waiting for this as I don't want to remove StyleCop just to use C# 7 features.

PR #2426, opened 2017-06-17
Add a test for SA0002 when deserialization throws JsonParseException

sharwell:
Previously we were testing for cases where `InvalidSettingsException` was thrown, but not for the case where the JSON syntax was invalid, or when some other unrelated exception was thrown. This commit adds tests covering additional cases.


sharwell:
❗️ This needs an additional test for the case where the exception filter evaluates to false.

PR #2442, opened 2017-06-19
Update embedded LightJson to be culture agnostic

### LINKED ISSUE #2440: TestAsNumber fails locally
TestAsNumber (part of JsonValueTests for LightJson) from current master fails on my machine.
Probably a culture issue. According to this Stack Overflow question, JSON is supposed to be culture agnostic:
https://stackoverflow.com/questions/4721143/how-to-localize-when-json-serializing
I'm investigating...

bjornhellander:
Fixes #2440.

Started by adding tests which are supposed to fail with the current behaviour.


bjornhellander:
Updated code to use InvariantCulture. I picked a reasonable number style, but I'm not sure it's exactly as in the simple version of TryParse that was previously used. On the other hand, we will probably not need any decimal number properties, so it shouldn't matter much...

PR #2460, opened 2017-06-22
Enable warnings as errors on AppVeyor

sharwell:
* Reduce severity in Visual Studio from error to warning (improves iteration times during development)
* Treat warnings as errors on command line and AppVeyor builds to ensure items are corrected before making their way into the main development branch


oatkins:
👍  Seems very logical to me. Notionally, a "style" violation is probably a better fit for a warning than an error. And turning on `TreatWarningsAsErrors` is good practise.

PR #2461, opened 2017-06-24
Fix SA1515 handling of comment after pattern-matching case label

### LINKED ISSUE #2462: SA1515 (Single-line comment must be preceded by blank line) should not be reported for single line comment directly following a C# 7 case
When using C# 7 `case` statements (with type testing included), `SA1515` should not flag single-line comments.

The following snippet produces such a warning in StyleCop.Analyzers 1.0.2:

```csharp
static void Main(object foo)
{
    switch (foo)
    {
        case string s:
            // hi there
            return;
        default:
            break;
    }
}
```

sharwell:
Fixes #2462.


AArnott:
Fixes #2462

PR #2483, opened 2017-06-27
Fixes Conflict between SA1008 and SA1009 for single-line if

### LINKED ISSUE #2475: Conflict between SA1008 and SA1009 for single-line if
**Input:**

```csharp
if (true) (true ? 1 : 0).ToString();
```

**Expected result:**

Neither SA1008 nor SA1009 is reported.

**Actual result:**

SA1009 is reported for the closing parenthesis of the `if` condition. Applying the code fix produces the following:

```csharp
if (true)(true ? 1 : 0).ToString();
```

After applying the code fix, SA1008 is reported on the opening parenthesis of the parenthesized ternary expression. Applying the code fix for that restores the original code.

vweijsters:
Fixes #2475

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/MetadataReferences.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/MetadataReferences.cs
@@ -5,7 +5,6 @@ namespace StyleCop.Analyzers.Test.Helpers
 {
     using System;
     using System.Collections.Immutable;
-    using System.IO;
     using System.Linq;
     using System.Reflection;
     using Microsoft.CodeAnalysis;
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -87,6 +87,13 @@ private static void HandleCloseParenToken(SyntaxTreeAnalysisContext context, Syn
             switch (nextToken.Kind())
             {
             case SyntaxKind.OpenParenToken:
+                // Allow a space between an open and a close paren when:
+                // - they are part of an if statement
+                // - they are on the same line
+                // - the open paren is part of a parenthesized expression
```
sharwell: :bulb: Can also happen for a tuple expression. It would be good to go ahead and cover this case.

```csharp
if (true) (1, 2).ToString();
```

vweijsters:
I've added test cases for tuple expressions as well.

I also ran into an interesting issue with 70+ of the C# 7 tests failing with CS8179. A bit of searching turned up this [information](https://support.microsoft.com/en-us/help/4015416/error-cs8179-when-you-debug-an-application-on-a-computer-that-has-net). I've made changes to `MetadataReferences.cs` to make sure that we are always using .NET 4.6 as target framework (inspired by this [comment](https://stackoverflow.com/a/34268178))

StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/MetadataReferences.cs
```diff
@@ -15,9 +16,9 @@ namespace StyleCop.Analyzers.Test.Helpers
     /// </summary>
     internal static class MetadataReferences
     {
-        internal static readonly MetadataReference CorlibReference = MetadataReference.CreateFromFile(typeof(object).Assembly.Location).WithAliases(ImmutableArray.Create("global", "corlib"));
-        internal static readonly MetadataReference SystemReference = MetadataReference.CreateFromFile(typeof(System.Diagnostics.Debug).Assembly.Location).WithAliases(ImmutableArray.Create("global", "system"));
-        internal static readonly MetadataReference SystemCoreReference = MetadataReference.CreateFromFile(typeof(Enumerable).Assembly.Location);
+        internal static readonly MetadataReference CorlibReference = CreateDotNetFrameworkMetadataReference("mscorlib").WithAliases(ImmutableArray.Create("global", "corlib"));
```
sharwell: :memo: This situation was fixed in the first commit of #2478.

vweijsters: Yup, but that one isn't merged yet (and I forgot about it, even though I reviewed it 😉).
I do like the solution here better, as it is more robust towards any future changes.

sharwell: Yes, unless someone doesn't have the 4.6 reference assemblies in that location.

sharwell: I think I'd prefer this just go with the simpler approach for now unless we're seeing problems with it later (i.e. no changes to this file, but everything else good).

vweijsters:
Rebased and I removed the MetadataReferences changes.

PR #2501, opened 2017-07-23
Allow underscore digit in enum members for SA1300

### LINKED ISSUE #1935: RC2: SA1300 for enum value `_12key`
I have an enum that includes a value named `_12key`.

This raises SA1300: Element `_12key` must begin with an uppercase letter.

There are a couple of issues with this:
1. The code fix provider suggest to rename the value to `_12key1`, which does not solve the issue
2. Given that the first three characters are not alphabetic, I would expect the violation not to be flagged or to be flagged for the lowercase `k`. However, renaming to `_12Key` does not fix the issue, either

vweijsters:
Fixes #1935

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1300UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1300UnitTests.cs
@@ -971,6 +971,39 @@ public interface IInterface
             await this.VerifyCSharpFixAsync(testCode, fixedCode).ConfigureAwait(false);
         }
 
+        [Fact]
+        [WorkItem(1935, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1935")]
+        public async Task TestUnderscoreExclusionAsync()
+        {
+            var testCode = @"public enum TestEnum
+{
+    _12clock,
+    _12Clock,
+    _tick,
+    _Tock,
+}
+";
+
+            var fixedCode = @"public enum TestEnum
+{
+    _12clock,
+    _12Clock,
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1300UnitTests.cs
```diff
@@ -971,6 +971,39 @@ public interface IInterface
             await this.VerifyCSharpFixAsync(testCode, fixedCode).ConfigureAwait(false);
         }
 
+        [Fact]
+        [WorkItem(1935, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1935")]
+        public async Task TestUnderscoreExclusionAsync()
+        {
+            var testCode = @"public enum TestEnum
```
DavidParks8: This can be made into a const local variable, which would be more memory efficient. A similar fix can be applied to other variables in this file.

sharwell: ➡️ This is just test code

vweijsters:
This was mainly added because the original StyleCop supported it. I personally have no preference for allowing digits or not. It would not be that hard to make it configurable, but what should be the default?

sharwell:
> This was mainly added because the original StyleCop supported it

I forgot about this part of the original discussion. It's been our default to follow StyleCop Classic except for cases where it's problematic or clearly a bug, and this issue doesn't seem to be either of the latter. A configuration option *might* make sense, but this really seems to be a corner case so I doubt people will be looking for ways to enforce it.

PR #2503, opened 2017-07-26
CPS cleanup

### LINKED ISSUE #2506: Adding resources will not update designer files in VS2017 solution
When adding a resource through the resource editor in VS2017, the designer files will not be updated automatically.

sharwell:
Fixes #2506


sharwell:
:memo: Early build failures were due to dotnet/corefx#20609, which is fixed for the upcoming release of Visual Studio 2017 but affects the current AppVeyor image.

PR #2507, opened 2017-07-31
Implemented SA1629 (+ codefix)

### LINKED ISSUE #149: SA1629: DocumentationTextMustEndWithAPeriod

vweijsters:
Fixes #149

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1629CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1629CodeFixProvider.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.DocumentationRules
+{
+    using System.Collections.Immutable;
+    using System.Composition;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.CodeFixes;
+    using Microsoft.CodeAnalysis.Text;
+    using StyleCop.Analyzers.Helpers;
+
+    /// <summary>
+    /// Implements a code fix for <see cref="SA1629DocumentationTextMustEndWithAPeriod"/>.
+    /// </summary>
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
```diff
@@ -32,28 +38,95 @@ namespace StyleCop.Analyzers.DocumentationRules
...
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(Descriptor, diagnosticLocations[0], NoCodeFixProperties));
+
+                            // only report a single instance of the diagnostic, as they will all be reported on the same location anyway.
+                            break;
+                        }
+                    }
+                }
+            }
+            catch (InvalidCastException)
```
sharwell: ❓ What's getting thrown?

vweijsters: Nothing, seems that something was left over from debugging. I'll remove it.

In more detail, I made a mistake with the included documentation, resulting in a generated xml comment from roslyn stating that there was something wrong with the included documentation. This results in an InvalidCastException on line 112.

Typing this has made me realize that if the end user makes a mistake in the included documentation then the same scenario can happen 😃 I'll adapt line 112 to use `OfType`, instead of `Cast`.

vweijsters:
Updated with better handling of invalid included documentation.

sharwell:
I should be able to go through this in detail next week. 😎

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1629CodeFixProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+                }
+            }
+
+            return SpecializedTasks.CompletedTask;
+        }
+
+        private static async Task<Document> GetTransformedDocumentAsync(Document document, Diagnostic diagnostic, CancellationToken cancellationToken)
+        {
+            var text = await document.GetTextAsync(cancellationToken).ConfigureAwait(false);
+            var newText = text.WithChanges(new TextChange(new TextSpan(diagnostic.Location.SourceSpan.Start, 0), "."));
```
DavidParks8: This would be a bit easier to read and debug if `new TextChange` and `new TextSpan` were stored in variables on their own lines. Doing that would also help minimize future merge conflicts.

PR #2510, opened 2017-08-02
Fix and test handling of missing tokens in SA1500 (BracesForMultiLineStatementsMustNotShareLine)

sharwell:


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500BracesForMultiLineStatementsMustNotShareLine.cs
```diff
@@ -186,7 +186,7 @@ private static void CheckBraceToken(SyntaxNodeAnalysisContext context, SyntaxTok
 
             int line = token.GetLineSpan().StartLinePosition.Line;
 
-            SyntaxToken previousToken = token.GetPreviousToken();
+            SyntaxToken previousToken = token.GetPreviousToken(includeZeroWidth: true);
```
sharwell: :memo: `GetPreviousToken` and `GetNextToken` can never return a token where `IsMissing` is true unless `includeZeroWidth: true` is explicitly passed.

PR #2535, opened 2017-09-27
Implemented missing parameter validation in separated syntax list acc…

### LINKED ISSUE #2388: Update separated syntax list accessor helper to validate arguments
Currently two tests in `LightupHelpersTests` (added in #2370) are disabled because argument validation is not performed for the second argument to `CreateSeparatedSyntaxListPropertyAccessor` and `CreateSeparatedSyntaxListWithPropertyAccessor`. This validation should be implemented and the tests enabled.

vweijsters:
…essor helper

Fixes #2388

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Lightup/LightupHelpersTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Lightup/LightupHelpersTests.cs
@@ -67,7 +67,7 @@ namespace StyleCop.Analyzers.Test.Lightup
             Assert.ThrowsAny<InvalidOperationException>(() => LightupHelpers.CreateSeparatedSyntaxListPropertyAccessor<ParameterListSyntax, ParameterSyntax>(typeof(BaseParameterListSyntax), nameof(BaseParameterListSyntax.Parameters)));
         }
 
-        [Fact(Skip = "Validation is not currently performed for the second argument.")]
+        [Fact]
         public void TestCreateSeparatedSyntaxListPropertyAccessorValidateElementType()
         {
             // The call *should* have been made with the second generic argument set to `ParameterSyntax`
@@ -98,7 +98,7 @@ namespace StyleCop.Analyzers.Test.Lightup
             Assert.ThrowsAny<InvalidOperationException>(() => LightupHelpers.CreateSeparatedSyntaxListWithPropertyAccessor<ParameterListSyntax, ParameterSyntax>(typeof(BaseParameterListSyntax), nameof(BaseParameterListSyntax.Parameters)));
         }
 
-        [Fact(Skip = "Validation is not currently performed for the second argument.")]
+        [Fact]
         public void TestCreateSeparatedSyntaxListWithPropertyAccessorValidateElementType()
         {
             // The call *should* have been made with the second generic argument set to `ParameterSyntax`
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Lightup/LightupHelpers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Lightup/LightupHelpers.cs
@@ -13,8 +13,6 @@ namespace StyleCop.Analyzers.Lightup
 
     internal static class LightupHelpers
     {
-        private const string WrappedTypePropertyName = "WrappedType";
-
         private static readonly ConcurrentDictionary<Type, ConcurrentDictionary<SyntaxKind, bool>> SupportedWrappers
             = new ConcurrentDictionary<Type, ConcurrentDictionary<SyntaxKind, bool>>();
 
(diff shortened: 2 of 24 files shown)


sharwell:
❓ Is it not possible to implement this without the `WrappedType` property?

vweijsters:
I see three options:
1. Do not check the second parameter when a wrapper type is detected
2. Assume that the class name of the wrapper type matches the original type + Wrapper. Then check the name.
3. Create an static immutable dictionary inside a helper class that maps wrappers to the corresponding wrapped types.

Not sure if option 1 is desirable.
Option 2 would be a no-go for me.
Option 3 could be a decent option, it would also be possible to remove wrapped type name and wrapped type members from the wrappers and have them resolve against the dictionary as well.

What would be your preference?

sharwell:
I suppose it's impossible to truly avoid ugliness given this is a light-up layer.

* Option 1 puts us back where we started (easy to make a mistake)
* Option 2 is totally unintuitive
* Option 3 seems reasonable, and should be testable (with a test that verifies that all implementations of `ISyntaxWrapper<>` in the assembly appear in the map)

The advantage of option 3 is it avoids adding items to the light-up API which are only used for internal validation. I'd be curious to see how it compares to the approach here.

vweijsters:
@sharwell I've updated it with an "option 3" solution. Can you have a look again?

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/LightupHelpers.cs
```diff
@@ -134,6 +136,11 @@ internal static bool CanWrapNode(SyntaxNode node, Type underlyingType)
 
             var propertySyntaxType = property.PropertyType.GenericTypeArguments[0];
 
+            if (!ValidatePropertyType(typeof(TProperty), propertySyntaxType))
+            {
+                throw new InvalidOperationException();
```
sharwell: :memo: It doesn't appear this exception case is covered by any tests (looking at codecov)

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/LightupHelpers.cs
```diff
@@ -258,6 +265,11 @@ internal static bool CanWrapNode(SyntaxNode node, Type underlyingType)
...
```
sharwell: :memo: It doesn't appear this exception case is covered by any tests (looking at codecov)

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/LightupHelpers.cs
```diff
@@ -280,5 +292,19 @@ internal static bool CanWrapNode(SyntaxNode node, Type underlyingType)
...
+        {
+            if (returnType.GetTypeInfo().ImplementedInterfaces.Any(t => t.GetGenericTypeDefinition() == typeof(ISyntaxWrapper<>)))
+            {
+                // The requested property is a wrapper type, so validate against the wrapped type.
+                var wrappedType = WrapperHelper.GetWrappedType(returnType);
+                return wrappedType == actualType;
+            }
+            else
+            {
+                return returnType == actualType;
```
sharwell: :memo: It doesn't appear this exception case is covered by any tests (looking at codecov)

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/LightupHelpers.cs
```diff
@@ -280,5 +292,19 @@ internal static bool CanWrapNode(SyntaxNode node, Type underlyingType)
                     valueParameter);
             return expression.Compile();
         }
+
+        private static bool ValidatePropertyType(Type returnType, Type actualType)
+        {
+            if (returnType.GetTypeInfo().ImplementedInterfaces.Any(t => t.GetGenericTypeDefinition() == typeof(ISyntaxWrapper<>)))
```
sharwell: :bulb: Since we now have `WrapperHelper`, this method can be simplified:

```csharp
var requiredType = WrapperHelper.GetWrappedType(returnType) ?? returnType;
return requiredType == actualType;
```

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/LightupHelpers.cs
```diff
@@ -13,6 +13,8 @@ namespace StyleCop.Analyzers.Lightup
 
     internal static class LightupHelpers
     {
+        private const string WrappedTypePropertyName = "WrappedType";
```
sharwell: :bulb: Constant value appears to be unused

StyleCop.Analyzers/StyleCop.Analyzers.Test/Lightup/LightupHelpersTests.cs
```diff
@@ -67,7 +67,7 @@ public void TestCreateSeparatedSyntaxListPropertyAccessor()
             Assert.ThrowsAny<InvalidOperationException>(() => LightupHelpers.CreateSeparatedSyntaxListPropertyAccessor<ParameterListSyntax, ParameterSyntax>(typeof(BaseParameterListSyntax), nameof(BaseParameterListSyntax.Parameters)));
         }
 
-        [Fact(Skip = "Validation is not currently performed for the second argument.")]
+        [Fact]
```
sharwell: :question: Do we know where the `InvalidOperationException` is getting thrown? Based on the codecov results, it's not from either of the new cases this pull request added.

vweijsters:
Apparently changing the 'ValidatePropertyType' implementation has fixed the coverage as well.

I'm currently assuming that the following line:
```C#
 if (returnType.GetTypeInfo().ImplementedInterfaces.Any(t => t.GetGenericTypeDefinition() == typeof(ISyntaxWrapper<>)))
```
is the culprit, as it will throw an InvalidOperationException when `t` is not a generic type most likely.

PR #2543, opened 2017-10-15
Update xunit version

scott-xu:


sharwell:
:memo: It looks like this upgrade was a bit more complicated than it normally is.

Superseded by #2553

PR #2553, opened 2017-11-05
Update xunit to 2.3.0

sharwell:
Supersedes #2543


vweijsters:
The build is failing because `Assert.Empty` uses an enumerator to check if the collection is empty, instead of checking the count. `SeparatedSyntaxListWrapper` does not implement the `GetEnumerator` call.

sharwell:
@vweijsters Nice, I'll look into it 😄

PR #2554, opened 2017-11-05
Add .editorconfig

### LINKED ISSUE #2551: include .editorconfig equivalents on the wiki
It would be nice if the documentation for each rule said what [the equivalent .editorconfig rule](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference) was (if applicable), so that teams could package settings to override Visual Studio automatically to match their StyleCop rules.

sharwell:
Related to #2551

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+root = true
+
+[*.cs]
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


vweijsters:
I don't think the intention of #2551 was to have an .editorconfig file for the solution. I think the goal was to have a listing of applicable editor config settings in the documentation for each rule.

sharwell:
@vweijsters I changed the wording to reference the issue but not close it.

PR #2561, opened 2017-11-17
Fix Turkish İ problem

monoblaine:
This PR makes sure tests don't fail on a platform with Turkish-specific globalization settings.

I also had to modify `UseCultureAttribute` class so that the `async` methods are executed with the given culture info.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/UseCultureAttribute.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/UseCultureAttribute.cs
@@ -33,6 +33,10 @@ namespace StyleCop.Analyzers.Test.Helpers
 
         private CultureInfo originalUiCulture;
 
+        private CultureInfo originalDefaultCulture;
+
+        private CultureInfo originalDefaultUiCulture;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="UseCultureAttribute"/>
         /// class with a culture.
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/UseCultureAttribute.cs
```diff
@@ -97,6 +100,9 @@ public override void Before(MethodInfo methodUnderTest)
         /// <param name="methodUnderTest">The method under test</param>
         public override void After(MethodInfo methodUnderTest)
         {
+            CultureInfo.DefaultThreadCurrentCulture = this.originalCulture;
+            CultureInfo.DefaultThreadCurrentUICulture = this.originalUiCulture;
+
```
vweijsters: The `CultureInfo.DefaultThreadCurrentCulture` and `CultureInfo.DefaultThreadCurrentUICulture` should be reset to their original values (which is most likely `null` and not the currently active culture.)

monoblaine:
Hey @vweijsters thanks for the review! You're right, [the default value of `CultureInfo.DefaultThreadCurrentCulture`](https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.defaultthreadcurrentculture(v=vs.110).aspx?f=255&mspperror=-2147217396#Anchor_1) is `null`. But instead of assigning `null` directly, I thought that saving the original value and restoring it later would be a better practice anyway.

PR #2562, opened 2017-11-18
Implemented SA1600 codefix

### LINKED ISSUE #764: Code fix: SA1600 (ElementsMustBeDocumented)

vweijsters:
Closes #764

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlSyntaxFactory.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlSyntaxFactory.cs
@@ -171,11 +171,6 @@ namespace StyleCop.Analyzers.Helpers
             return MultiLineElement(XmlCommentHelper.SummaryXmlTag, newLineText, content);
         }
 
-        public static XmlElementSyntax EmptySummaryElement(string newLineText)
-        {
-            return Element(XmlCommentHelper.SummaryXmlTag, SyntaxFactory.List(new XmlNodeSyntax[] { NewLine(newLineText) }));
-        }
-
         public static XmlElementSyntax RemarksElement(string newLineText, params XmlNodeSyntax[] content)
         {
             return RemarksElement(newLineText, List(content));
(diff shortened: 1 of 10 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlSyntaxFactory.cs
```diff
@@ -191,6 +196,16 @@ public static XmlElementSyntax ReturnsElement(string newLineText, SyntaxList<Xml
             return MultiLineElement(XmlCommentHelper.ReturnsXmlTag, newLineText, content);
         }
 
+        public static XmlElementSyntax CompactReturnsElement(params XmlNodeSyntax[] content)
```
sharwell: :bulb: The previous `ReturnsElement` method is only used by test code. You can replace the current behavior (`MultiLineElement`) with the behavior you added here, and the new methods can be removed.

StyleCop.Analyzers/StyleCop.Analyzers/StyleCop.Analyzers.csproj
```diff
@@ -37,4 +37,19 @@
     <None Include="**\*.json" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
   </ItemGroup>
 
+  <ItemGroup>
```
sharwell: :memo: The changes to this project file were triggered by a bug in the project system, and can be manually reverted after making any desired changes to the .resx file(s).

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlSyntaxFactory.cs
```diff
@@ -171,6 +171,11 @@ public static XmlElementSyntax SummaryElement(string newLineText, SyntaxList<Xml
             return MultiLineElement(XmlCommentHelper.SummaryXmlTag, newLineText, content);
         }
 
+        public static XmlElementSyntax EmptySummaryElement(string newLineText)
```
sharwell: 💭 The behavior of `SummaryElement(newLineText)` should be acceptable for an empty element. It will produce content with the following form:

```csharp
/// <summary>
/// 
/// </summary>
```

Conveniently this is the same form you get if you manually type `///` in the editor.

sharwell:
I went ahead and updated this pull request so I can get it included for the 1.1.0-beta007 release.

PR #2567, opened 2017-11-26
Fixes issue with bulk fixer of SA1500

### LINKED ISSUE #2566: SA1500 - No indentation of else and finally when applying fix to whole document
Hi,
When using the autofix for SA1500 "else" and "finally" statements aren't getting indented. Same with a catch without brackets after it.
The singular fix does correctly indent, but when applying to the whole document, project or solution they have no indents. 
Screenshot with example attached.
sa1500 not indenting

vweijsters:
Fixes #2566


vweijsters:
@sharwell can you retrigger the build on this one? The current build failed on an interdeministic SA1501 code fix (see #2010, #2067).

PR #2577, opened 2017-12-09
Fixed false positive for ambiguous method overloads in SA1410

### LINKED ISSUE #2572: SA1410 misfires for `new Thread(delegate ()`
On the following code, C# requires the `delegate()` parentheses, but SA1410 says I should remove them (which leads to a build break).

```csharp
            this.thread = new Thread(delegate()
            {
                // this is necessary to make sure a dispatcher exists for this thread
                Dispatcher unused = Dispatcher.CurrentDispatcher;
                unused.UnhandledException += new DispatcherUnhandledExceptionEventHandler(this.OnUnhandledException);

                // signal the other thread that it's safe to run
                resetEvent.Set();

                // start the dispatching pump
                Dispatcher.Run();
            });
```

vweijsters:
Fixes #2572

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1410UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1410UnitTests.cs
@@ -99,6 +99,109 @@ namespace StyleCop.Analyzers.Test.MaintainabilityRules
             await this.VerifyCSharpFixAsync(oldSource, newSource, cancellationToken: CancellationToken.None).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Verify that no diagnostic is produced when removing the parentheses from the delegate statement
+        /// would result in ambiguity on which method overload to call.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        [WorkItem(2572, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2572")]
+        public async Task TestMethodOverloadAmbiguityAsync()
+        {
+            var testCode = @"public class TestClass
+{
+    public delegate bool Delegate1();
+    public delegate bool Delegate2(int x);
+
+    public void TestMethod()
+    {
+        var thread = new System.Threading.Thread(delegate()
+        {
+            // ...
+        });
+
+        this.TestMethod2(delegate() { return true; });
+    }
+
+    public void TestMethod2(Delegate1 d)
+    {
+    }
+
+    public void TestMethod2(Delegate2 d)
+    {
+    }
+}
+";
+
+            await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Verify that the ambiguity detection is specific enough
+        /// </summary>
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1410RemoveDelegateParenthesisWhenPossible.cs
```diff
@@ -81,8 +83,59 @@ private static void HandleAnonymousMethodExpression(SyntaxNodeAnalysisContext co
...
+                }
+            }
+
             // Remove delegate parenthesis when possible
             context.ReportDiagnostic(Diagnostic.Create(Descriptor, syntax.ParameterList.GetLocation()));
         }
+
+        private static bool HasAmbiguousOverload(SyntaxNodeAnalysisContext context, int parameterIndex, SyntaxNode methodCallSyntax)
+        {
+            var methodSymbol = (IMethodSymbol)context.SemanticModel.GetSymbolInfo(methodCallSyntax, context.CancellationToken).Symbol;
```
sharwell: :memo: We've hit several cases recently where `GetSymbolInfo` caused a pronounced performance problem for an analyzer. We may be fairly safe for this analyzer since the call only occurs for cases involving the `delegate() { }` syntax.

StyleCop.Analyzers/StyleCop.Analyzers/MaintainabilityRules/SA1410RemoveDelegateParenthesisWhenPossible.cs
```diff
@@ -81,8 +83,59 @@ private static void HandleAnonymousMethodExpression(SyntaxNodeAnalysisContext co
...
+
+            var nameOverloads = methodSymbol.ContainingType.GetMembers(methodSymbol.Name);
```
sharwell: ❓ What about extension methods?

vweijsters: Those could be a problem as well, but I don't know how to find them efficiently. Do you have any ideas?

pdelvo: Maybe a good way to do this would be to use the speculative model and "try" to remove the parenthesis and check that it still finds the right symbol without errors.

sharwell:
@vweijsters I updated this to use the speculative semantic model, simplifying the code and revealing an edge case I wasn't previously aware of. Please review the update 👍

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1410UnitTests.cs
```diff
@@ -99,6 +99,134 @@ public void Bar()
...
+    public delegate bool Delegate1();
+    public delegate bool Delegate2(int x);
+
+    public void TestMethod()
+    {
+        this.TestMethod2(delegate() { return true; });
+        this.TestMethod3(0, delegate() { return true; });
+        this.TestMethod4(delegate() { return true; }, 1);
+        this.TestMethod5(delegate() { return true; });
+        this.TestMethod6(delegate() { return true; });
```
sharwell: :memo: I originally expected this case to be ambiguous, but it was not.

Issue #2578, opened 2017-12-09
Update tests to cover ref-like types


sharwell:
* `ref` modifier for types
* `readonly` modifier for types
* `in` modifier for parameters and arguments

See #2546 for a specific case where this was a problem.

sharwell:
Marking as **blocked** while I figure out the testing situation.

sharwell:
@pdelvo Can you expand on the tests in #2584 to include the following?

1. Cases that have both `readonly` and `ref`
2. Cases that have `partial`

PR #2584, opened 2017-12-14
SA1206 C# 7.2 test update

pdelvo:
I updated the test project to use the newest roslyn version and I updated the SA1206 tests appropriately. I had to remove one test I had previously because it tested code that is not valid C# (putting ref before the access modifier is illegal).

I also updated xunit to the current version.

This partially adresses #2578.


vweijsters:
@sharwell Can this be merged, or do we want to make separate test projects for C# 7.1, 7.2 etc?

PR #2589, opened 2017-12-29
SA1210 is inconsistent with IDE's Remove and Sort Usings

### LINKED ISSUE #2587: SA1210 is inconsistent with IDE's Remove and Sort Usings
SA1210 is inconsistent with IDE's Remove and Sort Usings when "Place 'System' directives first when sorting usings" settings is enabled for a UWP project. In that scenario, the IDE treats Windows.* namespaces as 'System' usings as well, where as SA1210 doesn't.
Sadly this setting exists and even more sadly the Windows.* idiosyncrasy exists, but SA1210 should respect that setting as well.

ali-hk:
Fixes #2587 
SA1208 and SA1210 are inconsistent with the IDE's Remove and Sort Usings when "Place 'System' directives first when sorting usings" settings is enabled for a UWP project. In that scenario, the IDE treats Windows.* namespaces as 'System' usings as well, whereas SA1208 and SA1210 don't.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.UsingsSorter.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.UsingsSorter.cs
@@ -438,7 +438,7 @@ namespace StyleCop.Analyzers.OrderingRules
                 var namespaceTypeName = namespaceSymbol.ToDisplayString(FullNamespaceDisplayFormat);
                 var firstPart = namespaceTypeName.Split('.')[0];
 
-                return string.Equals(SystemUsingDirectiveIdentifier, firstPart, StringComparison.Ordinal);
+                return string.Equals(SystemUsingDirectiveIdentifier, firstPart, StringComparison.Ordinal) || string.Equals(WindowsUsingDirectiveIdentifier, firstPart, StringComparison.Ordinal);
             }
 
             private void ProcessMembers(SyntaxList<MemberDeclarationSyntax> members)
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.cs
@@ -26,6 +26,7 @@ namespace StyleCop.Analyzers.OrderingRules
     internal sealed partial class UsingCodeFixProvider : CodeFixProvider
     {
         private const string SystemUsingDirectiveIdentifier = nameof(System);
+        private const string WindowsUsingDirectiveIdentifier = "Windows";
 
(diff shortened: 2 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.UsingsSorter.cs
```diff
@@ -435,10 +435,7 @@ private bool IsSeparatedSystemUsing(UsingDirectiveSyntax syntax)
                     return false;
                 }
 
-                var namespaceTypeName = namespaceSymbol.ToDisplayString(FullNamespaceDisplayFormat);
-                var firstPart = namespaceTypeName.Split('.')[0];
-
-                return string.Equals(SystemUsingDirectiveIdentifier, firstPart, StringComparison.Ordinal);
+                return syntax.IsSystemUsingDirective();
```
vweijsters: This change is incorrect. The original code was using the symbol info to get a fully expanded namespace in the using statement, something that cannot easily be done with only the syntax information.

ali-hk:
@vweijsters, yup, it seems the subset of tests that I'd run wasn't enough. Fixed and re-ran all tests.
I'll add tests in a subsequent commit.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/OrderingRules/UsingCodeFixProvider.UsingsSorter.cs
ali-hk: Correct, I misunderstood the original intent here. It might be worth adding a clarifying comment here.

ali-hk: Correct, I misunderstood the original intent here. It might be worth adding a clarifying comment here.

ali-hk:
@vweijsters, actually, adding tests for this scenario might be a bit difficult, because it'd need to resolve the Windows namespace, which would require adding a Windows.winmd reference to the compilation. That would introduce an unnecessary dependency on the UWP SDK for running the tests.

In reality the behavior isn't new and doesn't function any differently from the existing System namespace code.

What would your suggestion be to move forward?

vweijsters:
@ali-hk It is possible to fake the namespace (and type) definitions by defining them in a separate source text.

You can then use

```C#
protected Task VerifyCSharpDiagnosticAsync(string[] sources, DiagnosticResult[] expected, CancellationToken cancellationToken, string[] filenames = null)
```
and
```C#
protected async Task VerifyCSharpFixAsync(string[] oldSources, string[] newSources, string[] batchNewSources = null, string[] oldFileNames = null, string[] newFileNames = null, int? codeFixIndex = null, bool allowNewCompilerDiagnostics = false, int numberOfIncrementalIterations = DefaultNumberOfIncrementalIterations, int numberOfFixAllIterations = 1, CancellationToken cancellationToken = default(CancellationToken))
```

to verify the behavior.

ali-hk:
@vweijsters, I believe I've addressed your concerns, let me know if there's any other changes I should make.

PR #2592, opened 2018-01-10
Implement SA1135 (UsingDirectivesMustBeQualified)

### LINKED ISSUE #1761: Implement SA1135: Using directives must be qualified

vweijsters:
As #1762 is based of an unknown (deleted) branch, I've created a new pull request to revive it.

I've rebased the original pull request and did some (minor) fixes.

Fixes #1761


sharwell:
Can you add tests for the following cases?

#### References to using alias directives

I'm thinking no warning should be reported in the following case:

```csharp
using Tasks = System.Threading.Tasks;

namespace Namespace
{
  using Task = Tasks.Task;
}
```

#### References to intrinsics

```csharp
using Dictionary = System.Collections.Generic.Dictionary<int, string>;
```

#### References to types in the same namespace

I'm thinking the following should not report a warning, since only *namespaces* need to be fully-qualified for clarity.

```csharp
namespace Namespace
{
  using Class2 = Class;

  class Class { }
}
```

vweijsters:
Will do

vweijsters:
Added the requested test cases and rebased the PR.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        private static void CheckUsingDeclaration(SyntaxNodeAnalysisContext context, UsingDirectiveSyntax usingDirective)
+        {
+            if (!usingDirective.Parent.IsKind(SyntaxKind.NamespaceDeclaration))
+            {
+                // Usings outside of a namespace are always qualified.
+                return;
+            }
+
+            if (!usingDirective.StaticKeyword.IsKind(SyntaxKind.None))
```
sharwell: (Feel free to resolve this by filing a follow-up issue.)

❕  These directives sometimes need to be considered. For example:

```csharp
namespace My.First {
  internal static class Extensions { }
}

namespace My.Second {
  using static First.Extensions; // Should be My.First.Extensions
}
```

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                // using static types is not considered.
+                return;
+            }
+
+            string usingString = usingDirective.Name.ToString();
```
sharwell: (Feel free to resolve this by filing a follow-up issue.)

❕ Use the syntax tree for performance. This case is the following nesting:

* UsingDirectiveSyntax
    * QualifiedNameSyntax
        * AliasQualifiedNameSyntax

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+            var symbol = context.SemanticModel.GetSymbolInfo(usingDirective.Name, context.CancellationToken).Symbol;
+            if (symbol == null)
+            {
+                // if there is no symbol, do not proceed.
+                return;
+            }
+
+            string symbolString = symbol.ToString();
```
sharwell: 💭 It seems like this is prone to errors for different forms of generic types. In the tests, we covered `Dictionary<int, string>`. What happens if the user wrote `Collections.Dictionary<Int32, String>` instead? It should be easy to create unit tests which reveal the holes in this, and then file issues to resolve it in the future.

documentation/SA1135.md
```diff
@@ -0,0 +1,59 @@
...
+</tr>
+</table>
+
+## Cause
+
+A using directive is not qualified.
+
+## Rule description
+
+ A violation of this rule occurs when a using directive is contained within a namespace and is not qualified.
```
sharwell: :bulb: It would be good to include documentation saying that references to simple type names do not need to be qualified. Only namespace references need to be qualified.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/ReadabilityResources.resx
```diff
@@ -522,6 +522,21 @@
   <data name="SA1139CodeFix" xml:space="preserve">
     <value>Use literal suffix notation instead of casting</value>
   </data>
+  <data name="SA1135CodeFix" xml:space="preserve">
+    <value>Qualify using directive</value>
+  </data>
+  <data name="SA1135Description" xml:space="preserve">
+    <value>All using directives must be qualified.</value>
```
sharwell: (Feel free to resolve this by filing a follow-up issue.)

❕ This feature needs to be updated to account for the change in wording from *must* to *should*.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
vweijsters: See #2618

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
vweijsters: Modified to use `UsingDirectiveSyntaxHelpers.HasNamespaceAliasQualifier`

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
vweijsters: See #2619

documentation/SA1135.md
vweijsters: I've updated the documentation to state that alias definitions to types within the same namespace do not need to be qualified. The current implementation does no checking against simple types, not sure it that is important enough to add, but please file an issue if it should be corrected.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/ReadabilityResources.resx
vweijsters: ✔️

PR #2615, opened 2018-02-02
SA1633 Missing from Default ruleset.

ElanHasson:
I noticed SA1633 was missing from the default ruleset.


sharwell:
SA1633 is enabled by default for any project which has StyleCop Analyzers installed. It only needs to be added to the rule set if you want to change the severity to something other than Warning. 😄

ElanHasson:
@sharwell Yes-- I wanted to disable it. I didn't see how to do that other than a ruleset, and I noticed it was missing. Shouldn't it be included as Warning, just as the others are there as Warnings by default?

sharwell:
The rule set in this project only applies when you are working on StyleCop Analyzers itself. After installing StyleCop Analyzers in another project, you can modify the rule set there to disable SA1633:

```
<Rule Id="SA1633" Action="None" />
```

ElanHasson:
Got it-- thanks for clarification.

PR #2627, opened 2018-02-24
Migrate to portable PDB

sharwell:
* Fix the **opencover-report.ps1** script, and use it for AppVeyor builds
* Generate and package portable PDBs, which are much smaller and more easily consumed on other platforms
* Implement a conversion layer so we retain full OpenCover support even though the build is producing portable PDBs

### DIFF
--- a/build/opencover-report.ps1
+++ b/build/opencover-report.ps1
@@ -30,7 +30,8 @@ $opencover_console = "$packages_folder\OpenCover.$opencover_version\tools\OpenCo
 $xunit_runner_console = "$packages_folder\xunit.runner.console.$xunitrunner_version\tools\xunit.console.x86.exe"
 $report_generator = "$packages_folder\ReportGenerator.$reportgenerator_version\tools\ReportGenerator.exe"
 $report_folder = '.\OpenCover.Reports'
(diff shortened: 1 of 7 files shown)


build/opencover-report.ps1
```diff
@@ -24,32 +29,121 @@ $packageConfig = [xml](Get-Content ..\.nuget\packages.config)
...
+
+			# Workaround for https://github.com/OpenCover/opencover/issues/800
+			Copy-Item $assembly $outputDir
+		}
+	}
+}
+
+$target_dir = [IO.Path]::GetDirectoryName($target_dll)
+Get-ChildItem $target_dir -Filter *.dll | Foreach-Object { Convert-Coverage-Pdb -assembly $_.FullName -outputDir $symbols_folder_csharp6 }
+Get-ChildItem $target_dir -Filter *.dll | Foreach-Object { Extract-Coverage-Pdb -assembly $_.FullName -outputDir $symbols_folder_csharp6 }
```
vweijsters: ❓ Why is the Extract-Coverage-Pdb step necessary? As far as I can see, due to the changes made all builds will produce PDB files anyway, so the Convert-Coverage-Pdb should be enough right?

sharwell: I'm planning to use this to document a general approach for other repositories already using portable and/or embedded PDBs to adopt OpenCover as part of their build pipelines.

PR #2635, opened 2018-03-11
support .stylecop.json configuration file name

dotarj:
This PR adds dot file naming support for the StyleCop configuration file so that stylecop.json and .stylecop.json are supported file names. Just like many other configuration files (.gitignore, .appveyor.yml, .travis.yml, .codecov.yml, .editorconfig, ...) it would be nice if StyleCop would support this kind of configuration file naming convention.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsFileCodeFixProviderUnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsFileCodeFixProviderUnitTests.cs
@@ -79,7 +79,7 @@ namespace NamespaceName
         public async Task TestDotPrefixedSettingsFileAlreadyExistsAsync()
         {
             this.createSettingsFile = true;
-            this.settingsFileName = $".{SettingsHelper.SettingsFileName}";
+            this.settingsFileName = SettingsHelper.AltSettingsFileName;
 
             var offeredFixes = await this.GetOfferedCSharpFixesAsync(TestCode).ConfigureAwait(false);
             Assert.Empty(offeredFixes);
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsFileCodeFixProviderUnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsFileCodeFixProviderUnitTests.cs
@@ -26,6 +26,7 @@ namespace NamespaceName
 ";
 
         private bool createSettingsFile;
+        private string settingsFileName = SettingsHelper.SettingsFileName;
 
         /// <summary>
         /// Verifies that a file without a header, but with leading trivia will produce the correct diagnostic message.
@@ -64,6 +65,21 @@ namespace NamespaceName
         public async Task TestSettingsFileAlreadyExistsAsync()
         {
             this.createSettingsFile = true;
+            this.settingsFileName = SettingsHelper.SettingsFileName;
+
+            var offeredFixes = await this.GetOfferedCSharpFixesAsync(TestCode).ConfigureAwait(false);
+            Assert.Empty(offeredFixes);
+        }
+
+        /// <summary>
+        /// Verifies that a code fix will not be offered if the settings file is already present.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestDotPrefixedSettingsFileAlreadyExistsAsync()
+        {
+            this.createSettingsFile = true;
+            this.settingsFileName = $".{SettingsHelper.SettingsFileName}";
 
             var offeredFixes = await this.GetOfferedCSharpFixesAsync(TestCode).ConfigureAwait(false);
             Assert.Empty(offeredFixes);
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
```diff
@@ -66,14 +66,32 @@ internal static StyleCopSettings GetStyleCopSettings(this AnalyzerOptions option
...
         }
 
+        /// <summary>
+        /// Gets a value indicating whether the specified path points to a StyleCop settings file (stylecop.json or .stylecop.json).
+        /// </summary>
+        /// <param name="path">The path to test.</param>
+        /// <returns><c>true</c> if <paramref name="path"/> points to a StyleCop settings file; otherwise, <c>false</c>.</returns>
+        internal static bool IsStyleCopSettingsFile(string path)
+        {
+            var fileName = Path.GetFileName(path).ToLowerInvariant();
```
vweijsters: This also allows `stylecop.json` and `.stylecop.json` files in folders other than the root folder of the project, which is a change from the current behavior.

vweijsters: It would also be better to use a case insensitive string compare, instead of converting to lower case first, as this saves additional memory allocations.

vweijsters:
I've marked this `do not merge` pending the required discussion.

vweijsters:
For the discussion: I would be OK with this change, once the review comments have been addressed.

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
sharwell: > This also allows stylecop.json and .stylecop.json files in folders other than the root folder of the project, which is a change from the current behavior.

This is already indirectly possible using linked files.

sharwell: :memo: This approach appears to be a simple preservation of the original behavior.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsFileCodeFixProviderUnitTests.cs
```diff
@@ -64,6 +65,21 @@ namespace NamespaceName
...
+
+        /// <summary>
+        /// Verifies that a code fix will not be offered if the settings file is already present.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        public async Task TestDotPrefixedSettingsFileAlreadyExistsAsync()
+        {
+            this.createSettingsFile = true;
+            this.settingsFileName = $".{SettingsHelper.SettingsFileName}";
```
sharwell: ❗️ Define a second constant `AltSettingsFileName`:

```csharp
public const string AltSettingsFileName = ".stylecop.json";
```

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
```diff
@@ -159,24 +178,28 @@ private static StyleCopSettings GetStyleCopSettings(string path, SourceText text
             return new StyleCopSettings();
         }
 
-        private static SourceText TryGetStyleCopSettingsText(this AnalyzerOptions options, CancellationToken cancellationToken)
+        private static SourceText TryGetStyleCopSettingsText(this AnalyzerOptions options, CancellationToken cancellationToken, out string settingsFileName)
```
sharwell: :bulb: This is a file path, not a file name. It would be good to reflect this in the parameter name.

sharwell:
I'm not a huge fan of the change, but I understand where it's coming from and wouldn't block it if the issues are addressed. Overall the code seems a bit cleaner than it was before so that doesn't hurt.

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
dotarj: The string comparison is changed. This change did cause a unit test to fail which expected a NullReferenceException to be unhandled (when path is null, calling .ToLowerInvariant() causes an NullReferenceException). I've added a null check and now an ArgumentNullException is thrown when the path is null (the test works again).

StyleCop.Analyzers/StyleCop.Analyzers.Test/Settings/SettingsFileCodeFixProviderUnitTests.cs
dotarj: I've added the AltSettingsFileName field to the SettingsHelper class. The implementation of IsStyleCopSettingsFile is also updated to use the new field.

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
dotarj: Good catch, the parameter name is changed.

dotarj:
Still working on updating documentation/Configuration.md by the way. :)

sharwell:
> Still working on updating documentation/Configuration.md by the way.

👍 I am waiting for that and then for @vweijsters to verify that it covers what he was looking for.

dotarj:
The documentation is updated. Please let me know if you have any comments.

PR #2655, opened 2018-04-13
Fixed issue in SA1500 with multi-dimensional array initializers

### LINKED ISSUE #2632: SA1500 is producing a false positive
from @JimBobSquarePants (in #2607):

I'm still seeing this for multiline arrays in 1.1.0-beta006

e.g.

```C#
        private static readonly float[,] AtkinsonMatrix =
            new float[,]
            {
               { 0, 0, 1, 1 },
               { 1, 1, 1, 0 },
               { 0, 1, 0, 0 }
            };
```

This construction works fine in 1.1.0-beta004 and no warnings should be raised here.

vweijsters:
Fixes #2632

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1500/SA1500UnitTests.ArrayInitializers.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1500/SA1500UnitTests.ArrayInitializers.cs
@@ -433,5 +433,94 @@ public class TestClass
             await this.VerifyCSharpDiagnosticAsync(fixedTestCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
             await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
         }
+
+        /// <summary>
+        /// Verifies that a jagged array initialization produces the expected diagnostics.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1500/SA1500UnitTests.ArrayInitializers.cs
```diff
@@ -342,5 +342,96 @@ public void TestMethod()
...
             await this.VerifyCSharpFixAsync(testCode, fixedTestCode).ConfigureAwait(false);
         }
+
+        /// <summary>
+        /// Verifies that a multi-dimensional array initialization produces the expected diagnostics.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        [WorkItem(2632, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2632")]
+        public async Task VerifyMultidimensionalArrayInitializationAsync()
```
sharwell: ❓ Can you add an additional test covering the same test cases but using a jagged array `int[][]` with elements having the form `new[] { 0, 0, 1, 1 }`?

vweijsters:
I've added the additional test cases. The following test case:

```c#
    private static readonly int[][] TestMatrix4 =
        new int[][]
        {
            new[] { 0, 0, 1, 1 }, new[] { 1, 1, 1, 0 },
            new[] { 0, 1, 0, 0 }
        };
```

 does not produce a SA1500 diagnostic and nor do I think that it should.

PR #2659, opened 2018-04-22
SA1137: Each switch case with a block is checked separately

### LINKED ISSUE #2527: SA1137: Unexpected behaviour with switch statements
Suppose I have a method with a switch statement in it:

```
private string ToLongHand(int x)
{
    switch (x)
    {
        case 1:
            return "One";
        case 2:
        {
            // Do some complicated logic
            return "Two";
        }

        default:
            throw new ArgumentOutOfRangeException(nameof(x), "I cant count past 2...");
    }
}
```

SA1137 is raised for the brace immediately after case 2. Is that intended? Whilst personally I think this code is acceptable, I might be persuaded by a rule saying that all cases should have a brace or none of them should have.

Additionally, when I run the fix, it converts the code to the following:

```
private string ToLongHand(int x)
{
    switch (x)
    {
        case 1:
            return "One";
        case 2:
            {
            // Do some complicated logic
            return "Two";
        }

        default:
            throw new ArgumentOutOfRangeException(nameof(x), "I cant count past 2...");
    }
}
```

... which is definitely worse than before!

Theres also an issue if the first case statement has a brace and the second one doesnt - it highlights the second block and suggests that it should be aligned with the start of the "case" statement

I'm using StyleCop.Analyzers.1.1.0-beta004 on Visual Studio 2017 (v15.3.2)

morsiu:
Hello, I made a change in SA1137 that affects handling of `case`s in a `switch`.
After the change, when statements within a `case` are enclosed in a block, they are checked for consistent indentation within that block only, therefore independently from statements of other `case`s.

Fixes #2527

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1137UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1137UnitTests.cs
@@ -1460,18 +1460,37 @@ class ClassName
 
                 break;
 
-       case 1:
-case 2:
-     label2:
+        case 1:
+       {
+    label2:
+           if (true)
+           {
+           }
+
+               break;
+       }
+       case 2:
+case 3:
+     label3:
            while (true)
             {
             }
 
            break;
 
+      case 4:
+case 5:
+       {
+        label4:
+          while (true)
+            {
+            }
+
+          break;
+       }
 default:
-label3a:
- label3b:
+label5a:
+ label5b:
 break;
         }
     }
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1137UnitTests.cs
```diff
@@ -1493,17 +1512,36 @@ void MethodName()
...
+        case 4:
+        case 5:
+       {
+        label4:
+          while (true)
+            {
+            }
+
+          break;
+        }
```
sharwell: :bug: A separate bug needs to be filed for this case.

sharwell: ➡️ Filed #2665

PR #2668, opened 2018-05-02
Support empty lines in header comments

### LINKED ISSUE #2657: How to use empty line as header decoration without triggering SA1028, SA1120
I'm using version `1.1.0-beta006`, and I have below settings:

```json
  "settings": {
    "documentationRules": {
      "companyName": "Foo",
      "copyrightText": "\r\nCopyright (c) {companyName} Corporation. All rights reserved.\r\n",
      "xmlHeader": false
    }
```

What I want to achieve is that:
```csharp
//<- no spaces here
// Copyright (c) Foo Corporation. All rights reserved.
//<- no spaces here
```
But the generated header contains an extra space in the 1st and 3rd line, like this:

```csharp
// <- 1 space here
// Copyright (c) Foo Corporation. All rights reserved.
// <- 1 space here
```

And it's violating [SA1028](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1028.md), if I removed the extra space manually, it's violating [SA1120](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1120.md).
How can I achieve the result I expected, without turning off SA1028 and SA1120?

sharwell:
Fixes #2657

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/FileHeaderCodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/FileHeaderCodeFixProvider.cs
@@ -7,6 +7,7 @@ namespace StyleCop.Analyzers.DocumentationRules
     using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Composition;
+    using System.Linq;
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/FileHeaderCodeFixProvider.cs
```diff
@@ -139,8 +140,7 @@ private static SyntaxNode ReplaceWellFormedMultiLineCommentHeader(Document docum
             triviaStringParts[0] = commentIndentation + interlinePadding + " " + triviaStringParts[0];
             StringBuilder sb = StringBuilderPool.Allocate();
             string fileName = Path.GetFileName(document.FilePath);
-            var copyrightText = commentIndentation + interlinePadding + " " +
-                GetCopyrightText(commentIndentation + interlinePadding, settings.DocumentationRules.GetCopyrightText(fileName), newLineText);
```
sharwell: :memo: This was the only change in the second revision (I amended the original commit to fix build test errors)

PR #2673, opened 2018-05-07
Make fields readonly

sharwell:


vweijsters:
Just out of curiosity: What is the benefit of having everything readonly?

jizc:
I think the motivation may be based on this: https://www.jetbrains.com/help/resharper/FieldCanBeMadeReadOnly.Local.html

Paraphrasing:
> Additional safeguard to ensure that fields that are initialized only via the constructor, will not inadvertently be assigned from anywhere within its methods.

sharwell:
@vweijsters This primarily improves maintainability (easier to reason about behavior of read-only and immutable types).

PR #2678, opened 2018-05-09
Update to StyleCop Analyzers 1.1.0-beta007

sharwell:


StyleCop.Analyzers/StyleCop.Analyzers.ruleset
```diff
@@ -75,5 +75,6 @@
     <Rule Id="SA1412" Action="Warning" />
     <Rule Id="SA1600" Action="None" />
     <Rule Id="SA1609" Action="Warning" />
+    <Rule Id="SA1629" Action="None" />
```
sharwell: :memo: I wasn't happy with the number of false/questionable positives and ended up disabling the rule.

PR #2694, opened 2018-05-22
SA1629 reporting

### LINKED ISSUE #2679: false positive SA1629.
I have remarks with code:

```xml

        /// <remarks>
        /// Sample request:
        /// <code>
        ///     DELETE api/v1/regions/f788b478-b0df-e511-80ee-000c2933216e
        /// </code>
        /// </remarks>

```

And stylecop is complaining about.

### LINKED ISSUE #2680: SA1629 bug
The following code:

```c#
        /// <summary>
        /// The ID for diagnostics produced by the <see cref="SA1127GenericTypeConstraintsMustBeOnOwnLine"/>
        /// </summary>
        public const string DiagnosticId = "SA1127";
```
will flag a missing period just after the `the` word, where it should be after the `<see .../>` tag.

sharwell:
Fixes #2680 

Fixes support for interfaces in SA1608 and SA1629

Closes #2679

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
@@ -12,6 +12,7 @@ namespace StyleCop.Analyzers.DocumentationRules
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Microsoft.CodeAnalysis.Diagnostics;
     using Microsoft.CodeAnalysis.Text;
+    using StyleCop.Analyzers.Helpers;
 
     /// <summary>
     /// A section of the XML header documentation for a C# element does not end with a period (also known as a full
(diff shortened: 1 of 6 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
```diff
@@ -91,14 +98,20 @@ protected override void HandleXmlElement(SyntaxNodeAnalysisContext context, bool
...
 
                                 elementDone = true;
                             }
                         }
                     }
+                    else if (xmlElement.Content[i].IsInlineElement())
+                    {
+                        // If a diagnostic gets reported for the element, place the diagnostic after the last inline
+                        // element.
+                        reportingLocation = xmlElement.Content[i].Span.End;
```
vweijsters: This will fail for 2 edge cases:
- ```<summary>Text <see cref="x"/><see cref="y"/></summary>``` will report the wrong location.
- ```<summary>Test text. <see cref="x"/></summary>``` will not report anything and it should.

I think that when an inline element is found, there will not be a period, so a diagnostic can be reported straight away and the loop can be ended. This should solve both edge cases.

sharwell: I meant to write this:

```
reportingLocation = reportingLocation ?? ...;
```

That explains the first issue. I need to think more about the handling of empty elements because it's possible for other situations to occur with non-empty elements.

sharwell: @vweijsters I fixed the first case in ecee2c7e206890e27fb73d7bffc0c5938773d6e5 and then fixed the second case in bf6fec59f8f5e7f2692a12fc0ef5f55fb644a38d.

PR #2703, opened 2018-05-23
Made SA1130 codefix properly expand parameters when only delegate key…

### LINKED ISSUE #2593: SA1130 Code Fix Generates Incorrect Code
# Description
The line below, tagged "this line", flags SA1130:
## Sample 1: Violation
```csharp
using System;

namespace StyleCopDemo
{
    class Program
    {
        {
            // this line:
            EventHandler anon = delegate { Console.WriteLine("hello"); };
        }
    }
}
```

The fix, however, transforms it into something that's syntactically incorrect:
## Sample 2: Incorrect Fix
```csharp
using System;

namespace StyleCopDemo
{
    class Program
    {
        {
            // this line:
            EventHandler anon = () => { Console.WriteLine("hello"); };
        }
    }
}
```

If there's going to be a code fix for this, then a legal fix would be:
## Sample 3: Legal Fix
```csharp
using System;

namespace StyleCopDemo
{
    class Program
    {
        {
            // this line:
            EventHandler anon = (sender, args1) => { Console.WriteLine("hello"); };
        }
    }
}
```

# Environment
Visual Studio Enterprise 15.5.3
StyleCop.Analyzers: version 1.1.0-beta004 from NuGet

.NET Core 2.0, with the following .csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004" />
  </ItemGroup>
</Project>
```

vweijsters:
…word is used

Fixes #2593

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
@@ -65,15 +65,13 @@ namespace StyleCop.Analyzers.ReadabilityRules
 
         private static void HandleAnonymousMethodExpression(SyntaxNodeAnalysisContext context)
         {
-            var diagnosticProperties = ImmutableDictionary.CreateBuilder<string, string>();
-
             bool reportDiagnostic = true;
             var anonymousMethod = (AnonymousMethodExpressionSyntax)context.Node;
 
             switch (anonymousMethod.Parent.Kind())
             {
             case SyntaxKind.Argument:
-                reportDiagnostic = HandleMethodInvocation(context.SemanticModel, anonymousMethod, (ArgumentSyntax)anonymousMethod.Parent, diagnosticProperties);
+                reportDiagnostic = HandleMethodInvocation(context.SemanticModel, anonymousMethod, (ArgumentSyntax)anonymousMethod.Parent);
                 break;
 
             case SyntaxKind.EqualsValueClause:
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -47,35 +54,130 @@ public override void Initialize(AnalysisContext context)
...
+            foreach (var symbolParameter in symbolParameters)
+            {
+                var syntaxParameter = GetParameterSyntaxFromParameterSymbol(symbolParameter);
+                result.Add(syntaxParameter);
+            }
+
+            return result.ToImmutable();
+        }
+
+        private static ParameterSyntax GetParameterSyntaxFromParameterSymbol(IParameterSymbol symbolParameter)
```
sharwell: ❓ Is it possible to move this logic into the code fix instead? Moving it would improve the performance of the main analyzer by not requiring as much symbol analysis, and then the code fix is a bit slower but at that point users have explicitly requested the system take action.

If the new analysis logic affects when/where diagnostics are reported (and thus can't be moved), can you call those cases out?

vweijsters: I'll have a look

vweijsters: @sharwell I've moved the code necessary to determine the correct parameter list to the code fix. Can you have a look again?

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -45,37 +47,104 @@ public override void Initialize(AnalysisContext context)
...
+                break;
+            }
+
+            if (reportDiagnostic)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Descriptor, anonymousMethod.DelegateKeyword.GetLocation(), diagnosticProperties.ToImmutable()));
+            }
+        }
+
+        private static bool HandleMethodInvocation(SemanticModel semanticModel, AnonymousMethodExpressionSyntax anonymousMethod, ArgumentSyntax argumentSyntax, ImmutableDictionary<string, string>.Builder propertiesBuilder)
```
sharwell: :bulb: It looks like `propertiesBuilder` is no longer used

vweijsters: Yes, I've missed that one. Its fixed now.

PR #2706, opened 2018-05-30
Status site workaround

sharwell:


sharwell:
Didn't work like I hoped.

PR #2709, opened 2018-05-31
Automatically include stylecop.json

### LINKED ISSUE #2698: Automatically consume stylecop.json if it exists
I'd like to contribute a feature to the NuGet package where StyleCop can automatically consume my `stylecop.json`.

I'd like to add a file to the package `build\StyleCop.Analyzers.props` which will search the directory structure of a code base for `stylecop.json` and include it.  Alternatively, users can just set a property like `StyleCopJsonPath` and it will get added appropriately.

```xml
<Project>
  <PropertyGroup>
    <!-- Walk up the directory tree looking for a stylecop.json -->
    <StyleCopJsonPath Condition=" '$(StyleCopJsonPath)' == '' ">$([MSBuild]::GetPathOfFileAbove('stylecop.json'))<StyleCopJsonPath>
  </PropertyGroup>

  <!-- Add $(StyleCopJsonPath) to <AdditionalFiles /> if the property is set and exists -->
  <ItemGroup Condition=" '$(StyleCopJsonPath)' != '' And Exists($(StyleCopJsonPath)) ">
    <AdditionalFiles Include="$(StyleCopJsonPath)">
      <Link>stylecop.json</Link>
    </AdditionalFiles>
  </ItemGroup>
</Project>
```

This would make it so I can just add a `<PackageReference />` and create a `stylecop.json` in the root of my repository.  Thoughts?

jeffkl:
For legacy projects and older MSBuild, simply add the item as hidden and it will show up as a None item but still be passed as an AdditionalFiles item

For newer MSBuild, remove stylecop.json from None and add it to AdditionalFiles unless its already been done by the user.

Fixes #2698

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -44,7 +44,7 @@ Describes the configuration options for StyleCop.Analyzers
 
 **[How to enable the configuration](documentation/EnableConfiguration.md)**
 
-Describes how to enable the **stylecop.json** file for usage.
+Describes how to enable the **stylecop.json** file for usage in version 1.1.0 and below of the StyleCop.Analyzers package.
 
 **[Known changes](documentation/KnownChanges.md)**
 
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuild15.props
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuild15.props
@@ -1,0 +1,35 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<!--
+  Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+  Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition="Exists('stylecop.json')">
+    <!--
+      If stylecop.json exists in the project directory add it to <AdditionalFiles /> automatically.  The point of
+      this is that the user sees the file in Visual Studio Solution Explorer but they don't understand that they
+      need to add it to the <AdditionalFiles /> item group.
+      
+      1. Legacy project system - Files in the project directory are in an explicit <None /> item.
+      2. New project system - Files in the project directory are implicitly added to the <None /> item through a glob.
+    -->
+    <None Remove="stylecop.json" />
+
+    <AdditionalFiles Include="stylecop.json" Condition=" '@(AdditionalFiles->EndsWith('stylecop.json', System.StringComparison.OrdinalIgnoreCase))' != 'true' " />
+  </ItemGroup>
+
+  <ItemGroup Condition="!Exists('stylecop.json')">
+    <!--
+      If stylecop.json doesn't exist in the project directory, search parent directories for one.  This allows user
+      to simply create a file at the root of their repository and this will automatically include it.
+    -->
+    <AdditionalFiles Include="$([MSBuild]::GetPathOfFileAbove(stylecop.json, $(MSBuildProjectDirectory)..))"
+                     Condition=" '@(AdditionalFiles->EndsWith('stylecop.json', System.StringComparison.OrdinalIgnoreCase))' != 'true' And Exists('$([MSBuild]::GetPathOfFileAbove(stylecop.json, $(MSBuildProjectDirectory)..))')">
+      <!--
+        Show the file as a link as if its in the project directory so users can open it up in Visual Studio easily
+        and they know its affecting each project.
+      -->
+      <Link>stylecop.json</Link>
+    </AdditionalFiles>
+  </ItemGroup>
+</Project>
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuildOld.props
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuildOld.props
@@ -1,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<!--
+  Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+  Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+    This file is imported by versions of MSBuild older than 15.0 and must use compatible syntax.
+  -->
+
+  <ItemGroup Condition=" Exists('stylecop.json')">
+    <!--
+      If stylecop.json exists in the project directory add it to <AdditionalFiles /> automatically.  The point of
+      this is that the user sees the file in Visual Studio Solution Explorer but they don't understand that they
+      need to add it to the <AdditionalFiles /> item group.
+      
+      In this case, copy it to <AdditionalFiles /> and hide it so it doesn't show up twice and the user doesn't
+      have to do anything.
+    -->
+    <AdditionalFiles Include="stylecop.json">
+      <Visible>False</Visible>
+    </AdditionalFiles>
+  </ItemGroup>
+
+  <ItemGroup Condition=" !Exists('stylecop.json')">
+    <!--
+      If stylecop.json doesn't exist in the project directory, search parent directories for one.  This allows user
+      to simply create a file at the root of their repository and this will automatically include it.
+    -->
+    <AdditionalFiles Include="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory).., stylecop.json))\stylecop.json"
+                     Condition="Exists('$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory).., stylecop.json))\stylecop.json')">
+      <!--
+        Show the file as a link as if its in the project directory so users can open it up in Visual Studio easily
+        and they know its affecting each project.
+      -->
+      <Link>stylecop.json</Link>
+    </AdditionalFiles>
+  </ItemGroup>
+</Project>
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/StyleCop.Analyzers.targets
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/StyleCop.Analyzers.targets
@@ -1,0 +1,28 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<!--
+  Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+  Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+    Attempt to automatically include stylecop.json in the <AdditionalFiles /> item group.
+    
+    To disable this functionality, set the MSBuild property "AutoIncludeStyleCopJson" to "false".  Example:
+    
+    <PropertyGroup>
+      <AutoIncludeStyleCopJson>false</AutoIncludeStyleCopJson>
+    </PropertyGroup>
+  -->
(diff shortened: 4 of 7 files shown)


jeffkl:
I tested with legacy projects with MSBuild 15.0 and MSBuild 4.0 as well as new SDK-style projects and MSBuild 15.0.

If you've already manually moved `stylecop.json` to `AdditionalFiles`, it is left alone, otherwise it is removed from `None` and added to `AdditionalFiles`.

The logic also walks up the directory structure for a `stylecop.json` so you can simply create one at the root of your repo and it will automatically be picked up.

I'm pretty sure this makes the [enabling stylecop.json documentation](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/EnableConfiguration.md) obsolete.  Should I update it?

sharwell:
> I'm pretty sure this makes the enabling stylecop.json documentation obsolete. Should I update it?

Yep 👍

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuildOld.props
```diff
@@ -0,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<!--
+  Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+  Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
sharwell: ❓ If we leave out the `xmlns`, will it still work in Visual Studio 2015?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuild15.props
```diff
@@ -0,0 +1,35 @@
...
+      
+      1. Legacy project system - Files in the project directory are in an explicit <None /> item.
+      2. New project system - Files in the project directory are implicitly added to the <None /> item through a glob.
+    -->
+    <None Remove="stylecop.json" />
+
+    <AdditionalFiles Include="stylecop.json" Condition=" '@(AdditionalFiles->EndsWith('stylecop.json', System.StringComparison.OrdinalIgnoreCase))' != 'true' " />
+  </ItemGroup>
+
+  <ItemGroup Condition="!Exists('stylecop.json')">
```
sharwell: ❓ Why separate this into the two cases? It seems like it would be simpler to combine the two cases into one.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuild15.props
```diff
@@ -0,0 +1,35 @@
...
+    <!--
+      If stylecop.json doesn't exist in the project directory, search parent directories for one.  This allows user
+      to simply create a file at the root of their repository and this will automatically include it.
+    -->
+    <AdditionalFiles Include="$([MSBuild]::GetPathOfFileAbove(stylecop.json, $(MSBuildProjectDirectory)..))"
+                     Condition=" '@(AdditionalFiles->EndsWith('stylecop.json', System.StringComparison.OrdinalIgnoreCase))' != 'true' And Exists('$([MSBuild]::GetPathOfFileAbove(stylecop.json, $(MSBuildProjectDirectory)..))')">
```
sharwell: :memo: Seems like this condition is unnecessary (it comes up a few other places as well). During upgrade, users can either add the `AutoIncludeStyleCopJson` property or delete their own explicit `<AdditionalFiles>` elements to account for the change in behavior.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/StyleCop.Analyzers.targets
```diff
@@ -0,0 +1,28 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
sharwell: ❗️ This needs to be **StyleCop.Analyzers.props**.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/StyleCop.Analyzers.targets
```diff
@@ -0,0 +1,28 @@
...
+  <!--
+    Attempt to automatically include stylecop.json in the <AdditionalFiles /> item group.
+    
+    To disable this functionality, set the MSBuild property "AutoIncludeStyleCopJson" to "false".  Example:
+    
+    <PropertyGroup>
+      <AutoIncludeStyleCopJson>false</AutoIncludeStyleCopJson>
+    </PropertyGroup>
+  -->
+  <ImportGroup Condition=" '$(AutoIncludeStyleCopJson)' != 'false' ">
```
sharwell: :memo: Might need to move this condition to the `<ItemGroup>` elements in the imported files since the import group will be evaluated before the properties in the main project file take effect.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuild15.props
```diff
@@ -0,0 +1,35 @@
...
+<!--
+  Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+  Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition="Exists('stylecop.json')">
+    <!--
+      If stylecop.json exists in the project directory add it to <AdditionalFiles /> automatically.  The point of
+      this is that the user sees the file in Visual Studio Solution Explorer but they don't understand that they
+      need to add it to the <AdditionalFiles /> item group.
```
sharwell: > The point of this is that the user sees the file in Visual Studio Solution Explorer but they don't understand that they need to add it to the &lt;AdditionalFiles /&gt; item group.

This text makes sense for the feature request here on GitHub, but in the implementation itself you can focus more on the 'what'. It's probably safe to leave this sentence out.

documentation/EnableConfiguration.md
```diff
@@ -1,8 +1,15 @@
...
 
-At this time, the code fix is not able to fully configure the newly-created **stylecop.json** file for use. This is
-tracked in bug report [dotnet/roslyn#4655](https://github.com/dotnet/roslyn/issues/4655). In the mean time, users must
-manually perform the following additional steps after creating the **stylecop.json** file.
+## StyleCop.Analyzers v1.2 and above
+
+If you are using version 1.1.1 or newer, you no longer need to include stylecop.json using the methods below.  It is
+automatically included if it is next to the project or in a directory above the project.  
+
+## StyleCop.Analyzers v1.1 and below
```
sharwell: :bulb: Technically there is not yet a v1.1 stable package, so we don't need to increment the version. If you want, you can be specific in the text above to say:

> If you are using version 1.1.0 RC 2 or newer and place your **stylecop.json** file in the project directory or one of its ancestors (often including the solution directory), ...

DOCUMENTATION.md
```diff
@@ -44,7 +44,7 @@ Describes the configuration options for StyleCop.Analyzers
 
 **[How to enable the configuration](documentation/EnableConfiguration.md)**
 
-Describes how to enable the **stylecop.json** file for usage.
+Describes how to enable the **stylecop.json** file for usage in version 1.1.0 and below of the StyleCop.Analyzers package.
```
sharwell: :bulb: This probably doesn't need to change, because the configuration instructions are still relevant for users who place **stylecop.json** in a directory which is not on a path to the root from the project file.

sharwell:
:bulb: You can probably remove the following text:

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/065af6a475b335722b7f693626c2d7f3b40290e5/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/Settings/SettingsFileCodeFixProvider.cs#L27-L31

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuild15.props
jeffkl: If `stylecop.json` is in the project directory, there is almost guaranteed to already be a `<None />` item pointing to it.  I wanted to treat this case slightly differently by moving it to `<AdditionalItems />`.  However, if there is a `stylecop.json` in a parent folder, it gets added with a `Link` and any existing items are left alone.  

Technically I could combine them, I was just trying to be more exact.  I'm worried about the upgrade experience for users as they try out new versions of the package.  They already would have done the manual steps and upgrading could cause duplicates to show up in Solution Explorer.  If you're fine with some release notes documentation or something, I can simplify some of this.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuildOld.props
jeffkl: I am not sure, I didn't have a machine with Visual Studio 2015 to test it out.  Its probably safer to have it unless you have a strong opinion to leave it out.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/StyleCop.Analyzers.targets
jeffkl: If this is a `.props` file, it is imported _before_ the project items are declared.  This would make it impossible to move the `None` item to `AdditionalFiles`.  So I used `.targets` so the items are imported after the project items.

jeffkl:
I did some more testing and found a few snags.  I'm investigating how to make this work for all scenarios...

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/StyleCop.Analyzers.targets
sharwell: It's important that the AdditionalFiles items be added in a .props file (before the project file is evaluated), or it becomes impossible to use `Update` functionality inside the project file to modify the included element.

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/build/MSBuild15.props
sharwell: > If you're fine with some release notes documentation or something, I can simplify some of this.

If we get all the edge cases working, I'm totally fine with this.

sharwell:
@jeffkl I'm going to get the 1.1.0-beta008 out the door and then we can try again for beta009 for this change 😄

jeffkl:
Okay I might just close this PR.  There are a few complications:

1. Legacy projects don't show linked files in Solution Explorer.  If you have something like this in an import:

```xml
<ItemGroup>
  <AdditionalFiles Include="..\..\stylecop.json">
    <Link>stylecop.json</Link>
  </AdditionalFiles>
</ItemGroup>
```

It doesn't show up in Solution Explorer.  It still gets passed to the compiler and the stylecop config is read but its really hard to tell in Visual Studio where the configuration is coming from unless you manually add `stylecop.json` as a Solution Item.  It feels like a bad user experience.

2. New SDK-style projects don't need explicit items by default.  The auto-fix logic in StyleCop place an item directly in the project:

```xml
<ItemGroup>
  <None Include="stylecop.json" />
</ItemGroup>
```

But out-of-the-box SDK-style projects [include most files](https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#L30) in `None` items unless you've set `EnableDefaultItems` to `false`.  I ran the fix logic under the debugger to try and come up with a way to make it work but the Roslyn Workspace object just doesn't have enough flexibility to read evaluated MSBuild properties or to add the file to disk but not add an item to the project.

So for SDK-style projects you end up with an item in the project and the import removes it and adds back one from a parent directory and again, it feels like a bad user experience. 

There does not appear to be a way to tell from a RoslynWorkspace if the project is legacy or SDK-style so its hard to have more advanced logic.  Perhaps you know if there is a way to get MSBuild properties?  Or to not add an explicit item to the project and just create the file?

PR #2719, opened 2018-06-08
$schema value should end with '#'

skurdiukov:
Current json schema contains invalid schema reference, it should and with '#'


skurdiukov:
This is predefined value for draft-04 schema, as described in https://tools.ietf.org/html/draft-zyp-json-schema-04#section-6

PR #2727, opened 2018-06-13
Update to StyleCop Analyzers 1.1.0-beta008

sharwell:


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
```diff
@@ -15,7 +15,7 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
     using Xunit;
 
     /// <summary>
-    /// This class contains unit tests for <see cref="SA1642ConstructorSummaryDocumentationMustBeginWithStandardText"/>-
+    /// This class contains unit tests for <see cref="SA1642ConstructorSummaryDocumentationMustBeginWithStandardText"/>.
```
sharwell: 💭 This required a manual correction, but we might consider supporting it directly.

vweijsters: We probably could support it, but I doubt if it is worth the effort. Determining if something is a typo might get quite difficult and I doubt this will happen that much that a code fix for it is really needed.

PR #2731, opened 2018-06-15
Add two more fix all providers

pdelvo:
Here are two more fix all providers for SA1517 and the file header fixers. I was able to observe a ~5x speed up for the file header fixer compared to the batch fixer.


sharwell:
@pdelvo I noticed that the status site isn't showing the ⌛️ for code fixes that don't have one of the new Fix All providers. Must have been lost in the conversion if you want to take a look.

PR #2742, opened 2018-07-10
Allow generic type to be followed by colon token for switch pattern m…

### LINKED ISSUE #2735: SA1024 conflicting with SA1015 in pattern matching
I'm using a pattern matching expression in a switch statement, like this:
<code><pre>case nameof(MyItem.MyProperty) when value is List&lt;MyClass&gt;:</pre></code>

Written like this I'm getting 

> SA1015	Closing generic bracket should be followed by a space.

But then if I add a space between the closing generic bracket and the colon I get:

> SA1024	Colon should not be preceded by a space.

I have no other solution but to suppress here. This shouldn't be happening.

vweijsters:
…atching

Fixes #2735

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1015CSharp7UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1015CSharp7UnitTests.cs
@@ -50,5 +50,33 @@ public class Foo
             await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
             await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
         }
+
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1015CSharp7UnitTests.cs
```diff
@@ -50,5 +50,33 @@ public void TestMethod()
             await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
             await this.VerifyCSharpFixAsync(testCode, fixedCode, cancellationToken: CancellationToken.None).ConfigureAwait(false);
         }
+
+        /// <summary>
+        /// Verifies that a pattern in a switch statement that ends with a generic type does not produce a diagnostic.
```
sharwell: :memo: I've been treating `WorkItem` as sufficient to describe the reason why a *specific* regression test was created, but if you want to keep adding descriptive comments I won't stop you. 😄

PR #2743, opened 2018-07-10
Made the top-level tags that will not be analyzed by SA1629 configura…

### LINKED ISSUE #2721: SA1629 reports false-positives for author tags
We have encountered false-positives for the author tags in our code after updating to 1.1.0-beta007.

```csharp
/// <summary>
///     Gets or sets the name.
/// </summary>
/// <author> john.doe@example.com </author>
public string Name { get; set; }
```

The warning says that the documentation (in this case the mail address) should end with a period.

vweijsters:
…ble.

Fixes #2721

I've implemented option 4 (see https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2721#issuecomment-396066873)


vweijsters:
Updated. I went with `excludeFromPunctuationCheck` as that seems more appropriate as `excludeFromGrammarChecks` to me.

PR #2751, opened 2018-08-09
Update unit tests to Microsoft.CodeAnalysis.Testing

sharwell:
I was looking at dotnet/roslyn-sdk#123 and considering what it might look like to switch from a source code distribution to a NuGet-based distribution. This is a proof of concept that preserves much of the original form used for testing StyleCop Analyzers, but separates the StyleCop Analyzers-specific items from the base testing support.

/cc @mavasani @DustinCampbell @jmarolf @JohanLarsson


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1000CSharp7UnitTests.cs
```diff
@@ -8,6 +8,7 @@ namespace StyleCop.Analyzers.Test.CSharp7.SpacingRules
     using StyleCop.Analyzers.Test.SpacingRules;
     using TestHelper;
     using Xunit;
+    using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<StyleCop.Analyzers.SpacingRules.SA1000KeywordsMustBeSpacedCorrectly, StyleCop.Analyzers.SpacingRules.TokenSpacingCodeFixProvider>;
```
sharwell: :memo: This bit of magic is why the `Diagnostic()` and `new CSharpTest` references did not require new generic arguments when the inheritance was removed.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1000CSharp7UnitTests.cs
```diff
@@ -228,15 +231,16 @@ public void TestMethod()
...
+            await new CSharpTest
             {
-                this.CSharpDiagnostic().WithArguments("foreach", string.Empty, "followed").WithLocation(7, 13),
-                this.CSharpDiagnostic().WithArguments("foreach", string.Empty, "followed").WithLocation(8, 13),
-            };
-
-            await this.VerifyCSharpDiagnosticAsync(testCode, expectedDiagnostics, CancellationToken.None).ConfigureAwait(false);
-            await this.VerifyCSharpDiagnosticAsync(fixedCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
-            await this.VerifyCSharpFixAsync(testCode, fixedCode).ConfigureAwait(false);
+                TestCode = testCode,
```
JohanLarsson: I prefer ctor arguments and get only properties for required stuff.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1000CSharp7UnitTests.cs
```diff
@@ -228,15 +231,16 @@ public void TestMethod()
...
+                FixedCode = fixedCode,
+                ExpectedDiagnostics =
+                {
+                    Diagnostic().WithArguments("foreach", string.Empty, "followed").WithLocation(7, 13),
+                    Diagnostic().WithArguments("foreach", string.Empty, "followed").WithLocation(8, 13),
+                },
+            }.RunAsync(CancellationToken.None).ConfigureAwait(false);
```
JohanLarsson: Async adds a bit of noise and ways to get things wrong to tests. Doing antipattern task.wait() and exposing as blocking is an alternative.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1000CSharp7UnitTests.cs
sharwell: I tend to agree, but I like the uniformity of this approach when `TestCode` is the only required property. In the end, I'll probably have a helper method available to shorten the two primary and simplest use cases:

1. Verify code does not report diagnostics
2. Verify code reports diagnostics, and the fixer fixes all of them

The longer form seen here becomes only necessary for more complicated cases.

JohanLarsson:
Why pass in explicit empty collection of diagnostic results when testing valid code?

```cs
await this.VerifyCSharpDiagnosticAsync(testCode, EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
```

vs
```cs
await VerifyValidAsync(testCode).ConfigureAwait(false);
```
Also removing async:

```cs
VerifyValid(testCode);
```

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1000CSharp7UnitTests.cs
sharwell: ➡️ This is unlikely to change for StyleCop Analyzers. Also, while the version eventually going to dotnet/roslyn-sdk is likely to continue to use the asynchronous approach (a more common practice in xUnit than in NUnit), two changes are likely to reduce the noise:

1. Making the `CancellationToken` optional
2. Not requiring `ConfigureAwait(false)`

JohanLarsson: Making `CancellationToken` optional or removing it makes sense. I have never had a case where I needed to test cancellation.

Not requiring `ConfigureAwait` is not going to change much as caller will likely kneejerk writing it any way.

I'm not sure it is a more common practice when using nunit, async tests are async with NUnit too.
The question is if there is a real problem with doing `task.Wait()` inside the fixture. An alternative is to expose async overloads.

JohanLarsson:
```cs
[Test]
public async Task Test()
{
    await VerifyAsync(code).ConfigureAwait(false);
}
```
There are three points where a user can get ^ wrong.
1. Not making return type `async Task`
2. Forgetting `await`
3. Forgetting `ConfigureAwait`

I'm yet to see a deadlock from doing antipattern `Wait()` in asserts. Do you have a repro that  deadlocks a test?

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/GenericAnalyzerTest.cs
```diff
@@ -0,0 +1,1191 @@
...
+
+        public List<string> DisabledDiagnostics { get; } = new List<string>();
+
+        public int? CodeFixIndex { get; set; }
+
+        public string FixedCode
+        {
+            set
+            {
+                if (this.FixedSources.Count == 1)
```
sharwell: 💭 I don't think this should allow `FixedSources` to have a value. Otherwise this becomes possible with unexpected results:

```csharp
new CSharpTest
{
  TestSources = { firstSource },
  TestCode = secondSource,
}
```

sharwell:
At this point everything is implemented under the new test helpers aside from validation of changes to additional files. This feature is required by `SettingsFileCodeFixProviderUnitTests`, which is the last test class still derived from `DiagnosticVerifier`.

sharwell:
Updated to support additional files testing (including verification of changes to additional files). All tests converted and the original test base types removed.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -648,7 +648,7 @@ public void Bar(string x, string y)
         /// Tests the behavior of SA1503 when SA1503 is suppressed for many different forms of <c>if</c> statements.
         /// </summary>
         /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
-        [Fact]
+        [Fact(Skip = "https://github.com/dotnet/roslyn-sdk/pull/191")]
```
sharwell: 🐛 Need to file a bug for this

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1412UnitTests.cs
```diff
@@ -62,15 +62,15 @@ public async Task TestFileWithUtf8EncodingWithoutBOMAsync()
             }.RunAsync(CancellationToken.None).ConfigureAwait(false);
         }
 
-        [Theory]
+        [Theory(Skip = "https://github.com/dotnet/roslyn-sdk/issues/147")]
```
sharwell: 🐛 Need to file a bug for this

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1412UnitTests.cs
```diff
@@ -62,15 +62,15 @@ public async Task TestFileWithUtf8EncodingWithoutBOMAsync()
...
         [MemberData(nameof(NonUtf8Encodings))]
         public async Task TestFixAllAsync(int codepage)
         {
             await this.TestFixAllExecuterAsync(codepage, FixAllScope.Project).ConfigureAwait(false);
             await this.TestFixAllExecuterAsync(codepage, FixAllScope.Solution).ConfigureAwait(false);
         }
 
-        [Fact]
+        [Fact(Skip = "https://github.com/dotnet/roslyn-sdk/issues/147")]
```
sharwell: 🐛 Need to file a bug for this

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/StyleCopCodeFixVerifier`2.cs
```diff
@@ -44,14 +51,158 @@ internal static Task VerifyCSharpFixAsync(string source, DiagnosticResult[] expe
...
+                        && MetadataReferences.SystemValueTupleReference != null)
+                    {
+                        solution = solution.RemoveMetadataReference(projectId, MetadataReferences.SystemValueTupleReference);
+                    }
+
+                    return solution.AddMetadataReference(projectId, GenericAnalyzerTest.CSharpSymbolsReference);
+                });
+
+                this.AdditionalFilesFactories.Add(GenerateSettingsFile);
+                this.CodeFixValidationMode = CodeFixValidationMode.None;
```
sharwell: 🐛 Need to file a bug for this

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/StyleCopCodeFixVerifier`2.cs
```diff
@@ -44,14 +51,158 @@ internal static Task VerifyCSharpFixAsync(string source, DiagnosticResult[] expe
...
+            {
+                this.OptionsTransforms.Add(options =>
+                    options
+                    .WithChangedOption(FormattingOptions.IndentationSize, this.Language, this.IndentationSize)
+                    .WithChangedOption(FormattingOptions.TabSize, this.Language, this.TabSize)
+                    .WithChangedOption(FormattingOptions.UseTabs, this.Language, this.UseTabs));
+
+                this.SolutionTransforms.Add((solution, projectId) =>
+                {
+                    if (typeof(object).Assembly.GetType("System.ValueTuple`2", throwOnError: false) != null
```
sharwell: 🐛 Need to file a bug for this

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/StyleCopCodeFixVerifier`2.cs
```diff
@@ -44,14 +51,158 @@ internal static Task VerifyCSharpFixAsync(string source, DiagnosticResult[] expe
...
+                        && MetadataReferences.SystemValueTupleReference != null)
+                    {
+                        solution = solution.RemoveMetadataReference(projectId, MetadataReferences.SystemValueTupleReference);
+                    }
+
+                    return solution.AddMetadataReference(projectId, GenericAnalyzerTest.CSharpSymbolsReference);
```
sharwell: 🐛 Need to file a bug for this

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/GenericAnalyzerTest.cs
sharwell: ➡️ Moved to https://github.com/dotnet/roslyn-sdk/issues/202

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1412UnitTests.cs
sharwell: Fixed now

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
sharwell: Fixed now

StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1412UnitTests.cs
sharwell: Fixed now

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/StyleCopCodeFixVerifier`2.cs
sharwell: Fixed now

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/StyleCopCodeFixVerifier`2.cs
sharwell: Fixed now

sharwell:
@vweijsters @pdelvo @JohanLarsson This is now ready for review. After getting a sense of the changes from looking at a few updated test classes, I recommend reviewing the more recent commits to understand the direction change. The merge commits are helpful as well, e.g. 7781c99 shows how the tests added in #2779 were updated to account for the new library.

StyleCop.Analyzers/StyleCop.Analyzers.Test/Verifiers/StyleCopCodeFixVerifier`2.cs
sharwell: Filed #2785

sharwell:
I need to fix the mistake that some test classes are marked `internal` in this pull request, leading to the tests not running at all. This is the cause of the currently-reported coverage reduction.

PR #2758, opened 2018-08-23
Allow double underscore as lambda parameter name

jnm2:
Closes #1606 using https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1443/commits/58e7be66a6765d77b48ad69f360e2986ae946db1 as prior art.

⚠Tests `SA1100UnitTests.TestIndexerAsync` and `SA1100UnitTests.TestIndexerWithInvokationAsync` were failing immediately after cloning using ReSharper's runner, before any changes were made.

PR #2760, opened 2018-08-23
Fixed ISymbol equality comparisons

jnm2:
Fixes the broken tests mentioned in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/2758.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/StandardTextDiagnosticBase.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/StandardTextDiagnosticBase.cs
@@ -167,7 +167,7 @@ namespace StyleCop.Analyzers.DocumentationRules
             }
 
             INamedTypeSymbol expectedSymbol = semanticModel.GetDeclaredSymbol(constructorDeclarationSyntax.Parent, context.CancellationToken) as INamedTypeSymbol;
-            return actualSymbol.OriginalDefinition == expectedSymbol;
+            return Equals(actualSymbol.OriginalDefinition, expectedSymbol);
         }
 
         private static bool SeeTagIsCorrect(SyntaxNodeAnalysisContext context, XElement classReferencePart, BaseMethodDeclarationSyntax constructorDeclarationSyntax)
@@ -193,7 +193,7 @@ namespace StyleCop.Analyzers.DocumentationRules
             }
 
             INamedTypeSymbol expectedSymbol = semanticModel.GetDeclaredSymbol(constructorDeclarationSyntax.Parent, context.CancellationToken) as INamedTypeSymbol;
-            return actualSymbol.OriginalDefinition == expectedSymbol;
+            return Equals(actualSymbol.OriginalDefinition, expectedSymbol);
         }
 
         private static bool TextPartsMatch(string firstText, string secondText, XmlTextSyntax firstTextPart, XmlTextSyntax secondTextPart)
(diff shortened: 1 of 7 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1100DoNotPrefixCallsWithBaseUnlessLocalImplementationExists.cs
```diff
@@ -112,7 +112,7 @@ private static void HandleBaseExpression(SyntaxNodeAnalysisContext context)
             }
 
             var speculativeSymbol = context.SemanticModel.GetSpeculativeSymbolInfo(parent.SpanStart, speculativeExpression, SpeculativeBindingOption.BindAsExpression);
-            if (speculativeSymbol.Symbol != targetSymbol.Symbol)
+            if (speculativeSymbol.Symbol.Equals(targetSymbol.Symbol))
```
sharwell: ❗️ This should have been negated

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/StandardTextDiagnosticBase.cs
```diff
@@ -193,7 +193,7 @@ private static bool SeeTagIsCorrect(SyntaxNodeAnalysisContext context, XElement
             }
 
             INamedTypeSymbol expectedSymbol = semanticModel.GetDeclaredSymbol(constructorDeclarationSyntax.Parent, context.CancellationToken) as INamedTypeSymbol;
-            return actualSymbol.OriginalDefinition == expectedSymbol;
+            return actualSymbol.OriginalDefinition.Equals(expectedSymbol);
```
jnm2: @sharwell Do you happen to know whether `OriginalDefinition` can return null?

sharwell: I am not certain

jnm2:
@sharwell Ready for review.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -117,7 +117,7 @@ private static bool HandleMethodInvocation(SemanticModel semanticModel, Anonymou
                 var invocationExpression = originalInvocationExpression.ReplaceNode(anonymousMethod, lambdaExpression);
                 SymbolInfo newSymbolInfo = semanticModel.GetSpeculativeSymbolInfo(location.SourceSpan.Start, invocationExpression, SpeculativeBindingOption.BindAsExpression);
 
-                if (originalSymbolInfo.Symbol != newSymbolInfo.Symbol)
+                if (!originalSymbolInfo.Symbol.Equals(newSymbolInfo.Symbol))
```
sharwell: ❔ Are we sure `originalSymbolInfo.Symbol` cannot be null?

jnm2: Yes, or else `GetDelegateParameterList` would have thrown NRE due to the call on line 107. If that's too distant of a guarantee, I have no problem using `Equals`.

PR #2768, opened 2018-09-03
Enable TestNoSA1503CodeFixForStatementAsync for C# 7

sharwell:
This test was previously marked private, so it would not run as part of derived test classes.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
@@ -1017,7 +1017,7 @@ public class TypeName
         /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
         [Theory]
         [MemberData(nameof(TestStatements))]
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1501UnitTests.cs
```diff
@@ -1017,7 +1017,7 @@ protected override IEnumerable<string> GetDisabledDiagnostics()
         /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
         [Theory]
         [MemberData(nameof(TestStatements))]
-        private async Task TestNoSA1503CodeFixForStatementAsync(string statementText)
+        public async Task TestNoSA1503CodeFixForStatementAsync(string statementText)
```
vweijsters: This currently triggers SA1202. This method should be placed above
 `protected override IEnumerable<DiagnosticAnalyzer> GetCSharpDiagnosticAnalyzers()`

sharwell: Fixed in e578a1a697b01981ec7bb22a97297bcbe1a79502

Issue #2777, opened 2018-09-09
Analyzer incorrectly suggests documenting methods marked as "internal" and incorrectly says to make internal fields private.


AraHaan:
The Analyzer is dumb for cases like zlib.net where it has tons of internal fields that are shared internally with it's streams to make things like the ``ZStream``, ``ZInputStream``, and ``ZOutputStream`` work as expected. However the analyzer (even with latest beta) says to make them private, and in which case I had to hack them into properties to shut it up because otherwise the project would not compile as the internal fields used in the streams would say that it is ``inaccessible due to it's protection level.``. Please fix this annoyingness I had to turn off the documentation check as it also incorrectly gripes about not having a internal method or properties seen by only the assembly itself to be documented. Not everyone want their internal methods to appear in their ``<assembly output file name>.xml`` documentation file.

sharwell:
> ... says to make internal fields private ...

This is the expected behavior of [SA1401](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1401.md).

> ... documentation check ... gripes about not having a internal [members] be documented ...

This is the default behavior, but is configurable via the [`documentInternalElements`](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#documentation-requirements) property.

AraHaan:
Ah, interesting, why is the default value for that is true, but private ones are false?

sharwell:
> Ah, interesting, why is the default value for that is true, but private ones are false?

It is based on the following abstract observations:

* Developers should not need to understand the internal implementation details of a type in order to use it correctly. The documentation of the type and its members should be sufficient.
* Implementations of a type should be simple enough that the intent/behavior of `private` elements is clear even without documentation.

sharwell:
📝 Even though SA1600 is likely not changing as a result of this bug, reviewing the tests in response to this bug report led directly to improvements for this rule, including some bug fixes related to the new C# 7.2 accessibility `private protected`. 👍

PR #2778, opened 2018-09-09
Fixed Code Analysis issues from internal fields & functions needing t…

AraHaan:
…o be private / documented.

These incorrect analysis issues can break builds of stable API's and not all API's should document methods labeled as "internal" (known to break the build of zlib.net for example and I do not like dirtying the code with suppressions making it hard to maintain). As such these analyzers need this fix instead.

See issue #2777 for more details on how it can break builds of API's.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
@@ -188,8 +188,8 @@ using System;
             await this.TestFieldDeclarationDocumentationAsync(string.Empty, false, false).ConfigureAwait(false);
             await this.TestFieldDeclarationDocumentationAsync("private", false, false).ConfigureAwait(false);
             await this.TestFieldDeclarationDocumentationAsync("protected", true, false).ConfigureAwait(false);
-            await this.TestFieldDeclarationDocumentationAsync("internal", true, false).ConfigureAwait(false);
-            await this.TestFieldDeclarationDocumentationAsync("protected internal", true, false).ConfigureAwait(false);
+            await this.TestFieldDeclarationDocumentationAsync("internal", false, false).ConfigureAwait(false);
+            await this.TestFieldDeclarationDocumentationAsync("protected internal", false, false).ConfigureAwait(false);
             await this.TestFieldDeclarationDocumentationAsync("public", true, false).ConfigureAwait(false);
 
             // Re-test with the 'documentPrivateElements' setting enabled (doesn't impact fields)
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
@@ -224,8 +224,8 @@ using System;
             await this.TestFieldDeclarationDocumentationAsync(string.Empty, true, false).ConfigureAwait(false);
             await this.TestFieldDeclarationDocumentationAsync("private", true, false).ConfigureAwait(false);
             await this.TestFieldDeclarationDocumentationAsync("protected", true, false).ConfigureAwait(false);
-            await this.TestFieldDeclarationDocumentationAsync("internal", true, false).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


AraHaan:
@sharwell it seems that loading the PDB's for the build seems to be a issue, and starts failing from there for the coverage reports.

sharwell:
> loading the PDB's for the build seems to be a issue

It's fallout from OpenCover/opencover#814 still not being merged (we are stuck using an imperfect workaround that reports a bunch of "errors" in the build window).

AraHaan:
Ok, I am working on the test that seems to fail too on the internal field test where it expects a warning too. Although currently it seems to fail build locally saying access is denied on code signing or something.

sharwell:
I see two test failures [on AppVeyor](https://ci.appveyor.com/project/sharwell/stylecopanalyzers/build/1.0.3484/tests):

* SA1401UnitTests.TestClassWithInternalFieldAsync: This is a regression test showing that we expect SA1401 to report a diagnostic on internal fields. It's not an accident that this occurs, it's the behavior as-designed for this rule.
* SA1600UnitTests.TestFieldWithoutDocumentationAsync: This is again showing that the current behavior is the expected behavior. Specifically the test for this case is on the following line:
   https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/a5a4d1759b248df9c9503773adbcee5c1c351de5/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs#L191

> Although currently it seems to fail build locally saying access is denied on code signing or something.

What's the error you are seeing?

AraHaan:
@sharwell I just also remembered while fixing the tests to edit the case of ``protected internal`` as well, and decided to collapse them to a single if instead of 3 separate if blocks.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
```diff
@@ -188,8 +188,8 @@ public async Task TestFieldWithoutDocumentationAsync()
             await this.TestFieldDeclarationDocumentationAsync(string.Empty, false, false).ConfigureAwait(false);
             await this.TestFieldDeclarationDocumentationAsync("private", false, false).ConfigureAwait(false);
             await this.TestFieldDeclarationDocumentationAsync("protected", true, false).ConfigureAwait(false);
-            await this.TestFieldDeclarationDocumentationAsync("internal", true, false).ConfigureAwait(false);
-            await this.TestFieldDeclarationDocumentationAsync("protected internal", true, false).ConfigureAwait(false);
+            await this.TestFieldDeclarationDocumentationAsync("internal", false, false).ConfigureAwait(false);
+            await this.TestFieldDeclarationDocumentationAsync("protected internal", false, false).ConfigureAwait(false);
```
sharwell: ❗️ These should definitely not change. The test is running with `documentInternalElements` set to true. These elements have internal accessibility, so with this configuration a diagnostic is expected. I added tests in #2779 (specifically commit 5712cf82fe851afe2c3e5e0a02ec6f5709861a31) to verify that the analyzer already correctly accounts for `documentInternalElements` set to false.

AraHaan:
now that I think of it should I have the change to SA1401 be a configuration option to exclude internal fields from being marked as private?

sharwell:
> ALINK : error AL1078: Error signing assembly -- Access is denied.

That's a new one for me

> now that I think of it should I have the change to SA1401 be a configuration option to exclude internal fields from being marked as private

What about simply disabling SA1401 for your project? If you need to avoid exposed fields (more visible than `internal`, you could use the Public API analyzer that helps track members that are exposed.

AraHaan:
So, changing this to not warn on internal and private is fine then?

AraHaan:
Maybe what I really got to do insread of changing SA1401 is to implement a new SA that is for the whole ``Fields must be internal or private.`` and then ye disable the SA1401.

AraHaan:
Now that I think of it, It is better in the case of internal to actually flip the fields to auto properties. I think instead I should add a analyzer for that.

AraHaan:
Seems tat when starting VS in admin it compiles. Also I tried running the tests using powershell as admin.

sharwell:
> Seems that when starting VS in admin it compiles.

I *never* run VS as admin, so this is definitely not the expected result.

sharwell:
Appears to be a machine configuration error. Details and workaround can be found in dotnet/corefx#3784.

Issue #2585, opened 2017-12-19
When will 1.1.0 be released?


tdworshak:
Hi, What are the plans to release 1.1.0?  1.1.0 Beta 4 has been out for about 6 months now.  We are seeing issues with 1.0.2 with C# 7 code (understandable), but are reluctant to upgrade to a Beta release.  Thanks!

pdufrene:
Same comment here, with C#7, we have to exclude more and more rules locally
(With #pragma disable warning...)
This becomes tricky

tdworshak:
Ok, thanks for the guidance.  We will give the Beta a try.  Cheers.

Lexcess:
Given we're another month on, I'd like to add another voice to the desire to get a release out soon (I do see that we're at beta 6 now). 

Just today we've hit another set of issues around how `SA1001 `and `SA1008 ` play with Value Tuples. If you feel the beta is good enough for users to apply to production code should these not just be actual releases? 

The current big bang approach is forcing us to disable more rules over time and just seems to hold the risk of way more issues you finally do release. Rather than a quicker cadence with a smaller set of changes (and probably fewer resultant issues).

warrenrumak:
@sharwell   We're another three months on with no progress.  A couple more blinks and it will be a full year since the last non-beta release, and you're already at 840 commits since then.

Do you need some help getting 1.1.0 out the door?  It doesn't need to be perfect.  Things can wait for 1.2.0 or a later version.

vweijsters:
Can't really comment on that, as I leave the release part to @sharwell. But any help is appreciated. Where possible I will review and merge submitted pull requests.

sharwell:
I get more than 1000 notifications per day, so it's not uncommon to miss one. We should be able to update for the 7.3 release over the next two months before moving to a stabilization period with a target release date of August to October.

Note that for analyzer packages, the pre-release indicator has no real impact, so the "stable" release is cosmetic. It primarily indicates the release has gone through the stabilization period.

Arithmomaniac:
> Note that for analyzer packages, the pre-release indicator has no real impact, so the "stable" release is cosmetic. It primarily indicates the release has gone through the stabilization period.

One impact is that another Nuget package can't reference it unless it is also in pre-release. (My company has a metapackage with some built-in runsettings to enforce code quality.)

sharwell:
> One impact is that another Nuget package can't reference it unless it is also in pre-release.

[...]

PR #2803, opened 2018-10-26
Adopt Nerdbank.GitVersioning

sharwell:
* Create a new package StyleCop.Analyzers.Unstable which avoids the use
  of prerelease version numbers
* Use SemVer 2 for the prereleases of StyleCop.Analyzers
* Avoid the need to publish builds from dedicated release branches

Implements https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2585#issuecomment-397678483


sharwell:
This is the versioning strategy used by [DocumentationAnalyzers](https://github.com/DotNetAnalyzers/DocumentationAnalyzers). It offers simplicity in publishing (shorter time between prereleases) and addresses the feature described in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2585#issuecomment-397678483, but has a downside where the published builds do not use *sequential* numbering.

PR #2809, opened 2018-10-27
Make sure the SA1130 exception also works on other invocable expressions

pdelvo:
This was found using the Tester on the monodevelop source.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
@@ -446,5 +446,242 @@ public class TypeName
 
             await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
         }
+
+        [Fact]
+        public async Task TestDelegateUseAsConstructorArgumentsAsync()
+        {
+            var testCode = @"
+using System;
+public class TypeName
+{
+    public TypeName(Action argument)
+    {
+
+    }
+
+    public void Test()
+    {
+        new TypeName(delegate { });
+    }
+}";
+            string fixedCode = @"
+using System;
+public class TypeName
+{
+    public TypeName(Action argument)
+    {
+
+    }
+
+    public void Test()
+    {
+        new TypeName(() => { });
+    }
+}";
+            var expected = Diagnostic().WithLocation(12, 22);
+            await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestDelegateUseAsConstructorArgumentsWithConflictingExpressionOverloadAsync()
+        {
+            var testCode = @"
+using System;
+using System.Linq.Expressions;
+public class TypeName
+{
+    public TypeName(Action argument)
+    {
+     
+    }
+    
+    public TypeName(Expression<Action> argument)
+    {
+    
+    }
+    
+    public void Test()
+    {
+        new TypeName(delegate { });
+    }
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -95,11 +95,11 @@ private static bool HandleMethodInvocation(SemanticModel semanticModel, Anonymou
             // invocation -> argument list -> argument -> anonymous method
             var argumentListSyntax = argumentSyntax?.Parent as ArgumentListSyntax;
 
-            var originalInvocationExpression = argumentListSyntax?.Parent as InvocationExpressionSyntax;
-            if (originalInvocationExpression != null)
+            if (argumentListSyntax != null)
             {
-                SymbolInfo originalSymbolInfo = semanticModel.GetSymbolInfo(originalInvocationExpression);
-                Location location = originalInvocationExpression.GetLocation();
+                var originalInvocableExpression = argumentListSyntax?.Parent;
```
sharwell: :bulb: `argumentListSyntax` cannot be null here

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
```diff
@@ -380,5 +380,118 @@ public static void TestMethod2(TestDelegate testDelegate)
 
             await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
         }
+
+        [Fact]
+        public async Task TestDelegateUseAsConstructorArgumentsAsync()
```
sharwell: ❓ Can we also add tests for references to overloaded indexers?

pdelvo: Good catch! This actually caused the code to crash again because BracketedArgumentListSyntax cannot be casted to ArgumentListSyntax.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -50,12 +50,25 @@ public override void Initialize(AnalysisContext context)
...
         /// <param name="argumentIndex">The index of the argument containing the delegate.</param>
         /// <returns>A parameter list for the delegate parameters.</returns>
-        internal static ParameterListSyntax GetDelegateParameterList(IMethodSymbol methodSymbol, int argumentIndex)
+        internal static ParameterListSyntax GetDelegateParameterList(ISymbol symbol, int argumentIndex)
         {
-            var delegateType = (INamedTypeSymbol)methodSymbol.Parameters[argumentIndex].Type;
+            INamedTypeSymbol delegateType;
+            if (symbol is IMethodSymbol methodSymbol)
+            {
+                delegateType = (INamedTypeSymbol)methodSymbol.Parameters[argumentIndex].Type;
```
sharwell: ❔ Is it possible for this cast to fail? For example, consider the following invalid code:

```csharp
void Method(int* data) => throw null;
void Caller() => Method(delegate { });
```

pdelvo:
@sharwell I updated the PR

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
pdelvo: I added a test, but this is not a concern because overload resolution just fails in those cases.

PR #2810, opened 2018-10-27
Fix that the SA1130 code fix crashes on an invalid event symbol

pdelvo:
This was found using the Tester on the monodevelop source.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
@@ -38,10 +38,15 @@ namespace StyleCop.Analyzers.ReadabilityRules
         }
 
         /// <inheritdoc/>
-        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
             foreach (var diagnostic in context.Diagnostics)
             {
+                if (!(await CanFixAsync(context, diagnostic).ConfigureAwait(false)))
+                {
+                    continue;
+                }
+
                 context.RegisterCodeFix(
                     CodeAction.Create(
                         ReadabilityResources.SA1130CodeFix,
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
```diff
@@ -38,19 +38,34 @@ public override FixAllProvider GetFixAllProvider()
         }
 
         /// <inheritdoc/>
-        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
             foreach (var diagnostic in context.Diagnostics)
             {
+                if (await CanFixAsync(context, diagnostic).ConfigureAwait(false))
```
sharwell: ❗️ This is either a bad method name or the expression is inverted

```suggestion
                if (!(await CanFixAsync(context, diagnostic).ConfigureAwait(false)))
```

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
```diff
@@ -221,7 +249,14 @@ private static async Task<Document> GetTransformedDocumentAsync(Document documen
 
             var anonymousMethod = (AnonymousMethodExpressionSyntax)syntaxRoot.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
 
-            var newSyntaxRoot = syntaxRoot.ReplaceNode(anonymousMethod, ReplaceWithLambda(semanticModel, anonymousMethod));
+            var newNode = ReplaceWithLambda(semanticModel, anonymousMethod);
+
+            if (newNode == null)
+            {
+                return document;
```
sharwell: ❓ Is this reachable considering the condition added before registering the code fix?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
```diff
@@ -247,7 +282,17 @@ protected override async Task<SyntaxNode> FixAllInDocumentAsync(FixAllContext fi
...
                 }
 
-                return syntaxRoot.ReplaceNodes(nodes, (originalNode, rewrittenNode) => ReplaceWithLambda(semanticModel, rewrittenNode));
+                return syntaxRoot.ReplaceNodes(nodes, (originalNode, rewrittenNode) =>
+                {
+                    var newNode = ReplaceWithLambda(semanticModel, rewrittenNode);
+
+                    if (newNode == null)
+                    {
+                        return null;
```
sharwell: ❓ Shouldn't this case be returning `rewrittenNode`?

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
pdelvo: I dont think so

pdelvo:
@sharwell I implemented your comments

PR #2811, opened 2018-10-28
Improve SA1642 to work better with single-line empty elements

### LINKED ISSUE #2696: SA1642 code fix could be improved for single-line empty elements
See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/2695#discussion_r189967266

The test currently passes, but the expected result is not what users would expect. The test should be updated to reflect the expected outcome for users and the bug corrected.

pdelvo:
Fixes #2696.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
@@ -940,7 +940,7 @@ public class TestClass
     public class ClassName
     {{
         /// <summary>
-        /// Initializes a new instance of the <see cref=""ClassName""/> class.{emptyContent}</summary>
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
```diff
@@ -940,7 +940,7 @@ public ClassName()
     public class ClassName
     {{
         /// <summary>
-        /// Initializes a new instance of the <see cref=""ClassName""/> class.{emptyContent}</summary>
+        /// Initializes a new instance of the <see cref=""ClassName""/> class.</summary>
```
sharwell: ❔  Is there a straightforward way to also make this place the `</summary>` on the following line? If this isn't resolved before merge we should file a follow-up issue for it.

PR #2835, opened 2018-12-08
Fixed issues in SA1135 for wrapped using statements

### LINKED ISSUE #2820: SA1135 reported for fully-qualified wrapped using directives
Currently SA1135 is incorrectly reported for the following code:

```csharp
using System;

namespace Test {
    using Example = System.ValueTuple<
        System.Exception,
        System.Exception>;
}
```

📝 I already wrote tests for this case, with the failing case marked as skipped, and will submit a PR for them shortly.

vweijsters:
Fixes #2820


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SymbolNameHelpers.cs
```diff
@@ -0,0 +1,81 @@
...
+
+                AppendQualifiedSymbolName(builder, namedTypeSymbol);
+
+                if (namedTypeSymbol.IsGenericType)
+                {
+                    builder.Append(GenericTypeParametersOpen);
+
+                    foreach (var typeArgument in namedTypeSymbol.TypeArguments)
+                    {
+                        if (typeArgument is INamedTypeSymbol namedTypeArgument && typeArgument.IsTupleType())
```
sharwell: 💡 I believe full coverage of this condition would need a test involving an array type as a generic type argument.

vweijsters: I'll add that 👍

PR #2853, opened 2019-01-07
Fix that SA1134 Fix All maybe non-deterministic

### LINKED ISSUE #2438: SA1134 fixer may be non-deterministic
[AppVeyor build 1.0.3046](https://ci.appveyor.com/project/sharwell/stylecopanalyzers/build/1.0.3046) failed while testing SA1134, but none of the changes in that pull request should have affected those tests.

pdelvo:
Fixes #2438.


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1134CodeFixProvider.cs
```diff
@@ -29,7 +30,7 @@ internal class SA1134CodeFixProvider : CodeFixProvider
         /// <inheritdoc/>
         public override FixAllProvider GetFixAllProvider()
         {
-            return CustomFixAllProviders.BatchFixer;
```
sharwell: ❓ Any idea what's making this not deterministic? We originally added the custom batch fixer to address non-deterministic behavior in the one provided by Roslyn.

pdelvo: No. I had the test fail exactly once in the IDE. I have not been able to reproduce it since.

sharwell (approved):

PR #2863, opened 2019-01-31
Use DiagnosticDescriptor.GetEffectiveSeverity to check for suppressions

### LINKED ISSUE #976: Use DiagnosticDescriptor.GetEffectiveSeverity to check for suppressions
As mentioned in dotnet/roslyn#1286, this method is the most straightforward way to determine if a rule is suppressed or not.

sharwell:
Closes #976

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1304NonPrivateReadonlyFieldsMustBeginWithUpperCaseLetter.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1304NonPrivateReadonlyFieldsMustBeginWithUpperCaseLetter.cs
@@ -45,6 +45,9 @@ namespace StyleCop.Analyzers.NamingRules
 
         private static readonly Action<SyntaxNodeAnalysisContext> FieldDeclarationAction = HandleFieldDeclaration;
 
+        private static readonly DiagnosticDescriptor SA1307Descriptor =
+            new SA1307AccessibleFieldsMustBeginWithUpperCaseLetter().SupportedDiagnostics.Single();
+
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1304NonPrivateReadonlyFieldsMustBeginWithUpperCaseLetter.cs
```diff
@@ -45,6 +45,9 @@ internal class SA1304NonPrivateReadonlyFieldsMustBeginWithUpperCaseLetter : Diag
 
         private static readonly Action<SyntaxNodeAnalysisContext> FieldDeclarationAction = HandleFieldDeclaration;
 
+        private static readonly DiagnosticDescriptor SA1307Descriptor =
+            new SA1307AccessibleFieldsMustBeginWithUpperCaseLetter().SupportedDiagnostics.Single();
+
```
vweijsters: ❓ Wouldn't it be a better option to make the (already static) `Descriptor` member of SA1307 `internal` and use that instead?

sharwell: It made a mess of the ordering rules. But I suppose I can do that and suppress the diagnostic it causes 😄

sharwell: @vweijsters This is now fixed 😄

PR #2871, opened 2019-02-02
Attempt to configure OpenCover for faster builds

sharwell:
This is an experiment to see if we can drop the 15 minute AppVeyor builds to something more reasonable.


sharwell:
Well it dropped 1-2 minutes off...

PR #2873, opened 2019-02-06
Update the master branch to 2.0-beta

sharwell:


StyleCop.Analyzers/version.json
```diff
@@ -1,8 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.1.1-beta.{height}",
-  // The build number offset accounts for the delay in adopting Nerdbank.GitVersioning
-  "buildNumberOffset": 40,
+  "version": "2.0-beta.{height}",
```
vweijsters: Any particular reason why you want to move to 2.0? 
If we're not going to "break" stuff, then a move to 1.2.0-beta would make more sense. If we would, for example, remove support for VS2015, then a move to 2.0 seems appropriate.

sharwell:
Closing in favor of #2875

PR #2877, opened 2019-02-07
Make CancellationToken optional from tests

kzu:
Since `CancellationToken.None` is equivalent to `default(CancellationToken)`
according to https://github.com/dotnet/corefx/issues/495#issuecomment-72322134,
and since the latter can further be shortened to just `default`, make use of
both to cleanup a bunch of method calls that all ended up using just
`CancellationToken.None`.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/DocumentationRules/SA1623CSharp7UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/DocumentationRules/SA1623CSharp7UnitTests.cs
@@ -75,7 +75,7 @@ public class TestClass
 ";
 
             var expected = Diagnostic(PropertySummaryDocumentationAnalyzer.SA1623Descriptor).WithLocation(9, 7 + accessibility.Length + type.Length).WithArguments(expectedArgument);
-            await VerifyCSharpFixAsync(testCode, expected, fixedTestCode, CancellationToken.None).ConfigureAwait(false);
+            await VerifyCSharpFixAsync(testCode, expected, fixedTestCode).ConfigureAwait(false);
         }
     }
 }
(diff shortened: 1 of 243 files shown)


sharwell:
This is currently a style matter for this repository, along with the use of `ConfigureAwait(false)`. It's come up a few times before, e.g. #518.

sharwell:
It's possible this style changes in the future, but at this time I'm not ready to apply it. *If* it changes, the cancellation token would be omitted from signatures altogether as opposed to making it optional.

kzu:
Hm... [the reason for not wanting to make it optional](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/518#issuecomment-77762486), however, has nothing to do with what I submitted here, since the optional cancellation made optional is not in any test-defining method.

kzu:
I know this is just a matter of taste, but lines were already long enough that I was scrolling just to find out that it was all just `CancellationToken.None` and `ConfigureAwait(false)`. Makes the code less readable and (at least in the former case) it's not needed at all. Not sure about the latter in the context of a unit test either.... My guess is it's not needed there either, but I may be wrong...

kzu:
In any case, getting a PR rejected in 3' after all that regex-fu, is a bit discouraging... Anyway, on to what I was really doing now... ¯\\_(ツ)_/¯

sharwell:
> In any case, getting a PR rejected in 3' after all that regex-fu, is a bit discouraging.

Yeah, I knew that was going to be the case, but I figured it was better to close it early than leave it sitting here to never go anywhere. ☹️

kzu:
What about your comment that you didn't want it [because of the issue](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/518#issuecomment-77762486) with discovering/instantiating/invoking/whatever unit tests when optional parameters where part of the signature? Which is clearly different to what this PR does? 

The https://github.com/xunit/xunit/issues/345 issue isn't implemented and even if it were, this PR doesn't prevent a specific test to provide a cancellation if needed/wanted, there's just a default in case you don't care, which is currently the case in *all* the removed code. So, it just removes all the useless boilerplate all over that's currently not adding any value... So I fail to see how it limits any such future feature.

PR #2882, opened 2019-02-13
Fix incorrect qualification of types nested within generic types

### LINKED ISSUE #2879: Code for SA1135 UsingDirectivesMustBeQualified results in unbuildable code for a nested type within a generic type
_From @davkean on February 13, 2019 6:45_

``` C# 
namespace Microsoft.VisualStudio.ProjectSystem.Designers
{
    using System;
    using System.Collections.Generic;
    using System.Collections.Immutable;
    using System.Diagnostics;
    using System.Diagnostics.Contracts;
    using System.IO;
    using System.Linq;
    using System.Threading;
    using Microsoft.VisualStudio.ProjectSystem;
    using Microsoft.VisualStudio.ProjectSystem.Properties;
 
    using NodeLookupTableBuilder = System.Collections.Immutable.ImmutableDictionary<System.IntPtr, System.Tuple<UnattachedProjectTreeNode, System.IntPtr>>.Builder;
}
```

Expected: 

``` C#
    using NodeLookupTableBuilder = System.Collections.Immutable.ImmutableDictionary<System.IntPtr, System.Tuple<Microsoft.VisualStudio.ProjectSystem.Designers.UnattachedProjectTreeNode, System.IntPtr>>.Builder;
```

Actual:

``` C#
    using NodeLookupTableBuilder = System.Collections.Immutable.Builde>;
```

_Copied from original issue: StyleCop/StyleCop#211_

sharwell:
Fixes #2879

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
@@ -3,7 +3,6 @@
 
 namespace StyleCop.Analyzers.ReadabilityRules
 {
-    using System;
     using System.Collections.Immutable;
     using System.Text;
     using Microsoft.CodeAnalysis;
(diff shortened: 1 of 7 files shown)


vweijsters:
Haven't done a proper review yet, but from what I read in the issue, the error is in the code fix, not the analyzer. I missing at least a testcase for that.

sharwell:
> Haven't done a proper review yet, but from what I read in the issue, the error is in the code fix, not the analyzer. I missing at least a testcase for that.

I will add some tests for additional cases, but the current test does cover the original issue. Even though it's fully-qualified, the original code was reporting a diagnostic and replacing the directive with garbage. 😄

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1135UsingDirectivesMustBeQualified.cs
```diff
@@ -124,37 +125,114 @@ private static void CheckUsingDeclaration(SyntaxNodeAnalysisContext context, Usi
...
+
+                    builder.Append(" ");
+                    AppendCanonicalString(builder, refType.Type);
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
+            }
```
vweijsters: ❓ Would the runtime performance of `switch (type.Kind())` be better? It would definitely improve readability, as the tuple type and ref type no longer will have to handled in the default case.

sharwell: Fixed. Also added a bunch of tests which revealed more bugs.

PR #2893, opened 2019-02-18
Document SA1107 behavior anomaly

### LINKED ISSUE #2862: SA1107 Round Tripping Fix Does Not Correct Violation
Utilizing the provided StyleCopTester.exe (this tool is very strong/awesome by the way thank you) and after putting it in CI under our code base a single file kept popping out; the resulting format resulted in no changes which may indicate an issue with SA1107's Fix.

I stripped down our utility program to the bare minimum and ran the trouble file through Delta Debugging (https://www.st.cs.uni-saarland.de/dd/) using this utility which produced this nonsensical minimal test case, any removal of any character in this file will cause it to pass.

```csharp
d({{};
```

Using the minimal test case I worked it backwards to produce a more readable scenario:

```csharp
namespace SA1107Bug
{
    class Program
    {
        //Minimal TestCase d({{};
        {
            {
                System.Console.WriteLine("Test");
            };
        }
    }
}
```

Line 10 triggers SA1107 using StyleCop.Analyzers 1.0.2 from NuGet.org but as you can see attempting to preview the fix results in no fixes available:

image

It looks like some work was done to fix some corner cases under #1543 but it is unclear if this is related.

sharwell:
Closes #2862

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1107UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1107UnitTests.cs
@@ -5,10 +5,8 @@ namespace StyleCop.Analyzers.Test.ReadabilityRules
 {
     using System.Threading;
     using System.Threading.Tasks;
-    using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.ReadabilityRules;
-    using TestHelper;
     using Xunit;
     using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
         StyleCop.Analyzers.ReadabilityRules.SA1107CodeMustNotContainMultipleStatementsOnOneLine,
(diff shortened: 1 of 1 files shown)


vweijsters:
❓ Is the linked issue number correct? The code in the test is very different from the reported issue.

sharwell:
@vweijsters Yes, the reproducer for the issue is an empty statement following a block. The contents of the block were unnecessary to reproduce the issue. The test verifies three things:

1. A diagnostic is reported (on the `;`)
2. A code fix is offered
3. The code fix does not actually make any changes to the code

vweijsters:
That part is clear to me. What I don't understand is the `[|;|]` construction in the test code. Would a single semicolon be enough?

sharwell:
Ah! The `[|text|]` is markup syntax. It works when the diagnostic analyzer only has one supported `DiagnosticDescriptor` and means:

1. A diagnostic is reported with _exactly_ that span (start and end verified)
2. The diagnostic ID matches the descriptor
3. The diagnostic severity matches the default severity of the descriptor

vweijsters:
Now its clear 😄

PR #2896, opened 2019-02-19
Fixed crash in SA1134 code fix

### LINKED ISSUE #2894: SA1134CodeFixProvider crash
While exploring the new test case of #2893 in Visual Studio (copied the test code in an empty file), selecting the SA1134 code fix resulted in a crash. (v1.1.1-rc.101)

stacktrace:
```
System.InvalidCastException : Unable to cast object of type 'Microsoft.CodeAnalysis.CSharp.Syntax.IncompleteMemberSyntax' to type 'Microsoft.CodeAnalysis.CSharp.Syntax.AttributeListSyntax'.
   at async StyleCop.Analyzers.ReadabilityRules.SA1134CodeFixProvider.GetTransformedDocumentAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetChangedSolutionAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputeOperationsAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetPreviewOperationsAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.GetPreviewResultAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.<>c__DisplayClass11_0.<GetPreviewAsync>b__0(<Unknown Parameters>)
```

vweijsters:
Fixes #2894

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1134UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1134UnitTests.cs
@@ -409,5 +409,35 @@ namespace TestNamespace
 
             await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
         }
+
+        /// <summary>
+        /// Verifies that passing an invalid member syntax into the codefix will not change the code.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        [WorkItem(2894, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2894")]
+        public async Task VerifyInvalidMemberSyntaxInCodeFixAsync()
+        {
+            string testCode = @"class Program
+{
+    static void Main(string[] args)
+    {
+        {
+        }[;]
+    }
+}
+";
+
+            DiagnosticResult[] expected =
+            {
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1134UnitTests.cs
```diff
@@ -409,5 +409,35 @@ public class TestClass<[Test(""Test1"")][Test(""Test2"")]T>
...
+        {
+        }[;]
+    }
+}
+";
+
+            DiagnosticResult[] expected =
+            {
+                DiagnosticResult.CompilerError("CS1513").WithLocation(6, 10),
+                Diagnostic().WithLocation(6, 10),
```
sharwell: ❓ Would it be better to avoid reporting a diagnostic in this case?

vweijsters: I was wondering that myself as well. On the one hand, there is enough syntax available to allow the analyzer to report a diagnostic (without crashing on a wrong syntax), but on the other hand there is a problem there when registering the code fix.

I think this is an edge case and as the code fix was easy to repair, I opted for that.

PR #2905, opened 2019-02-24
Fixed SA1130 code fix for return statements and arrow expression clauses

### LINKED ISSUE #2902: SA1130 Delegate to lambda syntax CodeFixProvider crash
Example code

```
        private static EventHandler CatchError(ToolStripItem item, Action action)
        {
            return delegate
            {
                try
                {
                    action();
                }
                catch (Exception ex)
                {
                    throw new MyException($"Failed executing {item.Text}: {ex.Message}");
                }
            };
        }
```

Stack trace

```System.NullReferenceException : Object reference not set to an instance of an object.
   at System.Collections.Immutable.ImmutableArray`1.GetEnumerator()
   at StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.GenerateUniqueParameterNames(SemanticModel semanticModel,AnonymousMethodExpressionSyntax anonymousMethod,ImmutableArray`1 argumentNames)
   at StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.ReplaceWithLambda(SemanticModel semanticModel,AnonymousMethodExpressionSyntax anonymousMethod)
   at async StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.CanFixAsync(<Unknown Parameters>)
   at async StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.RegisterCodeFixesAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeFixes.CodeFixService.GetCodeFixesAsync(<Unknown Parameters>)
```

vweijsters:
Fixes #2902

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
@@ -72,6 +72,7 @@ namespace StyleCop.Analyzers.ReadabilityRules
         {
             var parameterList = anonymousMethod.ParameterList;
             SyntaxNode lambdaExpression;
+            SyntaxToken arrowToken;
 
             if (parameterList == null)
             {
@@ -98,6 +99,11 @@ namespace StyleCop.Analyzers.ReadabilityRules
 
                     argumentList = list.Value;
                     break;
+
+                case SyntaxKind.ArrowExpressionClause:
+                case SyntaxKind.ReturnStatement:
+                    argumentList = GetMemberReturnTypeArgumentList(semanticModel, anonymousMethod);
+                    break;
                 }
 
                 List<ParameterSyntax> parameters = GenerateUniqueParameterNames(semanticModel, anonymousMethod, argumentList);
@@ -107,12 +113,17 @@ namespace StyleCop.Analyzers.ReadabilityRules
                     : SyntaxFactory.SeparatedList<ParameterSyntax>();
 
                 parameterList = SyntaxFactory.ParameterList(newList)
-                    .WithLeadingTrivia(anonymousMethod.DelegateKeyword.LeadingTrivia)
+                    .WithLeadingTrivia(anonymousMethod.DelegateKeyword.LeadingTrivia);
+
+                arrowToken = SyntaxFactory.Token(SyntaxKind.EqualsGreaterThanToken)
                     .WithTrailingTrivia(anonymousMethod.DelegateKeyword.TrailingTrivia);
             }
             else
             {
                 parameterList = parameterList.WithLeadingTrivia(anonymousMethod.DelegateKeyword.TrailingTrivia);
+
+                arrowToken = SyntaxFactory.Token(SyntaxKind.EqualsGreaterThanToken)
+                    .WithTrailingTrivia(SyntaxFactory.ElasticSpace);
             }
 
             foreach (var parameter in parameterList.Parameters)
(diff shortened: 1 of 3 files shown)


sharwell:
💡 Can you add an additional test for an event with an initializer? I don't think it's going to change anything but it seems like a test gap:

```csharp
public static event EventHandler StaticEvent = delegate { };
public event EventHandler InstanceEvent = delegate { };
```

StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
```diff
@@ -193,6 +201,14 @@ private static ImmutableArray<string> GetEqualsArgumentList(SemanticModel semant
             return namedTypeSymbol.DelegateInvokeMethod.Parameters.Select(ps => ps.Name).ToImmutableArray();
         }
 
+        private static ImmutableArray<string> GetMemberReturnTypeArgumentList(SemanticModel semanticModel, AnonymousMethodExpressionSyntax anonymousMethod)
+        {
+            var enclosingSymbol = semanticModel.GetEnclosingSymbol(anonymousMethod.Parent.SpanStart);
+            var returnType = (INamedTypeSymbol)((IMethodSymbol)enclosingSymbol).ReturnType;
```
sharwell: ❓ Will the cast to `INamedTypeSymbol` fail for invalid code like this?

```csharp
public static EventHandler[] TestMethod2() => delegate { };
```

sharwell: ❓ Will the cast to `IMethodSymbol` fail for local functions?

```csharp
void Method() {
  EventHandler LocalFunction() => delegate { };
}
```

vweijsters: This fails, I will fix it.

vweijsters: It won't.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
```diff
@@ -710,5 +710,135 @@ static void Main(string[] args)
...
+            };
+
+            var test = new CSharpTest
+            {
+                TestCode = testCode,
+                FixedCode = testCode,
+            };
+
+            test.ExpectedDiagnostics.AddRange(expected);
+            test.RemainingDiagnostics.AddRange(expected);
```
sharwell: 📝 This line should not be necessary in the latest versions of the test framework. It will automatically propagate `ExpectedDiagnostics` to `RemainingDiagnostics` since the code fix did not make any changes.

```suggestion
```

sharwell:
@vweijsters You feeling confident about this one now? 😄

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
vweijsters: Apparently the version used in the stabilization branch is too old. Your change breaks the build 😞

vweijsters: I've removed the suggested change

vweijsters:
Should be OK now, but I'm not betting anything on it 😉

PR #2912, opened 2019-03-05
Use coverlet instead of opencover

sharwell:
🚧 This will likely take a few tries to get right.


sharwell:
Will address this later.

PR #2931, opened 2019-04-09
Added default rule set for StyleCop.Analyzerz

### LINKED ISSUE #2924: Include default .ruleset file in package
The [official documentation for Analyzer rulesets](https://docs.microsoft.com/en-us/visualstudio/code-quality/analyzer-rule-sets?view=vs-2017#use-analyzer-rule-sets) says:

> Predefined rule sets are included with some NuGet analyzer packages...

> After you install a NuGet analyzer package, locate the predefined rule set in its rulesets directory, for example `%USERPROFILE%\.nuget\packages\microsoft.codequality.analyzers<version>\rulesets`. From there, you can drag and drop, or copy and paste, one or more of the rulesets to your Visual Studio project in Solution Explorer.

If you download the [fxcopanalyzers package](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/), for example, it has a ton of documented ruleset files in the `/ruleset` directory that are very helpful for getting started.

Would it be possible to package at least the default .ruleset file along with the NuGet package(s) in the manner above?

vweijsters:
Fixes #2924


sharwell:
How do we ensure this remains correct in the future?

vweijsters:
Through careful reviewing. We don't introduce that many new rules, so it shouldn't be too much work.
It may be a good idea to add a review checklist to [CONTRIBUTING.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/CONTRIBUTING.md) at some point.

Issue #2599, opened 2018-01-22
Usage of _ in methods arguments raises SA1313


Onzi12:
Usage of _ in method arguments raises [SA1313](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1313.md).

Example:
```C#
public override void Foo( int arg, int _ )
{
...
}
```

Can you please make an exception to [SA1313](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1313.md), to support '_' in method arguments?

brinko99:
I agree. As is, an exception is made for lambda expressions...

In fact, I would request that any number of ignored parameters be supported in this way. This supports the cases where multiple parameters are to be ignored.

e.g. `MyProperty.PropertyChanged += (_, __) => ...`

sharwell:
I'm not sure I agree with this one. Plus, it's going to conflict with #1949.

@brinko99 Your case is a duplicate of #1606.

Onzi12:
Why are you not sure you agree? 
The purpose of the _ is to acknowledge the arg is not in use.
Without this enhancement, users can use this feature **ONLY** in lambda handlers.

About the conflict with  [#1949](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1949) I agree. Implementing this should include an exception to  [#1949](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1949).

sharwell:
Closing this one as **won't fix**. The recommended solution for method signatures when a parameter is intentionally not used is:

1. Give the parameter a meaningful name. This should match the original definition when applicable (see #1949).
2. Assign the parameter to a discard at the start of the method to clearly indicate to the reader that the parameter is intentionally unused.

For example:

```csharp
public void Method(T value)
{
  _ = value;
}
```

tutike2000:
Why close this as won't fix if so many people want it fixed?

sharwell:
@tutike2000 There is no requirement that a project enable SA1313 or use it without modification. Options for resolving this externally include:

* SA1313 can be disabled, and a new rule implemented separately that is used instead. The new rule can be based on SA1313 and even start with its code under the MIT license.
[...]

PR #2933, opened 2019-04-10
Code cleanup

vweijsters:
I was playing around with VS2019 and noticed a lot of messages about code improvements 😄 
There is room for more improvement, but #2599 is needed for that.


vweijsters:
I'll have a look tomorrow to see if I can break things up.

PR #2936, opened 2019-04-12
Added support for CSharp8 tests and fixed #2927

### LINKED ISSUE #2927: SA1011 false positive, C#8.0 and NullableContextOptions when used with arrays.
An example:
```
byte[]? data = null; // SA1011 Closing Bracket should be followed by a space
byte[] ? data = null; // SA1018 Nullable type symbol should not be preceded by a space
```

I think SA1011 is a false positive.

vweijsters:
Fixes #2927


vweijsters:
@sharwell I created this PR to add support for CSharp8. As part of that I had to upgrade Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.XUnit. As a result of this a lot of tests with included documentation fail now. Do you have any pointers on what could have caused that?

sharwell:
I'll take a look

sharwell:
Let's try again after https://github.com/dotnet/roslyn-sdk/pull/307 is merged.

vweijsters:
Works now 👍

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1011ClosingSquareBracketsMustBeSpacedCorrectly.cs
```diff
@@ -117,7 +117,7 @@ private static void HandleCloseBracketToken(SyntaxTreeAnalysisContext context, S
                     break;
 
                 case SyntaxKind.QuestionToken:
-                    precedesSpecialCharacter = nextToken.Parent.IsKind(SyntaxKind.ConditionalAccessExpression);
+                    precedesSpecialCharacter = nextToken.Parent.IsKind(SyntaxKind.ConditionalAccessExpression) || nextToken.Parent.IsKind(SyntaxKind.NullableType);
```
sharwell: ❓ Is this needed as part of the initial introduction of the C# 8 harness, or could it be submitted as a follow-up issue to keep this PR clean?

sharwell: I would have preferred to see #2927 fixed separately due to the size of the rest of the change, but if the CI passes I won't block this PR.

sharwell:
@vweijsters I believe you'll need to update [opencover-report.ps1](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/build/opencover-report.ps1) to run the new test project in AppVeyor.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1011ClosingSquareBracketsMustBeSpacedCorrectly.cs
vweijsters: In hindsight that would have been the better option, but given that the fix for #2927 is really small I didn't want to go through the hassle of separating them.

PR #2946, opened 2019-04-29
Code cleanup attempt #2

vweijsters:
Superseeds #2933, based on review remarks there is a single commit for each type of change and no cancellationTokens have been removed in the process 😉


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1121UnitTests.cs
```diff
@@ -668,6 +668,9 @@ public class Bar {{ }}
         [MemberData(nameof(AllTypes))]
         public async Task TestNameOfAsync(string predefined, string fullName)
         {
+            // Currently unused
```
sharwell: 💡 Maybe better to say "Not needed for this test" for these.

vweijsters: done 😉

Issue #2541, opened 2017-10-13
SA1201 - Any chance on specifing a custom order of elements???


Hylndr:
Hello All,
     I was curious if there is a chance to setup a way in which the users could specify the order of various elements???  We would like to enforce a defined order of elements, however, they are different than what is currently defined...  I would like some was of designating the particular order that we have defined in our own coding standards.  Right now we have to disable SA1201 or insert the suppressing statements in order to use this rule...  The addition of the suppression statements in every class is a bit messy.

Thanks in advance,

Kevin Orcutt
Senior Software Engineer
Meyer Tool Company

sharwell:
Yes, you can configure this in stylecop.json:

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#element-order

Hylndr:
Well,
Thanks, but I don't think that will help us out any.. Let me be a bit more specific:  
Current Ordering Rules per [https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md](url)

States:
Within a class, struct, or interface, elements should be positioned in the following order:

Fields
Constructors
Finalizers (Destructors)
Delegates
Events
Enums
Interfaces
Properties
Indexers
Methods
Structs
Classes*

Within a class, we would like:
Enums
Fields
Properties
Commands
Delegates
Constructors and Finalizers
Events
Functions and Methods

...
So I don't see a way of specifying that level of ordering within the class level. Correct me if I'm wrong on that.  It would be nice to set that.

Thanks again,

Kevin

vweijsters:
That would be an extension indeed. Marking it with the appropriate tags.

postb99:
Hi,

I noticed that our usual coding style also lists Fields, then Properties, then Constructors.

Any news on discussion on this enhancement?

Thanks

PS: we use StyleCop 1.1 beta because of .net core 2.2.

JoshuaLamusga:
I think variables and properties should come before the constructor. Here's my interpretation of both reasons:

**Variable, Constructor, Property**
Variables are all private, properties and methods that follow are part of the public interface. So this is about separating private and public interface.

**Variable, Property, Constructor**
[...]

PR #2962, opened 2019-05-28
Add custom kind order for SA1201 addressing 2541, 1610

mdepalma:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1610
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2541

both asking for this customization


mdepalma:
@sharwell do you have a few minutes to take a look at this? Since I haven't contributed to this project before I'm curious what else you'd be looking for before I'd be able to complete this.

Thanks!

mdepalma:
Hi @sharwell - can you take a look at this?

sharwell:
Resolving as **wontfix** per https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2541#issuecomment-533150009.

sharwell:
Hi @mdepalma,

Considering we were never able to make the experience for this rule great even for a fixed order, I don't believe we'll be able to produce a great experience for an even more complicated configurable rule in the foreseeable future. If a customized rule is needed for your work, please do consider creating a custom analyzer just to handle this case for your project. SA1201 from this project can be disabled when you enable the custom rule.

Thanks!

PR #2976, opened 2019-07-06
SA1642 - Fix single line summary when changing the reference

### LINKED ISSUE #2963: Codefix SA1642: single summary line split into 2 lines
Hi,
I face with code fix a minor problem when single line summary is corrected. For example class declaration:

`		/// <summary>Initializes a new instance of the DatePickerTheme class.</summary>`
`		public DatePickerTheme()`

Using CodeFix the first Summart opening tag becomes on own line and rest stays on second line instead keeping as original in same line:
`		/// <summary>`
`		/// Initializes a new instance of the <see cref="DatePickerTheme"/> class.</summary>`
`		public DatePickerTheme()`

Additional question: why new StyleCop does not accept the first sample as valid since this was accepted in StyleCop Classic?

tokarzkj:
The issue resulted from always passing in a newline character

When using the code fix provider. This was resolved by determining
if the summary was multiline or not. If it isn't multiline the newLineText
is no longer passed.

Fixes #2963

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
@@ -939,8 +939,7 @@ public class TestClass
 {{
     public class ClassName
     {{
-        /// <summary>
-        /// Initializes a new instance of the <see cref=""ClassName""/> class.</summary>
+        /// <summary>Initializes a new instance of the <see cref=""ClassName""/> class.</summary>
         public ClassName()
         {{
         }}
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
```diff
@@ -939,8 +939,7 @@ public ClassName()
 {{
     public class ClassName
     {{
-        /// <summary>
-        /// Initializes a new instance of the <see cref=""ClassName""/> class.</summary>
+        /// <summary>Initializes a new instance of the <see cref=""ClassName""/> class.</summary>
```
tokarzkj: This test began to fail after I made my changes. In the issue it said this should only apply when changing the reference in a summary. However, before I implement that part I wanted to make sure I wasn't taking it too literal. It seems to me if I start with a single line comment I want it to stay on a single line. Or at the very least have each part on a separate line like below.

```
/// <summary>
/// Initializes a new instance of the <see cref=""TestClass""/> class.
/// </summary>
```

sharwell: I'm fine with this change in behavior. The test `TestConstructorEmptyDocumentationAsync` below covers the case of empty multi-line comments.

vweijsters: This testcase contains the exact issue you're trying to solve, so it would have been bad if it hadn't failed 😉

sharwell:
Thank you @tokarzkj !

Issue #2923, opened 2019-03-31
Support for nontraditional casing in namespaces (SA1300)


jedidja:
Applying StyleCop analyzers to an existing codebase is always difficult, but what happens when the company name begins with a small letter, e.g. eBay? How can you apply SA1300 to everything _except_ the `namespace`s? Imagine tens of thousands of files with several hundred namespaces like:

```
eBay.Services.Shipping
eBay.Services.Shipping.Entities
eBay.Services.Auction
eBay.Services.Auction.Bidding
```

Although normally namespace names are `PascalCase`, Microsoft admits in their [naming guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces) that 

> If your brand employs nontraditional casing, you should follow the casing defined by your brand, even if it deviates from normal namespace casing.

If someone were to do the work and create a PR, would it be acceptable to either:

a) include a "known terms dictionary" that SA1300 could use to ignore certain words?
b) make a *very* specific exception for namespaces since company names do not always follow `PascalCase`?

Or, if neither of those would be accepted, is there any way to accomodate this situation?

vweijsters:
The preferable option for me would be a combination of both listed options: Add a configuration option to ignore certain words for namespaces only. The ignored words will then have to be matched in the exact case that they are specified in, to guard the consistency as much as possible.

jedidja:
@vweijsters Thanks for the response; your idea makes sense to me as I think it still keeps the underlying spirit of the StyleCop guideline alive while allowing for real-world issues. 

I took a look at the [contributing guidelines](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/CONTRIBUTING.md) but I'm still unsure if it's worth spending any time on creating a PR for this yet, or if I should wait for feedback from @sharwell or someone else first? It seems like perhaps I would at least wait for a specific label to be applied and then go from there? (Maybe I'm not even the right person to create the PR heh)

sharwell:
@jedida can you add a proposal for the exact change to stylecop.json for this?

jedidja:
Sure; here's a first stab at it :)

**Property**: `allowedNamespaceComponentTerms`

**Default Value**: `[]`

[...]

PR #2996, opened 2019-08-27
Support for nontraditional casing in namespaces (SA1300) #2923

jedidja:
- Tried to mimic how `allowedHungarianPrefixes` works as this feature seems similar.
- I took an initial stab at updating the documentation, but assume there are people who have a much better idea how to phrase this addition :)
- I noticed some of the documentation still refers to screenshots from the app; does that require updating as well?


StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1300UnitTests.cs
```diff
@@ -44,6 +44,31 @@ public async Task TestLowerCaseNamespaceAsync()
...
         }
 
+        [Fact]
+        public async Task TestAllowedLowerCaseNamespaceIsNotReportedAsync()
+        {
+            var customTestSettings = @"
+{
+  ""settings"": {
+    ""namingRules"": {
+      ""allowedNamespaceComponentTerms"": [ ""eBay"" ]
```
sharwell: ❓ Would it be better for this to be `allowedNamespaces`? For the `iPod` example below, the configuration would be `Apple.iPod`, which would then allow `Apple.iPod` itself along with any namespace under it, but would not allow other top-level namespaces like `Android.iPod`.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1300ElementMustBeginWithUpperCaseLetter.cs
```diff
@@ -29,6 +30,9 @@ namespace StyleCop.Analyzers.NamingRules
     /// class. A <c>NativeMethods</c> class is any class which contains a name ending in <c>NativeMethods</c>, and is
     /// intended as a placeholder for Win32 or COM wrappers. StyleCop will ignore this violation if the item is placed
     /// within a <c>NativeMethods</c> class.</para>
+    ///
+    /// <para>For namespace components that begin with a small letter, due to branding issues or other reasons, add the
+    /// term to the <c>allowedNamespaceComponentTerms</c> list.</para>
```
sharwell:
```suggestion
    /// term to the <see href="https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/ed1159193cd1642ee149674eb8c19f4bffa27670/documentation/Configuration.md#namespace-components"><c>allowedNamespaceComponentTerms</c> list</see>.</para>
```

sharwell: 📝 These hyperlinks and styles actually work in the IDE now, aside from the fact that `<remarks>` are currently hidden. But they _will_ work eventually 😆

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1300ElementMustBeginWithUpperCaseLetter.cs
```diff
@@ -82,13 +86,13 @@ public override void Initialize(AnalysisContext context)
...
-        private static void HandleNamespaceDeclaration(SyntaxNodeAnalysisContext context)
+        private static void HandleNamespaceDeclaration(SyntaxNodeAnalysisContext context, StyleCopSettings settings)
         {
             NameSyntax nameSyntax = ((NamespaceDeclarationSyntax)context.Node).Name;
-            CheckNameSyntax(context, nameSyntax);
+            CheckNameSyntax(context, nameSyntax, settings);
         }
 
-        private static void CheckNameSyntax(SyntaxNodeAnalysisContext context, NameSyntax nameSyntax)
+        private static void CheckNameSyntax(SyntaxNodeAnalysisContext context, NameSyntax nameSyntax, StyleCopSettings settings)
```
sharwell: 💡 Consider renaming this to `CheckNamespaceNameSyntax` to avoid confusion with other checks in this class.

documentation/Configuration.md
```diff
@@ -314,6 +314,30 @@ The following example shows a settings file which allows the common prefixes as
 }
 ```
 
+### Namespace Components
+
+The following property is used to configure allowable namespace component terms (e.g. ones that start with a lowercase letter).
+
+| Property | Default Value | Minimum Version | Summary |
+| --- | --- | --- | --- |
+| `allowedNamespaceComponentTerms` | `[ ]` | 1.2.0 | Specifies namespace component terms that are allowed to be used. See the example below for more information. |
```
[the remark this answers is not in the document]
sharwell: 1.2.0 is correct. After it ships we have the option of updating this to have the exact beta release it first appeared in.

StyleCop.Analyzers/StyleCop.Analyzers/NamingRules/SA1300ElementMustBeginWithUpperCaseLetter.cs
jedidja: Renamed to `CheckNamespaceNameSyntax `

StyleCop.Analyzers/StyleCop.Analyzers.Test/NamingRules/SA1300UnitTests.cs
jedidja: I was hoping to configure namespace components, so that `Apple.iPod` and `Android.iPod` would both be acceptable.

jedidja:
Hmm. I'm not sure what's going on with the build - the tests seem to pass locally.

image

PR #3020, opened 2019-09-16
Fix issue in SA1130 code fix with parameterless delegates.

### LINKED ISSUE #2997: NullReferenceException on SA1130 CodeFix
I was trying to write a code but it was suggested (Resharper I assume) to change by adding delegate. Then code analyzer picked up as stylecop violation and suggested to fix. As result the SA1130 code crashes with NullReferenceException.

The incomplete C# code on which code violation picked but fix is crashing:

```
		private Dictionary<string, Func<IEventActionPackageFactory>> mVirtualItemFactory =
		{
			{ Event.Types.COLLECTION_EVENT, delegate { return new EventActionPackageFactoryForEvent(mContext); } },
			{ Event.Types.ALARM, () => new EventActionPackageFactoryForAlarm(mContext) },
		};
```

Error stack trace:

```
System.NullReferenceException : Object reference not set to an instance of an object.
   at System.Collections.Immutable.ImmutableArray`1.GetEnumerator()
   at StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.GenerateUniqueParameterNames(SemanticModel semanticModel,AnonymousMethodExpressionSyntax anonymousMethod,ImmutableArray`1 argumentNames)
   at StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.ReplaceWithLambda(SemanticModel semanticModel,AnonymousMethodExpressionSyntax anonymousMethod)
   at async StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.CanFixAsync(<Unknown Parameters>)
   at async StyleCop.Analyzers.ReadabilityRules.SA1130CodeFixProvider.RegisterCodeFixesAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeFixes.CodeFixService.GetCodeFixesAsync(<Unknown Parameters>)
```

vweijsters:
Fixes #2997

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
@@ -76,7 +76,7 @@ namespace StyleCop.Analyzers.ReadabilityRules
 
             if (parameterList == null)
             {
-                ImmutableArray<string> argumentList = default;
+                ImmutableArray<string> argumentList = ImmutableArray<string>.Empty;
 
                 switch (anonymousMethod.Parent.Kind())
                 {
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs
```diff
@@ -216,19 +216,22 @@ private static List<ParameterSyntax> GenerateUniqueParameterNames(SemanticModel
         {
             var parameters = new List<ParameterSyntax>();
 
-            foreach (var argumentName in argumentNames)
+            if (!argumentNames.IsDefault)
```
sharwell: :question: would it be cleaner to change the initial value to `Empty` here?

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/53df141371b0e4767233dc4accd3a92b1199f144/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1130CodeFixProvider.cs#L79

vweijsters: That is better yes. My initial thoughts were that a difference between a non-existing parameter list and an empty parameter list was necessary, but that is not the case.

PR #3021, opened 2019-09-16
Changed default settings for SA1316 to PascalCase. 

### LINKED ISSUE #3019: The default tuple element names config should align with the new pascal case guidance
The BCL and language teams have decided on pascal case for tuple element names and have added an API to the BCL which sets the precedent. (https://github.com/dotnet/corefx/pull/35595, 
https://github.com/dotnet/corefx/issues/33553)

Under https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1316.md#rule-description:

> For example, the following code would produce a violation of this rule (with default settings):
> 
> ```csharp
> public class ExampleClass
> {
>   public (int ValueA, int ValueB) ExampleMethod() // SA1316
>   {
>   }
> }
> ```

This should not default to being a violation. I would even expect that camel case would be considered a violation as soon as this rule is enabled in a new project.

This rule does not exist in the latest release version, so maybe backwards compatibility won't be an issue.

vweijsters:
Added missing C# 8 tests for SA1316.

Fixes #3019

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/NamingRules/SA1316CSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/NamingRules/SA1316CSharp8UnitTests.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.Test.CSharp8.NamingRules
+{
+    using StyleCop.Analyzers.Test.CSharp7.NamingRules;
+
+    public class SA1316CSharp8UnitTests : SA1316CSharp7UnitTests
+    {
+    }
+}
--- a/documentation/Configuration.md
+++ b/documentation/Configuration.md
@@ -323,7 +323,7 @@ The following properties are used to configure the behavior of the tuple element
 | `includeInferredTupleElementNames` | false | 1.2.0 | Specifies whether inferred tuple element names will be analyzed as well. |
 | `tupleElementNameCasing` | "PascalCase" | 1.2.0 | Specifies the casing convention used for tuple element names. |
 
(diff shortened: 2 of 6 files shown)


sharwell:
Pending a resolution of discussion in #3019

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/NamingRules/SA1316CSharp8UnitTests.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.Test.CSharp8.NamingRules
+{
+    using StyleCop.Analyzers.Test.CSharp7.NamingRules;
+
+    public class SA1316CSharp8UnitTests : SA1316CSharp7UnitTests
```
sharwell: Was this just missed by accident?

documentation/Configuration.md
```diff
@@ -330,7 +330,7 @@ The following example shows a settings file which requires tuple element names t
   "settings": {
     "namingRules": {
       "includeInferredTupleElementNames": true,
-      "tupleElementNameCasing" : "PascalCase"
+      "tupleElementNameCasing" : "camelCase"
```
sharwell: :bulb: make sure the preceding text is consistent with this change (I can't see the full text on this small screen)

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/NamingRules/SA1316CSharp8UnitTests.cs
vweijsters: Yes, as SA1316 doesn't have any tests for CSharp 6, I probably missed it when adding the C# 8 tests.

documentation/Configuration.md
vweijsters: Fixed it

PR #3033, opened 2019-09-27
Add test cases for tuple deconstruction

### LINKED ISSUE #3031: SA1312 is raised for deconstructed tuple elements when SA1316 is set to PascalCase
The following code raises SA1312 for the tuple element names

```
var properties = new List<(string Name, string Value)>();
foreach ((string Name, string Value) in properties) // SA1312 Raised
{
}
```

But if i change the names to camelCase SA1316 is raised:
```
var properties = new List<(string Name, string Value)>();
foreach ((string name, string value) in properties) // SA1316 Raised
{
}
```

I'm not sure which one is right 🤔

sharwell:
I was not able to find a failing scenario. Perhaps this was already fixed by a prior change?

Closes #3031

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/NamingRules/SA1316CSharp7UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/NamingRules/SA1316CSharp7UnitTests.cs
@@ -8,6 +8,7 @@ namespace StyleCop.Analyzers.Test.CSharp7.NamingRules
     using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.Lightup;
     using StyleCop.Analyzers.NamingRules;
+    using StyleCop.Analyzers.Settings.ObjectModel;
     using Xunit;
     using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
         StyleCop.Analyzers.NamingRules.SA1316TupleElementNamesShouldUseCorrectCasing,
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/NamingRules/SA1316CSharp7UnitTests.cs
```diff
@@ -291,5 +292,83 @@ public void TestMethod2()
...
+    }
+}
+";
+            var settings = $@"{{
+  ""settings"": {{
+    ""namingRules"": {{
+      ""tupleElementNameCasing"": ""{TupleElementNameCase.CamelCase}""
+    }}
+  }}
+}}
```
vweijsters: The CamelCaseTestSettings constant can be used here.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/NamingRules/SA1316CSharp7UnitTests.cs
```diff
@@ -291,5 +292,83 @@ public void TestMethod2()
...
+}
+";
+            var settings = $@"{{
+  ""settings"": {{
+    ""namingRules"": {{
+      ""tupleElementNameCasing"": ""{TupleElementNameCase.PascalCase}""
+    }}
+  }}
+}}
+";
```
vweijsters: The PascalCaseTestSettings const can be used here

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/NamingRules/SA1316CSharp7UnitTests.cs
```diff
@@ -291,5 +292,83 @@ public void TestMethod2()
...
```
vweijsters: The PascalCaseTestSettings const can be used here

vweijsters:
I can't reproduce the issue either. I've tested with both VS2017 and VS2019.

PR #3049, opened 2019-10-28
Add support for SwitchExpressionArmSyntax

costin-zaharia-sonarsource:


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Lightup/SwitchExpressionArmSyntaxWrapperTests.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace StyleCop.Analyzers.Test.CSharp8.Lightup
+{
+    using System;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using StyleCop.Analyzers.Lightup;
+    using Xunit;
+
+    public class SwitchExpressionArmSyntaxWrapperTests
```
sharwell: 💡 This also needs tests for C# 6 and C# 7. Since these two will behave the same way, you can implement this test once (in the C# 6 tests), and then create an empty derived type in the C# 7 test project:

```csharp
public class CSharp7SwitchExpressionArmSyntaxWrapperTests : SwitchExpressionArmSyntaxWrapperTests
{
}
```

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Lightup/SwitchExpressionArmSyntaxWrapperTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            Assert.True(syntaxNode.IsKind(SyntaxKindEx.SwitchExpressionArm));
+
+            var wrapper = (SwitchExpressionArmSyntaxWrapper)syntaxNode;
+            Assert.Same(syntaxNode, wrapper.SyntaxNode);
+            Assert.Same(syntaxNode.Pattern, wrapper.Pattern.SyntaxNode);
+            Assert.Same(syntaxNode.WhenClause, wrapper.WhenClause.SyntaxNode);
+            Assert.Same(syntaxNode.Expression, wrapper.Expression);
+            Assert.Equal(syntaxNode.EqualsGreaterThanToken, wrapper.EqualsGreaterThanToken);
+
+            // Pattern
```
sharwell: 💡 Can you break these into a test for each property?

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Lightup/SwitchExpressionArmSyntaxWrapperTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            var newPattern = SyntaxFactory.ConstantPattern(SyntaxFactory.LiteralExpression(SyntaxKind.NullLiteralExpression));
+            var wrapperWithModifiedPattern = wrapper.WithPattern((PatternSyntaxWrapper)newPattern);
+            Assert.NotNull(wrapperWithModifiedPattern.SyntaxNode);
+            Assert.NotSame(syntaxNode.Pattern, wrapperWithModifiedPattern.Pattern.SyntaxNode);
+            Assert.Equal(SyntaxKind.ConstantPattern, wrapperWithModifiedPattern.Pattern.SyntaxNode.Kind());
```
sharwell: 💡 You should be able to also assert that the new expression arm has the correct pattern. I don't remember if `Assert.Same` would work here, but I believe there is least an `IsEquivalentTo` for syntax nodes that could be used to compare `wrapperWithModifiedPattern.Pattern` and `newPattern`.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Lightup/SwitchExpressionArmSyntaxWrapperTests.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            // Expression
+            var newExpression = SyntaxFactory.LiteralExpression(SyntaxKind.NullLiteralExpression);
+            var wrapperWithModifiedExpression = wrapper.WithExpression(newExpression);
+            Assert.NotNull(wrapperWithModifiedExpression.SyntaxNode);
+            Assert.NotSame(syntaxNode.Expression, wrapperWithModifiedExpression.Expression);
+            Assert.Equal(SyntaxKind.NullLiteralExpression, wrapperWithModifiedExpression.Expression.Kind());
+
+            // EqualsGreaterThanToken
+            Assert.Throws<ArgumentException>(() => wrapper.WithEqualsGreaterThanToken(SyntaxFactory.Token(SyntaxKind.EqualsEqualsToken)));
```
sharwell: ❔ Is it possible to test this with a passing case?

costin-zaharia-sonarsource: Since this will throw if the token is `!= SyntaxKind.EqualsGreaterThanToken` the only solution I see is to pass this as a parameter. It's not a perfect test since it has already the same value but at least we can see that it doesn't throw.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Lightup/SwitchExpressionArmSyntaxWrapperTests.cs
costin-zaharia-sonarsource: I think I cannot move all the tests on the C# 6 project because the `SyntaxFactory.ConstantPattern` and `SyntaxFactory.WhenClause` are available only starting with C# 7 and `SyntaxFactory.SwitchExpressionArm` with C# 8.

What I did in the end is to split the tests according with their availability.

sharwell: Here's an example of split tests:

* C# 6 (syntax not available): https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/c69ba488c8d23e23bf76772073ccb5cbacc7e319/StyleCop.Analyzers/StyleCop.Analyzers.Test/Lightup/DeclarationPatternSyntaxWrapperTests.cs
* C# 7 (syntax available): https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/c69ba488c8d23e23bf76772073ccb5cbacc7e319/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/Lightup/DeclarationPatternSyntaxWrapperTests.cs

sharwell: Ah, you should be good with the split approach here. It works for cases where the syntax is completely absent in the earlier releases, but not for cases where the syntax was just *changed*.

costin-zaharia-sonarsource: Yes, I think the approach I've used is similar. I've tried to avoid duplication where possible,

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Lightup/SwitchExpressionArmSyntaxWrapperTests.cs
sharwell: You can create a token with this kind, but set the leading or trailing trivia to something new and test that. For example, `/*1*/=>` would be semantically the same as `=>` but is still different enough to observe in a test. 👍

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/CSharp8.md
```diff
@@ -570,16 +570,16 @@ See [Microsoft.CodeAnalysis.CSharp release/dev16.3@c955f3c99b5698c906e0700ef691b
...
-* [ ] `Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax.WithWhenClause(Microsoft.CodeAnalysis.CSharp.Syntax.WhenClauseSyntax whenClause) -> Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax`
+* [x] `Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax`
+* [x] `Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax.EqualsGreaterThanToken.get -> Microsoft.CodeAnalysis.SyntaxToken`
+* [x] `Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax.Expression.get -> Microsoft.CodeAnalysis.CSharp.Syntax.ExpressionSyntax`
+* [x] `Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax.Pattern.get -> Microsoft.CodeAnalysis.CSharp.Syntax.PatternSyntax`
+* [x] ~~`Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax.Update(Microsoft.CodeAnalysis.CSharp.Syntax.PatternSyntax pattern, Microsoft.CodeAnalysis.CSharp.Syntax.WhenClauseSyntax whenClause, Microsoft.CodeAnalysis.SyntaxToken equalsGreaterThanToken, Microsoft.CodeAnalysis.CSharp.Syntax.ExpressionSyntax expression) -> Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax`
```
sharwell: :bulb: This needs the trailing `~~` as well for the strikethrough effect.

```suggestion
* [x] ~~`Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax.Update(Microsoft.CodeAnalysis.CSharp.Syntax.PatternSyntax pattern, Microsoft.CodeAnalysis.CSharp.Syntax.WhenClauseSyntax whenClause, Microsoft.CodeAnalysis.SyntaxToken equalsGreaterThanToken, Microsoft.CodeAnalysis.CSharp.Syntax.ExpressionSyntax expression) -> Microsoft.CodeAnalysis.CSharp.Syntax.SwitchExpressionArmSyntax`~~
```

sharwell:
❤️ Thanks for implementing this. I know I've been a bit lazy with some of the lightup layer tests but this is quite solid.

costin-zaharia-sonarsource:
Hi @sharwell! Thanks a lot for the review!

PR #3060, opened 2019-11-06
Do not report SA1141 for single-element tuples

### LINKED ISSUE #3055: SA1141 mis-fires for ValueTuple`1 use
Given this:

```cs
ValueTuple<int> ct = ValueTuple.Create(1);
```

SA1141 mis-fires and claims that `ValueTuple.Create(1)` can be replaced with `(1)` and in fact produces a code-fix to that effect. However it is not equivalent. In this case it will produce a compiler error because C# only lets you use the `(x,y,z)` value type syntax when you have *more* than one element in the value tuple.

Even worse than the compiler error is when the result of the expression is assigned to `object`, at which point instead of a compiler error, the type of the expression silently changes from `ValueTuple<int>` to a boxed int as `object`.

sharwell:
Fixes #3055

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1141UseTupleSyntax.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1141UseTupleSyntax.cs
@@ -121,6 +121,12 @@ namespace StyleCop.Analyzers.ReadabilityRules
             }
 
             var invocationExpression = (InvocationExpressionSyntax)context.Node;
+            if (invocationExpression.ArgumentList.Arguments.Count < 2)
+            {
+                // Tuple creation with less than two elements cannot use the language syntax
+                return;
+            }
+
             if (!invocationExpression.Expression.IsKind(SyntaxKind.SimpleMemberAccessExpression))
             {
                 return;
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1141UseTupleSyntax.cs
```diff
@@ -227,7 +233,9 @@ private static void CheckGenericName(SyntaxNodeAnalysisContext context, GenericN
         private static bool IsValueTuple(SyntaxNodeAnalysisContext context, ExpressionSyntax syntax)
         {
             var symbolInfo = context.SemanticModel.GetSymbolInfo(syntax, context.CancellationToken);
-            return (symbolInfo.Symbol is ITypeSymbol typeSymbol) && typeSymbol.IsTupleType();
+            return symbolInfo.Symbol is INamedTypeSymbol typeSymbol
+                && typeSymbol.IsTupleType()
+                && typeSymbol.TupleElements().Length > 1;
```
vweijsters: I would prefer that you rename the function to indicate that it is no longer considering tuples with 1 element.

sharwell: Now fixed

PR #3061, opened 2019-11-06
Fix spacing after closing brace in patterns

### LINKED ISSUE #3053: SA1013 should not warn when pattern case
```cs
private static void M(InvocationExpressionSyntax invocation)
{
    switch (invocation)
    {
        case { Expression: MemberAccessExpressionSyntax { Expression: { } expression } }:
            break;
    }
}
```

image

sharwell:
Fixes #3053

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
@@ -3,9 +3,64 @@
 
 namespace StyleCop.Analyzers.Test.CSharp8.SpacingRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
     using StyleCop.Analyzers.Test.CSharp7.SpacingRules;
+    using Xunit;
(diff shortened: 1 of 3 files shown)


JohanLarsson:
Nice!

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
```diff
@@ -3,9 +3,54 @@
...
+        {
+        case Exception { Message: { } message }:
+            break;
+        }
+    }
+}";
+
+            var expected = Diagnostic().WithSpan(9, 37, 9, 38).WithArguments(string.Empty, "followed");
+            await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
+        }
```
vweijsters: ❓ Given that it is the intention that the space before the colon is catched by SA1024, it may be good to add a testcase verifying that it isn't caught by SA1013.

sharwell: Now fixed

PR #3071, opened 2019-11-17
Make SA1011 work with exclamation (`!`)

### LINKED ISSUE #3052: SA1011 false positive, C#8.0 when using ! forgiveness operator
```cs
IReadOnlyList<object?>? arguments; // a parameter or assigned elsewhere
object o = arguments[0]!; // SA1011 does not like the `!` so close to the `]` character
```

Similar to #2927
Repros with 1.2.0-beta.66

pantosha:
Fixes #3052

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1011CSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1011CSharp8UnitTests.cs
@@ -67,28 +67,12 @@ namespace StyleCop.Analyzers.Test.CSharp8.SpacingRules
     {
         public void TestMethod(object?[] arguments)
         {
-            object o = arguments[0]!;
-            string s = arguments[0]!.ToString();
-        }
-    }
-}
-";
-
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1011CSharp8UnitTests.cs
```diff
@@ -52,6 +52,26 @@ public class TestClass
...
+        public async Task TestClosingSquareBracketFollowedByExclamationAsync()
+        {
+            var testCode = @"namespace TestNamespace
+{
+    public class TestClass
+    {
+        public void TestMethod(object?[] arguments)
+        {
+            object o = arguments[0]!;
+            string s = arguments[0]!.ToString();
```
sharwell: :bulb: can you add a test showing that a space here would get removed? A space after the `!` would likely be reported on the `.` or `;` token since those have dedicated diagnostic IDs, but a space before might be part of this analyzer. The tests should ideally be added to all the impacted diagnostics.

pantosha: Yes, I updated PR.

sharwell:
@pantosha I updated this to expand testing of this issue to include SA1002 and SA1019.

PR #3072, opened 2019-11-17
Initial Azure Pipelines build configuration

sharwell:
This pull request moves the testing portion of PRs to Azure Pipelines for improved performance. AppVeyor builds are still active since Azure Pipelines does not provide the artifact support required by our status page.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,6 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
     <add key="roslyn-analyzers" value="https://dotnet.myget.org/F/roslyn-analyzers/api/v3/index.json" />
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -47,6 +47,12 @@ jobs:
       filePath: build/opencover-report.ps1
       arguments: '$(_debugArg) -NoBuild -NoReport -Azure'
 
+  - task: PublishTestResults@2
+    displayName: Publish test results
+    inputs:
+      testResultsFormat: xUnit
+      testResultsFiles: 'build/*.xml'
+
   - task: PublishBuildArtifacts@1
     displayName: Publish build logs
     inputs:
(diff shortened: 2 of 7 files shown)


NuGet.config
```diff
@@ -1,6 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
jnm2: Why might I want to consider doing this?

azure-pipelines.yml
```diff
@@ -0,0 +1,74 @@
...
+  demands:
+  - msbuild
+  - visualstudio
+  - vstest
+
+jobs:
+- job: Build
+  variables:
+    BuildSolution: StyleCopAnalyzers.sln
+    BuildPlatform: Any CPU
```
jnm2: I usually remove this variable.

azure-pipelines.yml
```diff
@@ -0,0 +1,74 @@
...
+    displayName: 'Use NuGet 5.3.1'
+    inputs:
+      versionSpec: 5.3.1
+
+  - task: NuGetCommand@2
+    displayName: 'NuGet restore'
+    inputs:
+      restoreSolution: '$(BuildSolution)'
+      feedsToUse: 'config'
+      nugetConfigPath: 'NuGet.config'
```
jnm2: Vastly prefer `msbuildArgs: '/restore'`

azure-pipelines.yml
```diff
@@ -0,0 +1,74 @@
...
+
+  - task: VSBuild@1
+    displayName: 'Build solution StyleCopAnalyzers.sln'
+    inputs:
+      solution: '$(BuildSolution)'
```
jnm2: I usually remove this variable too and either remove `solution` or do `solution: .`.

azure-pipelines.yml
sharwell: We have to restore packages from a packages.config file.

NuGet.config
sharwell: It's in the commit message https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3072/commits/a5cbd747cdc7e656f6d1815bbdcb7098b1e0123a

azure-pipelines.yml
jnm2: Whoa, why is that?

NuGet.config
jnm2: Perfect, looked there but didn't think of looking for a multiline message. Thanks!

azure-pipelines.yml
sharwell: It's how we bring in the test runner and code coverage tools today.

PR #3105, opened 2020-01-11
Simplify names

NextTurn:
IDE0001 cleanups (namespaces and `var`).

Why don't StyleCop have these rules?


NextTurn:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/b15704eec6f407e39a32e78912ded87ea4c0c060/.editorconfig#L75

The editor is configured to prefer `var` only "where type is obvious". In such case I just set it to `false` to avoid `var` changes.

PR #3111, opened 2020-01-26
Remove an unused local variable

NextTurn:


sharwell:
💭 Was it supposed to be used?

NextTurn:
>  Was it supposed to be used?

It was a mistake. This value is not actually needed.

sharwell:
Thanks!

PR #3120, opened 2020-03-02
Switch to MIT license

sharwell:


NextTurn:
@sharwell GitHub failed to recognize the new license.

image

Issue #3094, opened 2019-12-19
Documentation rules should not be checked if an <exclude/> tag is present


RobSiklos:
[Sandcastle](https://ewsoftware.github.io/XMLCommentsGuide/html/a8ade450-e201-401a-923d-1a1169ef7828.htm) (and I believe NDoc) support the `<exclude/>` documentation element, which makes it so that the documentation for that member will be excluded from the generated documentation.

One could argue that if this tag is present, StyleCop should not report any documentation rule violations for the corresponding member.

Note: this WAS fixed in legacy StyleCop (CodePlex #7732) on 2016-10-21.

PR #3121, opened 2020-03-03
Fix for reported Issue 3094

CPAM34:
I have a fix for issue #3094 as reported in the main repository, concerning the exclusion of rule enforcement for Doc rules when `<exclude/>` is present. I have created accompanying new unit tests in the StyleCop.Analyzers.Test project, all of which have passed.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1604UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1604UnitTests.cs
@@ -68,6 +68,21 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
         [InlineData("class")]
         [InlineData("struct")]
         [InlineData("interface")]
+        public async Task TestTypeWithExcludedDocumentationAsync(string typeName)
+        {
+            var testCode = @"
+/// <exclude/>
+{0} TypeName
+{{
+}}";
+            await VerifyCSharpDiagnosticAsync(string.Format(testCode, typeName), DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [InlineData("enum")]
+        [InlineData("class")]
+        [InlineData("struct")]
+        [InlineData("interface")]
         public async Task TestTypeWithoutDocumentationAsync(string typeName)
         {
             var testCode = @"
@@ -130,6 +145,16 @@ void TypeName();";
         }
 
         [Fact]
+        public async Task TestDelegateWithExcludedDocumentationAsync()
+        {
+            var testCode = @"
+/// <exclude/>
+public delegate
+void TypeName();";
+            await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
         public async Task TestDelegateWithoutDocumentationAsync()
         {
             var testCode = @"
@@ -183,6 +208,21 @@ public class ClassName
 public class ClassName
 {
     /// <inheritdoc/>
+    public void Test() { }
+}";
+            await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestMethodWithExcludedDocumentationAsync()
+        {
+            var testCode = @"
+/// <summary>
+/// 
+/// </summary>
+public class ClassName
+{
+    /// <exclude/>
     public void Test() { }
 }";
             await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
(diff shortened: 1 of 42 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1605PartialElementDocumentationMustHaveSummary.cs
```diff
@@ -95,6 +95,13 @@ protected override void HandleXmlElement(SyntaxNodeAnalysisContext context, bool
 
             if (completeDocumentation != null)
             {
+                // This documentation rule is excluded via the <exclude /> tag
```
RobSiklos: What's up with this commented code?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609PropertyDocumentationMustHaveValue.cs
```diff
@@ -58,6 +58,12 @@ protected override void HandleXmlElement(SyntaxNodeAnalysisContext context, bool
 
             if (completeDocumentation != null)
             {
+                // This documentation rule is excluded via the <exclude /> tag
+                /*if (completeDocumentation.Nodes().OfType<XElement>().Any(element => element.Name == XmlCommentHelper.ExcludeXmlTag))
+                {
+                    return;
+                }*/
+
```
RobSiklos: More commented code

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/FileHeaderAnalyzers.cs
```diff
@@ -203,16 +203,26 @@ public static void HandleSyntaxTree(SyntaxTreeAnalysisContext context, StyleCopS
...
-                    if (fileHeader.IsMalformed)
+                    if (fileHeader != null)
                     {
-                        context.ReportDiagnostic(Diagnostic.Create(SA1633DescriptorMalformed, fileHeader.GetLocation(context.Tree)));
-                        return;
+                        // Don't process documentation rules when an <exclude /> tag is present
+                        /*if (fileHeader.GetElement("exclude") != null)
+                        {
+                            return;
+                        }*/
```
RobSiklos: Why commented?

RobSiklos: Why commented? Anyway, you don't need to modify this code. The "file header" is not true XML documentation, and can't have an <exclude/> tag. See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1633.md

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1605PartialElementDocumentationMustHaveSummary.cs
```diff
@@ -95,6 +95,13 @@ protected override void HandleXmlElement(SyntaxNodeAnalysisContext context, bool
...
+                /*if (completeDocumentation.Nodes().OfType<XElement>().Any(element => element.Name == XmlCommentHelper.ExcludeXmlTag))
+                {
+                    return;
+                }*/
+
+                // var hasExcludeTag = completeDocumentation.Nodes().OfType<XElement>().Any(element => element.Name == XmlCommentHelper.ExcludeXmlTag);
```
RobSiklos: What's up with this commented code?

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1604UnitTests.cs
```diff
@@ -188,6 +213,21 @@ public class ClassName
...
+        public async Task TestMethodWithExcludedDocumentationAsync()
+        {
+            var testCode = @"
+/// <summary>
+/// 
+/// </summary>
+public class ClassName
+{
+    /// <exclude/>
+    public void Test() { }
```
sharwell, line 226: ❔ Should we also disable checks for a member if the containing type is marked with `<exclude/>`?

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1604UnitTests.cs
```diff
@@ -63,6 +63,21 @@ public async Task TestTypeWithInheritedDocumentationAsync(string typeName)
...
 
+        [Theory]
+        [InlineData("enum")]
+        [InlineData("class")]
+        [InlineData("struct")]
+        [InlineData("interface")]
+        public async Task TestTypeWithExcludedDocumentationAsync(string typeName)
+        {
+            var testCode = @"
+/// <exclude/>
```
sharwell, line 74: 💭  It seems like the `<exclude/>` element shouldn't always exclude the element, but rather it should force a public element to be treated like a non-public element. If documentation is required for non-public elements, diagnostics should still be reported for missing documentation.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1609PropertyDocumentationMustHaveValue.cs
sharwell: ➡️ This was resolved

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1605PartialElementDocumentationMustHaveSummary.cs
sharwell: ➡️ This was resolved

PR #3128, opened 2020-03-17
Fix SA1642 cctor doc requirement

### LINKED ISSUE #3127: SA1642 doesn't allow static constructors to append text as instance constructors can
SA1642 allows the standard ctor text to begin with the canned text but then go on to elaborate. But it doesn't allow the same ability on cctor (static constructors).

> Constructor summary documentation should begin with standard text

Note in both ctor and cctor cases, the diagnostic says  "should *begin* with", and yet cctor isn't allowed to merely begin.

**Repro**

In the following class, the ctor is fine, but SA1642 disallows the cctor.

```cs
public class A
{
	/// <summary>
	/// Initializes static members of the <see cref="A"/> class
	/// so that it works.
	/// </summary>
	static A()
	{
	}

	/// <summary>
	/// Initializes a new instance of the <see cref="A"/> class
	/// so that it works.
	/// </summary>
	public A()
	{
	}
}
```

StyleCop version: 1.2.0-beta.164

AArnott:
Fixes #3127

Should this be rebased and retargeted to a maintenance branch so we can fix a stable version?

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/DocumentationResources.Designer.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/DocumentationResources.Designer.cs
@@ -1763,7 +1763,7 @@ namespace StyleCop.Analyzers.DocumentationRules {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to  {0}..
+        ///   Looks up a localized string similar to  {0}.
         /// </summary>
         internal static string StaticConstructorStandardTextSecondPart {
             get {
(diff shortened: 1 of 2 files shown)


sharwell:
> Should this be rebased and retargeted to a maintenance branch so we can fix a stable version?

This bug has existed for a few years with only one report. I wouldn't expect it to meet a servicing bar so **master** is fine.

AArnott:
There are already tests for this. But they're dependent on the resx file strings. So by changing the product, it implicitly changed the test. Perhaps not the best design, but you can consider it tested. :)

PR #3130, opened 2020-03-19
Add init script which installs the .NET SDK required by global.json

AArnott:
This simply makes it easier to get started on the repo.

The init script by default does _not_ require elevation, but for better VS interop the `-InstallLocality machine` switch can be added to require elevation and install the .NET Core SDK at a machine-wide location where VS will discover it.
Without elevation, command line build will work after running init and VS will work if launched from that same command line.

### DIFF
--- a/init.ps1
+++ b/init.ps1
@@ -1,0 +1,57 @@
+<#
+.SYNOPSIS
+Installs dependencies required to build and test the projects in this repository.
+.DESCRIPTION
+This MAY not require elevation, as the SDK and runtimes are installed to a per-user location,
+unless the `-InstallLocality` switch is specified directing to a per-repo or per-machine location.
+See detailed help on that switch for more information.
+.PARAMETER InstallLocality
+A value indicating whether dependencies should be installed locally to the repo or at a per-user location.
+Per-user allows sharing the installed dependencies across repositories and allows use of a shared expanded package cache.
+Visual Studio will only notice and use these SDKs/runtimes if VS is launched from the environment that runs this script.
+Per-repo allows for high isolation, allowing for a more precise recreation of the environment within an Azure Pipelines build.
+When using 'repo', environment variables are set to cause the locally installed dotnet SDK to be used.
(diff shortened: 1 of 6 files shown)


init.cmd
```diff
@@ -0,0 +1,4 @@
+@echo off
```
sharwell: ❗️  This file is not necessary, and impairs usability by undermining the parameter help support provided by PowerShell. This repository uses PowerShell directly.

init.ps1
```diff
@@ -0,0 +1,57 @@
...
+    [ValidateSet('repo','user','machine')]
+    [string]$InstallLocality='user',
+    [Parameter()]
+    [switch]$NoPrerequisites,
+    [Parameter()]
+    [switch]$NoRestore
+)
+
+if (!$NoPrerequisites) {
+    & "$PSScriptRoot\tools\Install-DotNetSdk.ps1" -InstallLocality $InstallLocality
```
sharwell:
```suggestion
    & "$PSScriptRoot\build\Install-DotNetSdk.ps1" -InstallLocality $InstallLocality
```

init.ps1
```diff
@@ -0,0 +1,57 @@
...
+When using 'repo', environment variables are set to cause the locally installed dotnet SDK to be used.
+Per-repo can lead to file locking issues when dotnet.exe is left running as a build server and can be mitigated by running `dotnet build-server shutdown`.
+Per-machine requires elevation and will download and install all SDKs and runtimes to machine-wide locations so all applications can find it.
+.PARAMETER NoPrerequisites
+Skips the installation of prerequisite software (e.g. SDKs, tools).
+.PARAMETER NoRestore
+Skips the package restore step.
+#>
+[CmdletBinding(SupportsShouldProcess=$true)]
+Param (
```
sharwell: ❓ Should we update the AppVeyor and Azure Pipelines definitions to invoke this before builds?

init.cmd
AArnott: Removed

init.ps1
AArnott: Done

PR #3158, opened 2020-05-31
Fix SA1008 for attributed tuple parameters

### LINKED ISSUE #3117: False Positive on SA1008 for Attributes on Array of ValueTuples
Code sample:

```csharp
private void UpdateSomething([NotNull] (string key, string value)[] keyValuePairs) 
{
}
```

SA1008 is thrown for the space between the tuple and the Attribute

dymanoid:
Fixes #3117 

Parameter attributes should be followed by a space. SA1008 reported this space for tuples because of missing handling of the attribute special case.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1008CSharp7UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1008CSharp7UnitTests.cs
@@ -997,11 +997,11 @@ namespace TestNamespace
 
         void TestMethod3([NotNull][Custom] (string, string) tuple) { }
 
-        void TestMethod4([NotNull](string, string) tuple) { }
+        void TestMethod4([NotNull]{|#0:(|}string, string) tuple) { }
 
-        void TestMethod5([NotNull, Custom](string, string) tuple) { }
+        void TestMethod5([NotNull, Custom]{|#1:(|}string, string) tuple) { }
 
-        void TestMethod6([NotNull][Custom](string, string) tuple) { }
+        void TestMethod6([NotNull][Custom]{|#2:(|}string, string) tuple) { }
     }
 }
 ";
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1008CSharp7UnitTests.cs
```diff
@@ -978,5 +978,65 @@ public void TestMethod(params (string name, string value)[] options)
...
+        void TestMethod6([NotNull][Custom] (string, string) tuple) { }
+    }
+}
+";
+
+            DiagnosticResult[] expectedDiagnostics =
+            {
+                Diagnostic(DescriptorPreceded).WithLocation(15, 35),
+                Diagnostic(DescriptorPreceded).WithLocation(17, 43),
+                Diagnostic(DescriptorPreceded).WithLocation(19, 43),
```
sharwell, line 1036: If possible, can you do one of the following?

1. If the `WithLocation(int)` overload is available, use this syntax:

    ```suggestion
                    Diagnostic(DescriptorPreceded).WithLocation(0),
                    Diagnostic(DescriptorPreceded).WithLocation(1),
                    Diagnostic(DescriptorPreceded).WithLocation(2),
    ```

    Then you can mark the locations 0, 1, and 2 above by replacing the `(` where the diagnostic occurs with `{|#0:(|}` (and using 1 and 2 for the others).

2. If the `WithLocation(int)` overload is _not_ available (meaning we are still on an older version of the testing library, just replace the calls to `WithLocation` with calls to `WithSpan`. I believe this is the correct change:

    ```suggestion
                    Diagnostic(DescriptorPreceded).WithSpan(15, 35, 15, 36),
                    Diagnostic(DescriptorPreceded).WithSpan(17, 43, 17, 44),
                    Diagnostic(DescriptorPreceded).WithSpan(19, 43, 19, 44),
    ```

dymanoid: That's a great suggestion, thanks! The new markup syntax is already available in the test projects. Just pushed a commit which fixes that.

PR #3144, opened 2020-04-21
documentation "Readability Rules" SA1119 missed fix

bocus:

PR #3165, opened 2020-07-01
Fixed SA1119 listed on wrong file

tallesl:


sharwell:
SA1119 is already listed in the correct file. See also #3144.

PR #3177, opened 2020-07-25
Exclude null-forgiving operator for SA1013

### LINKED ISSUE #3172: SA1013 should not apply to the null-forgiving operator
When using C#8 nullable reference types, SA1013 ("Closing brace should be followed by a space") has undesirable behavior when using the null-forgiving operator (`!`):

```
// It is idiomatic for the null-forgiving operator have no white-space between itself and what it is "operating" on.
// StyleCop.Analyzers is ok with these usages.
var a = 5!;
var b = this.DoSomething()!;

// SA1013: "Closing brace should be followed by a space"
var c = new { Value = "a" }!;

// how to appease SA1013, this is weird and not inline with other usages of !
var d = new { Value = "a" } !;
```

My proposal is that SA1013 should make an exception for the null-forgiving operator to be inline with idiomatic usage.

vweijsters:
Fixes #3172

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
@@ -3,17 +3,11 @@
 
 namespace StyleCop.Analyzers.Test.CSharp8.SpacingRules
 {
-    using System;
-    using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
 
-    using Microsoft.CodeAnalysis;
-    using Microsoft.CodeAnalysis.CodeFixes;
     using Microsoft.CodeAnalysis.CSharp;
-    using Microsoft.CodeAnalysis.Diagnostics;
     using Microsoft.CodeAnalysis.Testing;
-    using Microsoft.CodeAnalysis.Testing.Verifiers;
 
     using StyleCop.Analyzers.Test.CSharp7.SpacingRules;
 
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
```diff
@@ -62,5 +74,54 @@ public void TestMethod(object value)
...
+        }
+
+        private class NullableCSharpAnalyzerTest : AnalyzerTest<XUnitVerifier>
+        {
+            public override string Language => LanguageNames.CSharp;
+
+            protected override string DefaultFileExt => "cs";
+
+            protected override CompilationOptions CreateCompilationOptions()
+                => new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary, allowUnsafe: true, nullableContextOptions: NullableContextOptions.Enable);
```
sharwell, line 116: ❔ Why not just put `#nullable enable` at the top of the file?

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
```diff
@@ -62,5 +74,54 @@ public void TestMethod(object value)
...
+        public async Task TestCloseBraceWithNullForgivingOperatorAsync()
+        {
+            const string testCode = @"
+public class Foo
+{
+    public void TestMethod()
+    {
+        var test = new[]
+        {
+            new { Value = default(string) },
```
sharwell, line 93:
```suggestion
```

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/SpacingRules/SA1013CSharp8UnitTests.cs
vweijsters: Lack of knowledge :wink: I'll change it.

PR #3178, opened 2020-07-26
SA1119: Added support for typecast followed by switch expression

### LINKED ISSUE #3171: "SA1119 Statement should not use unnecessary parenthesis" on required parens around switch expression
```cs
class C
{
    string M(int n, object a, object b)
    {
        // ⚠ SA1119 Statement should not use unnecessary parenthesis
        //             ↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓
        return (string)(n switch { 1 => a, 2 => b });
    }
}
```

Parens are required. Applying the fix changes the meaning of the code and includes this stray space:

```cs
class C
{
    string M(int n, object a, object b)
    {
        // ❌ CS0030 Cannot convert type 'int' to 'string'
        //     ↓↓↓↓↓↓↓↓↓↓
        return (string) n switch { 1 => a, 2 => b };
    }
}
```

Nitpick: Shouldn't the message say "parentheses" instead of "parenthesis"? If they are showing as unnecessary, both are always present.

vweijsters:
Fixes #3171

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/MaintainabilityRules/SA1119CSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/MaintainabilityRules/SA1119CSharp8UnitTests.cs
@@ -42,7 +42,45 @@ public class Foo
         }
 
         /// <summary>
-        /// Verifies that a swith expression with unnecessary parenthesis is handled correcly.
+        /// Verifies that a type cast followed by a switch expression with unnecessary parenthesis is handled correctly.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [Fact]
+        [WorkItem(3171, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3171")]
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/MaintainabilityRules/SA1119CSharp8UnitTests.cs
```diff
@@ -3,9 +3,80 @@
...
+        [Fact]
+        [WorkItem(3171, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3171")]
+        public async Task TestTypeCastFollowedBySwitchExpressionIsHandledCorrectlyAsync()
+        {
+            const string testCode = @"
+public class Foo
+{
+    public string TestMethod(int n, object a, object b)
+    {
+        return (string)(n switch { 1 => a, 2 => b });
```
sharwell, line 36: 💡 Can you add a second test for this?

```
(string)((n switch { 1 => a, 2 => b }))
```

vweijsters: done

PR #3182, opened 2020-08-06
Update "LICENSE" so that it's recognized as MIT

xtqqczze:
This PR replicates dotnet/runtime#36744 (it also removes the All Rights Reserved).

License text is from: https://choosealicense.com/licenses/mit/


xtqqczze:
Github is now recognising the DotNetAnalyzers/StyleCopAnalyzers license as MIT.

PR #3186, opened 2020-08-27
Add missing rule SA1119 to table

WillMoreland:
## Changes
The Readabilty Rules table was missing one of the rules, I've added it 🙏🏻


sharwell:
Duplicate of #3144

sharwell:
@WillMoreland this rule is currently listed in a different table. The category was changed after the ID was created (but still back in the days of StyleCop Classic). See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3144#issuecomment-622446238.

Issue #2801, opened 2018-10-24
SA1500 fires for the while clause of do/while statement


smaillet:
The following generates an SA1500 error when it shouldn't:

```C#
bool condition = true;
do
{
    condition = DetermineCondition();
} while( condition );
```
Forcing placement of while onto a separate line is counter to common code style practices. At the very least there should be a configuration option to set how this rule should handle the placement of the while clause.

vweijsters:
The behaviour is by design. I've marked this `needs discussion` to discuss if a configuration is wanted.

zahirtezcan-bugs:
👍
When I google "c# do while", results show the tutorials which have similar syntax. That is an indication for the common sense imho.
And there is also [Microsoft C# Language Reference for do keyword](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/do)

Note: I assume that the spacing inside and around condition parantheses are non-intentional in the @smaillet's sample code.

AJH16:
I'd second this being an odd case.  I've always learned that the while should be on the same line as the end of the loop for a do while on the few cases where they are actually useful.  I could see the argument that front loaded conditions are on a separate line so why wouldn't the follower, but given how rarely one sees a do/while, the difference in formatting is probably helpful for highlighting the difference.

werwolfby:
@vweijsters Are we still discussing it?

This issues is really annoying, I want to keep this inspection, but additionally want to keep while on the same line with closing bracket.

I can prepare fix to ignore this for `do/while`.

Do we need separate inspection, or separate setting in `stylecop.json` to show hint for `do/while`?

zvonimir:
Hi all. Any news on this one? It would be really great to have it fixed since currently because of this I cannot enable SA1500. Thanks!

vweijsters:
@sharwell there is no config option for this in .editorconfig, but I'm willing to allow someone else to pick this up and make it configurable. What are your thoughts on it?

Kevin-Andrew:
> @sharwell there is no config option for this in .editorconfig, but I'm willing to allow someone else to pick this up and make it configurable. What are your thoughts on it?

@vweijsters, I have implemented a new configuration setting to support this. I called it `allowDoWhileOnClosingBrace`, as a boolean property of the `layoutRules`. By default, it is `false`, which means the default behavior of StyleCop remains the same, requiring the `while` expression to be on its own line.

When the setting is `true`, it allows for the following:
```
[...]

PR #3196, opened 2020-09-06
Issue #2801: SA1500 fires for the while clause of do/while statement

Kevin-Andrew:
Provide new configuration setting, `layoutRules.allowDoWhileOnClosingBrace` such that when enabled, the following will be allowed:
```
do
{
    Console.WriteLine("test");
} while (false);
```

See Issue #2801.


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/LayoutRules/SA1500CodeFixProvider.cs
```diff
@@ -102,19 +102,23 @@ private static async Task<Document> GetTransformedDocumentAsync(Document documen
...
 
                     // Check if we need to apply a fix after the brace. No fix is needed when:
                     // - The closing brace is followed by a semi-colon or closing paren
                     // - The closing brace is the last token in the file
+                    // - The closing brace is followed by the while expression of a do/while loop and the
+                    //   allowDoWhileOnClosingBrace setting is enabled.
                     var nextToken = braceToken.GetNextToken();
                     var nextTokenLine = nextToken.IsKind(SyntaxKind.None) ? -1 : LocationHelpers.GetLineSpan(nextToken).StartLinePosition.Line;
                     var isMultiDimensionArrayInitializer = braceToken.IsKind(SyntaxKind.OpenBraceToken) && braceToken.Parent.IsKind(SyntaxKind.ArrayInitializerExpression) && braceToken.Parent.Parent.IsKind(SyntaxKind.ArrayInitializerExpression);
+                    var allowDoWhileOnClosingBrace = settings.LayoutRules.AllowDoWhileOnClosingBrace && nextToken.IsKind(SyntaxKind.WhileKeyword) && (braceToken.Parent?.IsKind(SyntaxKind.Block) ?? false) && (braceToken.Parent.Parent?.IsKind(SyntaxKind.DoStatement) ?? false);
```
Kevin-Andrew, line 116: This line shows only partial code coverage from the Unit Tests and would be the reason for the decrease in Codecov.

image

The reason is because there isn't a test in which the `braceToken.Parent` is `null`. I'm not completely familiar with the flow of execution and what is possible and what isn't, so I wrote this just being safe. I suppose, given the `SyntaxTree`, it may be impossible for `braceToken.Parent` to be `null`, but I just don't know. I tried to write some Unit Tests that had malformed syntax and they would fail with a compiler error.

If someone is sure that `braceToken.Parent` can never be `null`, then I'd be happy to change it. I know the line of code also uses `braceToken.Parent` without the Null-Conditional operator, but I wasn't sure if the fact that it is applicable to a Multi-Dimensional Array Initializer, that the syntax may be different. So I just wanted to be extra safe.

But we can remove it if we want the Codcov percentage to go back up.

documentation/Configuration.md
```diff
@@ -418,6 +418,7 @@ The following properties are used to configure layout rules in StyleCop Analyzer
 | --- | --- | --- | --- |
 | `newlineAtEndOfFile` | `"allow"` | 1.0.0 | Specifies the handling for newline characters which appear at the end of a file |
 | `allowConsecutiveUsings` | `true` | 1.1.0 | Specifies if SA1519 will allow consecutive using statements without braces |
+| `allowDoWhileOnClosingBrace` | `false` | >1.2.0 | Specifies if SA1500 will allow the `while` expression of a `do-while` loop to be on the same line as the closing brace, as is generated by the default code snippet of Visual Studio |
```
Kevin-Andrew, line 421: I see that there is a 1.2.0-beta in the works:
image
So I put the version number here as `>1.2.0` because I don't know how to predict what version this might end up in.

We can change it to something else, or omit these changes from this PR and make another one or something later.

sharwell:
@Kevin-Andrew I may still merge this. Let me give it some thought today.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1500BracesForMultiLineStatementsMustNotShareLine.cs
```diff
@@ -288,11 +288,10 @@ private static void CheckBraceToken(SyntaxNodeAnalysisContext context, SyntaxTok
...
                     // places the while expression on the same line as the closing brace, some users
                     // may want to allow that and not have SA1500 report it as a style error.
-                    if (context.Options.GetStyleCopSettings(context.CancellationToken).LayoutRules.AllowDoWhileOnClosingBrace)
+                    if (context.GetStyleCopSettings(context.CancellationToken).LayoutRules.AllowDoWhileOnClosingBrace)
                     {
-                        var openBracePreviousToken = openBraceToken.GetPreviousToken(includeZeroWidth: true);
-
-                        if (openBracePreviousToken.IsKind(SyntaxKind.DoKeyword))
+                        if (openBraceToken.Parent.IsKind(SyntaxKind.Block)
+                            && openBraceToken.Parent.Parent.IsKind(SyntaxKind.DoStatement))
```
Kevin-Andrew, line 294: Since I don't know the syntax tree well enough, I just want to make sure, as I had similar code and this same question elsewhere, "Is using `.Parent.Parent` safe here?" We know that `openBraceToken.Parent` is not `null` and is of kind `Block`, but does that guarantee that `.Parent.Parent` will never be `null`? For example what if I have code like the following that uses some extra curly braces for scoping:
```
{ // Some braces for scoping
   var x = 0;
   do
   {
      x = 1;
   } while (x == 0);
}
```

sharwell: `IsKind` allows null, and `openBraceToken.Parent.IsKind(SyntaxKind.Block)` is only true if `openBraceToken.Parent` is not null.

Kevin-Andrew:
Thanks for looking at this @sharwell. I see that one of the required automation builds is failing.
image
I've looked at the information/logs in the build pipeline, but I'm not familiar with it enough to tell what the actual problem might be. But if there is anything more I can help with, I'd be happy to do so.

sharwell:
I restarted the build, and it should pass this time. I'm guessing our timeout is too short to handle all cases, but I can increase it.

sharwell (approved):

PR #3209, opened 2020-09-26
Update naming rules to support records

### LINKED ISSUE #3168: Since C# primary constructor parameters will have to be pascal-cased, what should additional constructors do?
https://github.com/dotnet/csharplang/blob/master/proposals/records.md
https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-05-27.md

Public properties are created with the same names as the parameters, including casing. I'm assuming camel-cased public properties are out. That leaves a requirement of pascal-casing the parameters if you use the C# 9 positional records feature.

Should consistency between constructors in the same type be enforced?

```cs
record Person(string FirstName, string? MiddleName, string LastName)
{
    public Person(string FirstName, string LastName) // Casing?
        : this(FirstName, MiddleName: null, LastName)
    {
    }
}
```

What about non-positional records? Should all constructors still use pascal-cased parameter names in case the record is refactored into a positional record, since changing the casing at that point would be a source-breaking change?

```cs
record Person
{
    public Person(string FirstName, string? MiddleName, string LastName) // Casing?
    {
        this.FirstName = FirstName;
        this.MiddleName = MiddleName;
        this.LastName = LastName;
    }

    public Person(string FirstName, string LastName) // Casing?
        : this(FirstName, MiddleName: null, LastName)
    {
    }

    public string FirstName { get; init; }
    public string MiddleName { get; init; }
    public string LastName { get; init; }
}
```

### LINKED ISSUE #3181: SA1313 is shown three times on a positional record
Using StyleCopAnalyzers 1.2.0-beta.205 and VS 16.7-p6 with .NET Core SDK 3.1.400-preview-015203:

```cs
// ⚠ SA1313 Parameter 'A' should begin with lower-case letter
// ⚠ SA1313 Parameter 'A' should begin with lower-case letter
// ⚠ SA1313 Parameter 'A' should begin with lower-case letter
public record R(int A);
```

sharwell:
Closes #3168
Fixes #3181

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/NamingRules/SA1300CSharp9UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/NamingRules/SA1300CSharp9UnitTests.cs
@@ -3,9 +3,86 @@
 
 namespace StyleCop.Analyzers.Test.CSharp9.NamingRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.CSharp;
     using StyleCop.Analyzers.Test.CSharp8.NamingRules;
+    using StyleCop.Analyzers.Test.Verifiers;
+    using Xunit;
+    using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
+        StyleCop.Analyzers.NamingRules.SA1300ElementMustBeginWithUpperCaseLetter,
+        StyleCop.Analyzers.NamingRules.RenameToUpperCaseCodeFixProvider>;
 
     public class SA1300CSharp9UnitTests : SA1300CSharp8UnitTests
     {
+        [Fact]
+        public async Task TestPositionalRecord1Async()
+        {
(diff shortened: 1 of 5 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/NamingRules/SA1300CSharp9UnitTests.cs
```diff
@@ -3,9 +3,86 @@
...
+            {
+                ReferenceAssemblies = GenericAnalyzerTest.ReferenceAssembliesNet50,
+                TestCode = testCode,
+                ExpectedDiagnostics =
+                {
+                    // /0/Test0.cs(2,15): warning SA1300: Element 'r' should begin with an uppercase letter
+                    Diagnostic().WithLocation(0).WithArguments("r"),
+
+                    // /0/Test0.cs(2,15): warning SA1300: Element 'r' should begin with an uppercase letter
+                    Diagnostic().WithLocation(0).WithArguments("r"),
```
sharwell, line 51: 📝 This is a bug in the analyzer driver

PR #3218, opened 2020-10-06
Configuration.md: Add link to rule set example file

perlun:
It might be easy to create this file from Visual Studio (?), but not all of us are using Visual Studio (using Rider myself on Linux). And in general, it doesn't hurt to provide an example that users can benefit from reading.

### DIFF
--- a/documentation/Configuration.md
+++ b/documentation/Configuration.md
@@ -14,7 +14,7 @@ StyleCop Analyzers is configured using two separate mechanisms: code analysis ru
 
 Code analysis rule sets are the standard way to configure most diagnostic analyzers within Visual Studio. Information about creating and customizing these files can be found in the [Using Rule Sets to Group Code Analysis Rules](https://docs.microsoft.com/visualstudio/code-quality/using-rule-sets-to-group-code-analysis-rules) documentation on docs.microsoft.com.
 
-Here is an example `.ruleset` file from the Roslyn SDK: https://github.com/dotnet/roslyn-sdk/blob/3e94c5fe1ee981a885c34e2e39383bc4420f2557/RoslynSDK.ruleset. This file illustrates how the severities for individual rules can easily be configured as needed.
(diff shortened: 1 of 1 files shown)


documentation/Configuration.md
```diff
@@ -14,6 +14,8 @@ StyleCop Analyzers is configured using two separate mechanisms: code analysis ru
 
 Code analysis rule sets are the standard way to configure most diagnostic analyzers within Visual Studio. Information about creating and customizing these files can be found in the [Using Rule Sets to Group Code Analysis Rules](https://docs.microsoft.com/visualstudio/code-quality/using-rule-sets-to-group-code-analysis-rules) documentation on docs.microsoft.com.
 
+Here is an example `.ruleset` file from the Roslyn SDK: https://github.com/dotnet/roslyn-sdk/blob/3e94c5fe1ee981a885c34e2e39383bc4420f2557/RoslynSDK.ruleset. This file illustrates how the severities for individual rules can easily be configured as needed.
```
sharwell, line 17: 💡 This should link to [StyleCopAnalyzersDefault.ruleset](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/rulesets/StyleCopAnalyzersDefault.ruleset)

sharwell (changes requested):

documentation/Configuration.md
sharwell:
```suggestion
An example rule set file containing the default StyleCop Analyzers configuration is available at <https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/rulesets/StyleCopAnalyzersDefault.ruleset>.
```

perlun: Good point, even better that way. :+1: :bow:

sharwell (approved):

PR #3251, opened 2020-11-27
Update Configuration.md

mayermart:
I think $(MSBuildThisFileDirectory) can be used to simplify the .props file

### DIFF
--- a/documentation/Configuration.md
+++ b/documentation/Configuration.md
@@ -690,15 +690,10 @@ Example `.props` file:
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


sharwell:
💭 This seems fine, but I'm not one of the users of this configuration form. It would be good to get a review from one of the original contributors of this section.

ngeor:
Hi, I haven't worked with .NET pretty much since I did this contribution more than 2 years ago :-( 

The configuration sample was about having an external NuGet package ( MyCompany.MyRules.nuget ) that holds StyleCop configuration, so I can share it in multiple projects without copy pasting around the config file.

If this "MSBuildThisFileDirectory" variable somehow points to a NuGet package, go ahead with the changes. If your change is actually in a different direction (copy pasting the stylecop file rules in multiple projects),  I think then you should add it to the docs as a new section.

mayermart:
The [Documentation](https://docs.microsoft.com/de-de/visualstudio/msbuild/msbuild-reserved-and-well-known-properties?view=vs-2019) includes a hint which covers this scenario:

`Tip: You can specify a relative path in a targets file that's relative to the targets file and not relative to the original project file.`

The NuGet Package i created this way works at least with PackageReference in old-style and sdk-style Projects

sharwell (approved):

PR #3252, opened 2020-11-30
Fix SA1023 for C# 9 function pointer parameters

NextTurn:


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1023CSharp9UnitTests.cs
```diff
@@ -3,9 +3,66 @@
...
+}
+";
+
+            var fixedCode = @"public class TestClass
+{
+    unsafe delegate*<int*> FuncPtr;
+}
+";
+
+            var expected = Diagnostic(DescriptorNotPreceded).WithLocation(3, 26);
```
sharwell, line 45: 💡  Can you switch these tests to use the `WithLocation(0)` form? The `testCode` will contain matching syntax like `{|#0: |}` to indicate the location of diagnostic 0 within the input.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1023CSharp9UnitTests.cs
```diff
@@ -3,9 +3,66 @@
...
+        StyleCop.Analyzers.SpacingRules.TokenSpacingCodeFixProvider>;
 
     public class SA1023CSharp9UnitTests : SA1023CSharp8UnitTests
     {
+        [Fact]
+        public async Task TestFunctionPointerParameterValidSpacingAsync()
+        {
+            var testCode = @"public class TestClass
+{
+    unsafe delegate*<int*> FuncPtr;
```
sharwell, line 23: 💡 Can you add tests for spaces before and after the `*` in `delegate*`?

sharwell, line 23: 💡 Can you add tests for spaces around an optional calling convention modifier?

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1023DereferenceAndAccessOfSymbolsMustBeSpacedCorrectly.cs
```diff
@@ -125,40 +137,27 @@ private static void HandleAsteriskToken(SyntaxTreeAnalysisContext context, Synta
                 allowAtLineStart = false;
                 allowAtLineEnd = true;
                 allowPrecedingSpace = false;
-                var nextToken = token.GetNextToken();
```
NextTurn, line 120: A local variable or function named 'nextToken' is already defined in this scope

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1023DereferenceAndAccessOfSymbolsMustBeSpacedCorrectly.cs
```diff
@@ -125,40 +137,27 @@ private static void HandleAsteriskToken(SyntaxTreeAnalysisContext context, Synta
...
-                case SyntaxKind.AsteriskToken:
-                    allowTrailingSpace = false;
-                    break;
-
-                default:
-                    allowTrailingSpace = true;
-                    break;
-                }
-
+                    SyntaxKind.OpenBracketToken or
```
sharwell, line 142: ❕ binary operators go at the beginning of the next line when the expression is split

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1023DereferenceAndAccessOfSymbolsMustBeSpacedCorrectly.cs
sharwell: This can be resolved with less code churn by removing `var`.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1023CSharp9UnitTests.cs
```diff
@@ -3,9 +3,166 @@
...
+}
+";
+
+            var expected = Diagnostic(DescriptorNotFollowed).WithLocation(0);
+            await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestFunctionPointerTypeWithCallingConventionValidSpacingAsync()
```
sharwell, line 119: 💡 Groups of tests for variations of the same scenario can be written as a single test using the following form:

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/7ae14438aac88d8c9ba1591cb21b20be3483f318/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp7/SpacingRules/SA1008CSharp7UnitTests.cs#L28-L42

This helps the reader see that all the cases are covered.

The example was implemented before the new `WithLocation` syntax was available, but you can combine both here by using `WithLocation(0)` for the first span, `WithLocation(1)` for the second, and so on.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1023CSharp9UnitTests.cs
```diff
@@ -3,9 +3,166 @@
...
+    unsafe delegate*<int*> FuncPtr;
+}
+";
+
+            var expected = Diagnostic(DescriptorNotFollowed).WithLocation(0);
+            await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task TestFunctionPointerTypeValidSpacingAsync()
```
sharwell, line 69: 💡 This is actually a duplicate case from `TestFunctionPointerParameterInvalidTrailingSpaceAsync`, since the `fixedCode` is run through the analyzer automatically. If the test cases are grouped as described in the previous comment, the duplication will go away at the same time.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1023CSharp9UnitTests.cs
```diff
@@ -3,9 +3,74 @@
...
+
+        [Fact]
+        public async Task TestFunctionPointerTypeInvalidSpacingAsync()
+        {
+            var testCode = @"public class TestClass
+{
+    unsafe delegate {|#0:*|}<int*> FuncPtr1;
+    unsafe delegate{|#1:*|} <int*> FuncPtr2;
+    unsafe delegate {|#2:*|} managed<int*> FuncPtr3;
+    unsafe delegate{|#3:*|}managed<int*> FuncPtr4;
```
sharwell, line 52: 💡 IIRC, the `unmanaged` keyword only works for .NET 5 and newer. I believe there are some instances of tests in the repository using `ReferenceAssemblies.Net.Net50` explicitly. Considering function pointers are only officially supported for .NET 5 and newer, it would be fine to just update this test to always use it and test both `managed` and `unmanaged` cases. I'm happy to leave this for a follow-up PR though.

sharwell (approved):

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1023CSharp9UnitTests.cs
NextTurn: I'm not sure I get your point. C# 9.0 is supported only on .NET 5 and newer. Reference assemblies are not needed in this case since the function pointer is not a new invention at the IL level.

NextTurn: Well... error CS8889: The target runtime doesn't support extensible or runtime-environment default calling conventions.

sharwell: It's looking for this field, which is only available in .NET 5 and newer:
https://github.com/dotnet/runtime/blob/2445324a501efaca19e136b7a7ca46d8ab5d95ec/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeFeature.cs#L18-L21

NextTurn: Shall I specify a calling convention to make the build pass as in #3254?

sharwell: I'd prefer the test be updated to this form instead:

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/6399f4986f460445e1d12054cedfb62d4c9c605a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/NamingRules/SA1300CSharp9UnitTests.cs#L41-L54

sharwell (approved):

sharwell:
Thanks @NextTurn !

PR #3253, opened 2020-11-30
Remove unnecessary usings

NextTurn:


sharwell (approved):
Keep in mind that most changes to just the code style are closed without merge. I'm making an exception for this one since it was easy to review.

PR #3254, opened 2020-11-30
Fix SA1011 for function pointer unmanaged calling convention list

NextTurn:


sharwell (approved):
I suspect this may need .NET 5 reference assemblies, but if the build passes as-is then this works for me.

sharwell:
Looks like it worked. Thanks @NextTurn !

PR #3259, opened 2020-12-05
SA1114: Allow all valid trivia

NextTurn:
Allow all valid trivia by simply counting the line breaks in between.

You want more tests?

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1114ParameterListMustFollowDeclaration.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1114ParameterListMustFollowDeclaration.cs
@@ -195,22 +195,8 @@ namespace StyleCop.Analyzers.ReadabilityRules
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


TaffarelJr:
Any chance of getting some traction on this? It's been almost 2 years.
Getting really tired of this false positive.

bjornhellander:
Seems to be tests missing here. Any other opinions?

bjornhellander:
I had another look. This PR allows trivia in all cases that SA1114 checks, e.g. array declarations and attribute argument lists, not only method invocations. I am also pretty sure it allows more kinds of trivia than @vweijsters' earlier fix for directive trivia in method declarations, e.g. `#define`. I'm not saying this is necessarily a problem, bit a bit odd to e.g. allow a `#define` inside an indexer declaration.

Other ways forward could be to 1) apply @vweijsters' change for method declarations for invocations as well (but this PR reduces code quite a bit, so I like the general idea of it), or 2) do a mix, where we are still strict about what trivia to allow, and maybe even in what places, but try to reduce code in a similar way to this PR. I would vote for alternative 2. What do you think, @sharwell?

Tests should still be added, regardless of path taken.

PR #3264, opened 2020-12-05
Support records in several additional rules

### LINKED ISSUE #3236: SA1201 for records says '<unknown>' should not follow an enum
grafik

sharwell:
* Add tests for records in many rules
    * SA1106
    * SA1137 (with fixes)
    * SA1201 (with fixes)
    * SA1205 (with fixes)
    * SA1400 (with fixes)
    * SA1500
    * SA1502 (with fixes)
    * SA1505
    * SA1508
    * SA1604
    * SA1605
    * SA1606
    * SA1607
    * SA1608
    * SA1618
    * SA1619
    * SA1625
* Update SA1205 to support `private protected`

Fixes #3236


ivan-nosar:
Issue still persists for the files that contains nested record type definitions (accessibility modifiers doesn't matter)

I encountered the following analyzer failure when attempted to build the code:

```cs
public class TopLevelClass : IDisposable
{
   // Some logic that refers NestedRecordType

   private record NestedRecordType(string name, int value);
}
```

Error message: `Analyzer 'StyleCop.Analyzers.OrderingRules.SA1201ElementsMustAppearInTheCorrectOrder' threw an exception of type 'System.Collections.Generic.KeyNotFoundException' with message 'The given key was not present in the dictionary.'.`

Record types defined with `public` modifier at the top-level doesn't cause that issue.

bjornhellander:
@ivan-nosar I suspect that you are using an old version. I can at least not reproduce this with 1.2.0-beta.556. The latest beta is currently the only way to get a good experience with these analyzers when using "new" language features. In particular, version 1.1.118 is from 2019 and does not handle modern c# well. If you are indeed using a late beta, then please create a new issue.

ivan-nosar:
@bjornhellander you're right, I'm using version `1.1.118` as it is latest one marked as stable - unfortunately, I can't update to beta, so I'll just avoid creating nested record types. I'm glad to hear that this issue was fixed in latest releases, though.

PR #3289, opened 2021-01-06
SA1130: Handle optional named arguments correctly

### LINKED ISSUE #3279: SA1130UseLambdaSyntax throws NullReferenceException
Using `StyleCop.Analyzers` version `1.1.118` and `1.2.0-beta.312`; `netstandard2.1`/`netcoreapp3.1` and C# `8.0`; we are getting NRE when building from Visual Studio, Rider and even when running build in CI/CD (`dotnet build`).

Exception:
```
Analyzer 'StyleCop.Analyzers.ReadabilityRules.SA1130UseLambdaSyntax' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
```

Code snippet (using GraphQL package):
```
            Field<ListGraphType<GenericGraphType>>(
                "foo",
                resolve: delegate
                {
                    var x = repository.GetAll().Select(e => new GenericModel { Id = e.Id, Name = e.Name });
                    return x;
                });
```

(I'm not able to open StyleCopAnalyzers on my machine (nor build it), won't be able to tackle this further; I can see there's test for this, not sure what exactly is causing the issue)

wdolek:
Fixes #3279, `NullReferenceException` is thrown when delegate is used as (optional) named parameter and argument index doesn't match method signature.

Fix is handling special case when argument is used with name, e.g. `F(argName: delegate { ... })`, otherwise everything works as before.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
@@ -880,5 +880,57 @@ public class TestClass
 
             await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
         }
+
+        [Fact]
+        [WorkItem(3279, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3279")]
+        public async Task TestDelegateUsedAsSecondNamedArgumentAsync()
+        {
+            var testCode = @"
+using System;
+using System.Linq;
+public class TypeName
+{
+    public void Test()
+    {
+        string[] capture = new string[] { ""test"" };
+        Test2(resolve: delegate
+        {
+            return capture.Single(v => v == ""test"");
+        });
+    }
+
+    private void Test2(string description = null, Func<object, string> resolve = null)
+    {
+        resolve(0);
+    }
+}";
+
+            string fixedCode = @"
+using System;
+using System.Linq;
+public class TypeName
+{
+    public void Test()
+    {
+        string[] capture = new string[] { ""test"" };
+        Test2(resolve: arg =>
+        {
+            return capture.Single(v => v == ""test"");
+        });
+    }
+
+    private void Test2(string description = null, Func<object, string> resolve = null)
+    {
+        resolve(0);
+    }
+}";
+
+            var expected = new[]
+            {
+                Diagnostic().WithSpan(9, 24, 9, 32),
+            };
+
+            await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -96,6 +96,28 @@ internal static ParameterListSyntax GetDelegateParameterList(ISymbol symbol, int
             return SyntaxFactory.ParameterList(SyntaxFactory.SeparatedList(syntaxParameters));
         }
 
+        internal static int FindArgumentIndex(SymbolInfo originalSymbolInfo, ArgumentSyntax argumentSyntax, BaseArgumentListSyntax argumentListSyntax)
```
sharwell, line 99: 💭 Perhaps this is better named `FindParameterIndex`?

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -96,6 +96,28 @@ internal static ParameterListSyntax GetDelegateParameterList(ISymbol symbol, int
...
+        {
+            // if delegate is passed as named argument of method try to find its position by argument name
+            if (originalSymbolInfo.Symbol.Kind == SymbolKind.Method
```
sharwell, line 102: ❔ Is this condition required since we have the `IMethodSymbol` check?

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
```diff
@@ -96,6 +96,28 @@ internal static ParameterListSyntax GetDelegateParameterList(ISymbol symbol, int
...
+                && originalSymbolInfo.Symbol is IMethodSymbol methodSymbol
+                && argumentSyntax.NameColon != null)
```
sharwell, line 104: 💡 This (`NameColon`) check is the fast and common case; consider moving it to first in this condition

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
```diff
@@ -880,5 +880,57 @@ public class TestClass
...
+        {
+            var testCode = @"
+using System;
+using System.Linq;
+public class TypeName
+{
+    public void Test()
+    {
+        string[] capture = new string[] { ""test"" };
+        Test2(resolve: delegate
```
sharwell, line 896: 💡 Can you add a test showing an explicit named argument, where the name doesn't match any parameter? This would be an error, but we still want to make sure the analyzer doesn't throw an exception.

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1130UseLambdaSyntax.cs
wdolek: Not required really, my intention was to compare enum value quickly before doing typecast (`is`), I will remove that condition.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
wdolek: @sharwell Not sure if I understand: using parameter name which doesn't fit method signature _should_ produce CS1739, shouldn't it? Is your intention to test that CS1739 is raised?

sharwell: Compiler errors prevent the build from succeeding, but the analyzers still run. We want to make sure the analyzer isn't going to fail for this case.

PR #3303, opened 2021-02-20
Added GlobalSuppressions.cs example for SA1649

whoward3:
Just added a GlobalSuppressions sample for SA1649 as it didn't have one.

### DIFF
--- a/documentation/SA1649.md
+++ b/documentation/SA1649.md
@@ -50,6 +50,10 @@ To fix a violation of this rule, rename the file to  match the name of the first
 ## How to suppress violations
 
 ```csharp
+[SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1649:FileNameMustMatchTypeName", Justification = "Reviewed.")]
+```
+
+```csharp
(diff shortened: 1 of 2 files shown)


documentation/SA1649.md
```diff
@@ -49,6 +49,10 @@ To fix a violation of this rule, rename the file to  match the name of the first
 
 ## How to suppress violations
 
+```csharp
+[SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1649:FileNameMustMatchTypeName", Justification = "Reviewed.")]
```
sharwell, line 53: ❓ Does this work? If so, can you add a test in [SA1649UnitTests.cs](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/5ba5c03f1ddb302c38771a6b5615ec5142ef8c8a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1649UnitTests.cs) showing that this syntax will prevent the issue from being reported?

sharwell (changes requested):

sharwell (approved):

Issue #3307, opened 2021-02-25
[New analyzer] SA1143 Use implicitly assignable foreach variable type


maxkoshevoi:
Related discussion: https://github.com/dotnet/runtime/issues/48529
Original issue: https://github.com/dotnet/roslyn-analyzers/issues/4479

### Describe the problem you are trying to solve

As described [here](https://rules.sonarsource.com/csharp/RSPEC-3217) and [here](https://stackoverflow.com/a/949835/4312132) `foreach` can throw `InvalidCastException` at runtime because of hidden explicit cast that is does under the hood.

This is very dangerous behavior for statically typed language to have since programmers don't expect successfully compiled code to throw `InvalidCastException` when they didn't do any casting.
I came across this behavior when I changed type of a collection to more generic one and then got this error at runtime.

This behavior was needed until generics appeared in C# 2, but it's still here with us 15 years after it's introduction.

### Describe suggestions on how to achieve the rule

While solution could be as easy as "always use `var` for `foreach` items", I'm not a fan of this approach. `var` makes code less readable in this scenario, and forcing someone to use `var` is not in the spirit of this keyword anyway (it was made to be optional).

I'm proposing analyzer that would detect that instance of type `T` from `IEnumerable<T>` that is passed to `foreach` can be cast to type specified for `item` of the `foreach`.

Something like this, but using [IsAssignableFrom](https://docs.microsoft.com/en-us/dotnet/api/system.type.isassignablefrom?view=net-5.0):
```cs
List<A> aList = new() { new() };
Foreach<B, A>(aList);

List<B> bList = new() { new() };
Foreach<A, B>(bList);

void Foreach<TItem, TCollectionValue>(IEnumerable<TCollectionValue> collection) where TCollectionValue : TItem
{
    foreach (TItem item in collection)
    {
    }
}

class A { }
class B : A { }
```
![image](https://user-images.githubusercontent.com/6609929/100783634-b4ab9280-3416-11eb-92d9-364b307f2043.png)

### Additional context

Here is code to reproduce the issue:
```cs
List<A> aList = new() { new() };
foreach (B item in aList)
{
}

class A { }
class B : A { }
```
![image](https://user-images.githubusercontent.com/6609929/100781552-d9523b00-3413-11eb-9ac4-ea41b8f50679.png)

maxkoshevoi:
Regarding what is and isn't a false positive, there are too two groups so far:
[...]

PR #3308, opened 2021-02-25
[New analyzer] SA1143 Use implicitly assignable foreach variable type

maxkoshevoi:
Port of PR from roslyn analyzers: https://github.com/dotnet/roslyn-analyzers/issues/4479

Fixes: #3307 

image

This analyzer reports on unobvious explicit conversion done under the hood by `foreach` loop. This conversion can result in runtime exception, but it's not guaranteed.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1143UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1143UnitTests.cs
@@ -1,0 +1,616 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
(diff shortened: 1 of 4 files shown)


maxkoshevoi:
Could you please help me port original PR? Right now it cannot find following types: 
`IForEachLoopOperation`, `IVariableDeclaratorOperation`, `CommonConversion`, `Compilation.ClassifyCommonConversion()`, `WellKnownTypeProvider`

Also `AnalyzeLoop` doesn't get invoked when I run the tests for some reason.

sharwell:
Closing per https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3307#issuecomment-785952062

@maxkoshevoi it looks like you've done quite a lot of work to implement this rule, only to find it doesn't into any of the well-known analyzer packages. My suggestion would be to create a brand new project in your own GitHub account, implement the rule the rule there, and publish your own NuGet package with it. Users who wish to use this rule can then install your package directly (multiple analyzer packages can be installed in a single project without interfering with each other).

PR #3371, opened 2021-07-19
SA1629 should allow full-sentence links instead of forcing the period to glow white

jnm2:
This looks ugly to me, and SA1629 should not force it:

image

The same affordance should be given for wrapping a whole sentence with `<a>`, `<see>`, and `<seealso>` which is currently given for wrapping a whole sentence with parentheses.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1629UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1629UnitTests.cs
@@ -545,15 +545,15 @@ public interface ITest
         }
 
         [Theory]
-        [InlineData("a")]
-        [InlineData("see")]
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1629UnitTests.cs
```diff
@@ -544,6 +544,23 @@ public interface ITest
             await VerifyCSharpFixAsync(testCode, expected, fixedTestCode, default).ConfigureAwait(false);
         }
 
+        [Theory]
+        [InlineData("a")]
+        [InlineData("see")]
+        [InlineData("seealso")]
```
sharwell, line 550: 📝 `<seealso>` is a section-level element, so it would not be expected inside `<summary>`.

sharwell (approved):

PR #3374, opened 2021-07-28
Add documentation for .editorconfig properties

FantasyTeddy:
I added some documentation for the .editorconfig properties introduced with #3285.


JobaDiniz:
Basically, all the `json` settings are supported in `editorconfig` file, right? You just need to "flat" the json path to create the **key** to be used in the editor config, is that right?

Issue #3402, opened 2021-11-12
Documentation culture does not respect DOTNET_SYSTEM_GLOBALIZATION_INVARIANT


rickardp:
Normally, you want to enforce culture-invariant builds to not risk affecting the build output by the culture of the build agent or (worse) workstation.
To resolve this, we run all builds with `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT`. This crashes some documentation rules because these are newing up a CultureInfo with the hard coded default of "en-US".
As a workaround we can add a config setting the "documentationCulture" to "" (which means the invariant culture).

This is needed because it seems the behaviour of this mode was changed in .NET 6.0, previously setting it meant that all cultures were treated as the invariant culture, as documented here
https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-invariant-mode.md#cultures-and-culture-data

With .net 6 it appears it instead throws an exception, so this is possibly to be treated as a .net 6 compatibility issue in this library.

(Note that there are many ways to enable the global invariant mode, so checking for the environment variable is not sufficient)

sharwell:
It's not clear what the issue here is.

rickardp:
For a repro case, try to run StyleCop in a Docker dotnet 6 SDK environment 

Repro.csproj:
```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.354" />
  </ItemGroup>
</Project>
```

TestClass.cs
```
namespace Test
{
    /// <summary>
    /// Hello
    /// </summary>
    public class TestClass
    {
    }
}
```

Dockerfile:
```
FROM mcr.microsoft.com/dotnet/sdk:6.0
ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1

COPY *.csproj /src/
COPY *.cs /src/
WORKDIR /src
RUN dotnet build
```

To repro:
Run `docker build .`

Expected output:
```
/src/TestClass.cs(1,1): warning SA1633: The file header is missing or not located at the top of the file. [/src/Repro.csproj]
/src/TestClass.cs(4,14): warning SA1629: Documentation text should end with a period [/src/Repro.csproj]
```

Actual output:
[...]

PR #3411, opened 2021-12-02
Issue #3402: Removed creation of CultureInfo with arbitrary names

bjornhellander:
Attempted fix for #3402. A bit of a hack :-), but should hopefully work.

Note: I haven't been able to verify this fix on my machine, due to not understanding how, but the problematic calls to the CultureInfo constructor with a culture name are gone. I can try again if someone can explain what to do.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1642SA1643CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1642SA1643CodeFixProvider.cs
@@ -83,7 +83,7 @@ namespace StyleCop.Analyzers.DocumentationRules
         {
             bool isStruct = typeDeclaration.IsKind(SyntaxKind.StructDeclaration);
             var settings = document.Project.AnalyzerOptions.GetStyleCopSettings(methodDeclaration.SyntaxTree, cancellationToken);
-            var culture = new CultureInfo(settings.DocumentationRules.DocumentationCulture);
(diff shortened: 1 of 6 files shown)


bjornhellander:
A test for SA1134 failed on the release build. I have seen this sometimes before this pr and have now run the tests locally a couple of times again without any problems. Pushing again to trigger a new build.

bjornhellander:
Oookey. No code changes relative to the first push and now lots of tests fail with one of the following exceptions:
- System.Runtime.InteropServices.COMException : Signature specified is zero-sized
- System.BadImageFormatException : Bad binary signature

Stilll don't think this is related to my change, so pushing again to trigger a new build.

sharwell (changes requested):
Based on an internal discussion, this change does not appear to adhere to the design/intent of `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT`. The documented approach to handling this situation is setting one of the following environment variables:

* Set `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` to `false`
* Set `DOTNET_SYSTEM_GLOBALIZATION_PREDEFINED_CULTURES_ONLY` to `false`

If I get a new suggestion from the team that implemented this option, we may reconsider the approach used by this tool.

Separately, we would accept a change to make the _default_ culture work in this configuration (i.e. cases where `DocumentationCulture` has not been set). The solution in this case is to use the invariant culture directly instead of trying to create a new instance of a different culture.

sharwell:
I submitted #3426 separately to make the default value "en-US" work with `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT`.

sharwell:
Superseded by #3426

PR #3452, opened 2022-02-08
Collect incremental state without passing compilation to output stage

sharwell:


StyleCop.Analyzers/StyleCop.Analyzers.CodeGeneration/SyntaxLightupGenerator.cs
```diff
@@ -30,22 +30,70 @@ private enum NodeKind
 
         public void Initialize(IncrementalGeneratorInitializationContext context)
         {
-            var compilation = context.CompilationProvider;
+            var referencedAssemblies = context.CompilationProvider.SelectMany(
```
sharwell, line 33: @chsienki this is `IncrementalValuesProvider<IAssemblySymbol>`, which should be efficient provided the `IAssemblySymbol` for a metadata reference is carried forward from one compilation to the next. Do you know if this is true for the IDE and/or command line?

StyleCop.Analyzers/StyleCop.Analyzers.CodeGeneration/SyntaxLightupGenerator.cs
```diff
@@ -1393,5 +1441,20 @@ public string GetAccessorResultType(SyntaxData syntaxData)
                 return null;
             }
         }
+
+        private record CompilationData(ImmutableDictionary<string, ExistingTypeData> ExistingTypes)
+        {
+        }
+
+        private record ExistingTypeData(string TypeName, ImmutableArray<string> MemberNames)
```
sharwell, line 1449: @chsienki is there a straightforward way to have a properly equatable list of strings in a record, without having to implement custom equality?

StyleCop.Analyzers/StyleCop.Analyzers.CodeGeneration/SyntaxLightupGenerator.cs
```diff
@@ -1393,5 +1441,20 @@ public string GetAccessorResultType(SyntaxData syntaxData)
                 return null;
             }
         }
+
+        private record CompilationData(ImmutableDictionary<string, ExistingTypeData> ExistingTypes)
```
sharwell, line 1445: @chsienki is there a straightforward way to have a properly equatable dictionary in a record, without having to implement custom equality?

StyleCop.Analyzers/StyleCop.Analyzers.CodeGeneration/SyntaxLightupGenerator.cs
chsienki: Yes, they should be carried forward. There is also `context.MetadataReferences` which provides the same thing. You don't get the `IAssemblySymbol` directly, but if you new up a compilation with the reference you can retrieve it, and the input will then never change unless the user adjusts references.

StyleCop.Analyzers/StyleCop.Analyzers.CodeGeneration/RoslynHashCode.cs
```diff
@@ -0,0 +1,462 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+// <auto-generated/>
+
+#nullable enable
+
+// NOTE: This code is derived from an implementation originally in dotnet/runtime:
+// https://github.com/dotnet/runtime/blob/v5.0.3/src/libraries/System.Private.CoreLib/src/System/HashCode.cs
```
manfred-brands, line 8: Could you reference [Microsoft.Bcl.HashCode](https://www.nuget.org/packages/Microsoft.Bcl.HashCode) instead?

StyleCop.Analyzers/StyleCop.Analyzers.CodeGeneration/SyntaxLightupGenerator.cs
sharwell: I'll switch over to `context.MetadataReferences` after 4.2.0 is released (this is the version that corrects the signature for this property).

StyleCop.Analyzers/StyleCop.Analyzers.CodeGeneration/RoslynHashCode.cs
sharwell: I prefer to avoid dependencies in the source generator, since it makes build integration much more complex.

Issue #3351, opened 2021-06-15
SA1516: misfire in C# 9 top-level statements


AArnott:
Given this Program.cs file:

```cs
// Copyright (c) Andrew Arnott. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.

using MinerForeman;

var foreman = new Foreman();
await foreman.RunAsync();
```

The last line in the file above produces:

```
Warning	SA1516	Elements should be separated by blank line	Program.cs
```

This repros with StyleCop.Analyzers 1.2.0-beta.333

ylabade:
Any news on this issue?

thomaslevesque:
I'm seeing this as well. Apparently _all_ top-level statements must be separated by a blank line, otherwise I have this warning.

What's weird is that the diagnostic doesn't report any file location... I have a warning, but I don't know where.

sharwell:
@thomaslevesque The reporting location bug is #3242, and should be fixed in newer releases.

thomaslevesque:
@sharwell thanks. Yes, I probably need to update to a newer version.

msmolka:
What is the status of this issue? do you know maybe in which pre-release it will be done? I'm using 1.2.0-beta.406 and still seeing issue on top=level statements

sharwell:
I don't believe anyone has submitted a fix for this issue yet.

madhusameena:
Any comments on this ?
or any workaround for this ?

maxkoshevoi:
> I don't believe anyone has submitted a fix for this issue yet.

@sharwell Created a fix: #3482

maxkoshevoi:
@sharwell Thanks for merging so quickly! Is there an ETA on next NuGet release?

domagojmedo:
Still not fixed in 1.2.0-beta.406

ultimaweapon:
That one is released before #3482 is merged.

sharwell:
Yep, waiting for a few more items to finish/merge and the next release should be soon.

PR #3482, opened 2022-04-18
Fix false positive in SA1516 between global statements

maxkoshevoi:
Fixes: #3351

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/LayoutRules/SA1516CSharp9UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/LayoutRules/SA1516CSharp9UnitTests.cs
@@ -24,7 +24,7 @@ namespace StyleCop.Analyzers.Test.CSharp9.LayoutRules
         /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
         [Fact]
         [WorkItem(3242, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3242")]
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/LayoutRules/SA1516CSharp9UnitTests.cs
```diff
@@ -55,5 +55,64 @@ public async Task TestStatementSpacingInTopLevelProgramAsync()
...
+                {
+                    OutputKind = OutputKind.ConsoleApplication,
+                    Sources = { testCode },
+                    ExpectedDiagnostics =
+                    {
+                        // /0/Test0.cs(2,1): warning SA1516: Elements should be separated by blank line
+                        Diagnostic().WithLocation(0),
+
+                        // /0/Test0.cs(2,1): warning SA1516: Elements should be separated by blank line
+                        Diagnostic().WithLocation(0),
```
sharwell, line 111: ❔ Do we know why this is reported twice?

sharwell (approved):

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/LayoutRules/SA1516CSharp9UnitTests.cs
maxkoshevoi: It's also reported twice in the first test in this file (from which I copied this one), so it's an existing behavior.
Not sure why it's reposted twice (or even is it actually reported twice, or it's a test system issue)

PR #3489, opened 2022-04-24
Feature/stylecop

bjornhellander:
Upgraded the StyleCop.Analyzers package reference, since the previously used version was rather old


bjornhellander:
This actually failed on my machine, but pushing anyway to see if it is a local problem...

bjornhellander:
Ok, I realized why it fails, although unit tests with the same code passes. There is 5 day old fix for generic delegate files names, which is not in the beta 406. I could temporarily change the file name to match beta 406 and then we would have to change it back when a newer beta exists, but..

... there is also a problem with blank lines in using groups with beta 406, which is hidden in the build since we bail out on warnings, but happens locally. I suspect that a change from 6 days ago would fix those problems, where we previously mapped dotnet_separate_import_directive_groups=false to blankLinesBetweenUsingGroups=omit instead of allow which we do now. So to fix that I would either have to remove the blank lines (again temporarily) or add blankLinesBetweenUsingGroups=allow to stylecop.json. I really would like to do neither of them.

So, what to do here? I would really prefer to get a new beta and update the reference for internal use, since we are quite a bit "behind" internally. What do you say @sharwell?

sharwell:
> Ok, I realized why it fails, although unit tests with the same code passes. There is 5 day old fix for generic delegate files names, which is not in the beta 406.

Yes, I created this fix because the update to 406 was problematic. We'll update to the one after 406 after it's ready.

sharwell:
1.2.0-beta.406 has some bugs that impact this upgrade. We'll wait for the next one. 👍

PR #3505, opened 2022-05-13
Update SA1000 to handle checked operator declarations correctly

### LINKED ISSUE #3478: SA1000 doesn't handle checked operator definitions
C# 11 is introducing user-defined checked operators (`public static T operator checked +(T value);`) and currently triggers `SA1000` saying no space is expected after `checked`.

The analyzer should be updated to allow a space when checked is used in this context, seeing as it differs from the normal `checked(expr)` syntax.

bjornhellander:
Fixes #3478

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1000KeywordsMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1000KeywordsMustBeSpacedCorrectly.cs
@@ -119,13 +119,23 @@ namespace StyleCop.Analyzers.SpacingRules
 
                 case SyntaxKind.CheckedKeyword:
                 case SyntaxKind.UncheckedKeyword:
-                    if (token.GetNextToken().IsKind(SyntaxKind.OpenBraceToken))
+                    switch (token.Parent.Kind())
                     {
+                    case SyntaxKind.CheckedStatement:
+                    case SyntaxKind.UncheckedStatement:
+                    case SyntaxKind.OperatorDeclaration:
+                    case SyntaxKind.ConversionOperatorDeclaration:
                         HandleRequiredSpaceToken(ref context, token);
-                    }
-                    else
-                    {
+                        break;
+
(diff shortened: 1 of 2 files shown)


bjornhellander:
Any comment on this, @sharwell?

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1000KeywordsMustBeSpacedCorrectly.cs
```diff
@@ -115,13 +115,22 @@ private static void HandleSyntaxTree(SyntaxTreeAnalysisContext context)
 
                 case SyntaxKind.CheckedKeyword:
                 case SyntaxKind.UncheckedKeyword:
-                    if (token.GetNextToken().IsKind(SyntaxKind.OpenBraceToken))
+                    switch (token.Parent.Kind())
                     {
+                    case SyntaxKind.CheckedStatement:
+                    case SyntaxKind.UncheckedStatement:
+                    case SyntaxKind.OperatorDeclaration:
```
sharwell, line 122: ❔ Does this need to support conversion operators as well?

bjornhellander: It does. Thank you! I didn't realize it was a different syntax kind. On its way.

bjornhellander: Updated

bjornhellander:
Any other comments on this, @sharwell? The Roslyn version needed for the tests in this change does not have the formatting problem that 4.4.0 has. Can this get merged before upgrading to a Roslyn with official c# 11 support?

bjornhellander:
@sharwell?

sharwell (approved):

sharwell:
@bjornhellander this looks good but needs to be rebased to fix a conflict

sharwell (approved):

PR #3507, opened 2022-05-14
Update SA1010 to not trigger on list patterns

### LINKED ISSUE #3503: SA1010 incorrectly reported on opening bracket in C# 11 list pattern
```cs
// ⚠️ SA1010 Opening square brackets should not be preceded by a space
//                 ↓
if (new[] { 1 } is [1])
{
}
```

SDK 6.0.200 or newer required. Example csproj:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <LangVersion>preview</LangVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
  </ItemGroup>

</Project>
```

bjornhellander:
Fixes #3503

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/SpacingRules/SA1010CSharp11UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/SpacingRules/SA1010CSharp11UnitTests.cs
@@ -3,9 +3,48 @@
 
 namespace StyleCop.Analyzers.Test.CSharp11.SpacingRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
     using StyleCop.Analyzers.Test.CSharp10.SpacingRules;
+    using StyleCop.Analyzers.Test.Verifiers;
+    using Xunit;
+
+    using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
+        StyleCop.Analyzers.SpacingRules.SA1010OpeningSquareBracketsMustBeSpacedCorrectly,
+        StyleCop.Analyzers.SpacingRules.TokenSpacingCodeFixProvider>;
 
     public class SA1010CSharp11UnitTests : SA1010CSharp10UnitTests
     {
+        [Theory]
+        [InlineData("x is [1]")]
+        [InlineData("x is not [1]")]
+        [InlineData("x is ([1] or [2])")]
+        [InlineData("x is ([1] or not [2])")]
+        [InlineData("x is ([1] and [1])")]
+        [InlineData("x is ([1] and not [2])")]
+        [WorkItem(3503, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3503")]
+        public async Task TestListPatternAsync(string condition)
+        {
+            var testCode = $@"
+using System.Collections.Generic;
+
+namespace TestNamespace
+{{
+    public class TestClass
+    {{
+        public void TestMethod(List<int> x)
+        {{
+            _ = {condition};
+        }}
+    }}
+}}
+";
+
+            await new CSharpTest()
+            {
+                ReferenceAssemblies = GenericAnalyzerTest.ReferenceAssembliesNet50,
+                TestCode = testCode,
+            }.RunAsync(CancellationToken.None).ConfigureAwait(false);
+        }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1010OpeningSquareBracketsMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1010OpeningSquareBracketsMustBeSpacedCorrectly.cs
@@ -11,6 +11,7 @@ namespace StyleCop.Analyzers.SpacingRules
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.Diagnostics;
     using StyleCop.Analyzers.Helpers;
+    using StyleCop.Analyzers.Lightup;
 
     /// <summary>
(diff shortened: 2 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/SpacingRules/SA1010CSharp11UnitTests.cs
```diff
@@ -5,9 +5,46 @@
...
+        }
+    }
+}
+";
+
+            await new CSharpTest()
+            {
+                ReferenceAssemblies = GenericAnalyzerTest.ReferenceAssembliesNet50,
+                TestCode = testCode,
+            }.RunAsync().ConfigureAwait(false);
```
sharwell, line 47:
```suggestion
            }.RunAsync(CancellationToken.None).ConfigureAwait(false);
```

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/SpacingRules/SA1010CSharp11UnitTests.cs
```diff
@@ -5,9 +5,46 @@
...
+    public class TestClass
+    {
+        public void TestMethod(List<int> x)
+        {
+            _ = x is [1];
+            _ = x is not [1];
+            _ = x is ([1] or [2]);
+            _ = x is ([1] or not [2]);
+            _ = x is ([1] and [1]);
+            _ = x is ([1] and not [2]);
```
sharwell, line 37: 💡 This should be converted to a code fix test, which includes examples of various incorrect formatting that is then corrected (there is no need to separately test valid syntax, since all code fix tests automatically verify that the fixed code contains no undeclared diagnostics).

bjornhellander: This specific analyzer only suggests to remove spaces, never to add them. I assume there is a reason for that, even though I haven't figured it out yet :-)

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/SpacingRules/SA1010CSharp11UnitTests.cs
bjornhellander: Fixed

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/SpacingRules/SA1010CSharp11UnitTests.cs
sharwell: Interesting. Based on other analyzers (e.g. `{`), SA1010 would be the analyzer that wants to add the space here.

bjornhellander: Since this seems to be in line with the current implementation, would it be ok to merge this if I create another issue to investigate the resposibility of this analyzer?

sharwell: 💭 Is this analyzer not responsible for reporting a this?

```
x in ( [1]);
       ^
```

In #3511 a very similar scenario does report a diagnostic.

bjornhellander: *Should* is a difficult word :-) The only diagnostic I would **like** to see in that example is SA1008. I am thinking that the list pattern's opening bracket itself should not trigger any opinion about preceding space, anymore than a literal value does. It is always something else that is causing the opinion, like a preceding open parenthesis, operator, keyword, etc. Before this PR, your example would trigger both SA1008 and SA1010. Is that a behaviour that you would like to keep? I could very well not be seeing the whole picture here, but I feel that not having an opinion at all is the correct behaviour for this rule in this specific case.

bjornhellander: I actually feel the same way about the property pattern changes made in SA1012, now that I think of it. I assume that both these statements will trigger two diagnostics:
```
_ = x is ( { P: 0 }, { P: 0 }) // SA1008 and SA1012
_ = x is [ { P: 0 }, { P: 0 }] // SA1010 and SA1012
```
With my thinking, only SA1008 and SA1010 should be triggered.

bjornhellander: Returning to SA1010 again, there was already a similar case implemented where this code only triggers SA1012 and not SA1010:
`var test = new Dictionary<ulong, string>() {[100] = "100" };`

bjornhellander:
I rebased this since it was getting old and turned the test into a theory to make it easier to debug

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/SpacingRules/SA1010CSharp11UnitTests.cs
bjornhellander: @sharwell?

sharwell (approved):

bjornhellander:
Thanks!

PR #3514, opened 2022-05-25
Use GetText instead of ToFullString

sharwell:
See dotnet/roslyn#61506


Neme12:
Wow, I'm happy that I was useful 😄

PR #3519, opened 2022-06-06
Performance improvement: Reduce time spent reading settings file

bjornhellander:
I profiled using StyleCopTester and noticed that a lot of time was spent reading the settings file. Previously a SourceTextValueProvider was used to cache the parsed json data, but that was no longer used. Updating SettingsHelper to again use a SourceTextValueProvider improved the time needed to retrieve the diagnostics in StyleCopTester by around 15% on my machine, when running on StyleCop.Analyzers.sln.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
@@ -26,10 +26,10 @@ namespace StyleCop.Analyzers
         public static void RegisterSyntaxTreeAction(this AnalysisContext context, Action<SyntaxTreeAnalysisContext, StyleCopSettings> action)
         {
             context.RegisterSyntaxTreeAction(
-                context =>
+                c =>
                 {
-                    StyleCopSettings settings = context.GetStyleCopSettings(context.CancellationToken);
-                    action(context, settings);
+                    StyleCopSettings settings = c.GetStyleCopSettings(c.CancellationToken);
+                    action(c, settings);
                 });
         }
 
@@ -42,10 +42,10 @@ namespace StyleCop.Analyzers
         public static void RegisterSyntaxTreeAction(this CompilationStartAnalysisContext context, Action<SyntaxTreeAnalysisContext, StyleCopSettings> action)
         {
             context.RegisterSyntaxTreeAction(
-                context =>
+                c =>
                 {
-                    StyleCopSettings settings = context.GetStyleCopSettings(context.CancellationToken);
-                    action(context, settings);
+                    StyleCopSettings settings = c.GetStyleCopSettings(c.CancellationToken);
+                    action(c, settings);
                 });
         }
 
@@ -81,10 +81,10 @@ namespace StyleCop.Analyzers
             where TLanguageKindEnum : struct
         {
             context.RegisterSyntaxNodeAction(
-                context =>
+                c =>
                 {
-                    StyleCopSettings settings = context.GetStyleCopSettings(context.CancellationToken);
-                    action(context, settings);
+                    StyleCopSettings settings = c.GetStyleCopSettings(c.CancellationToken);
+                    action(c, settings);
                 },
                 syntaxKinds);
         }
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
```diff
@@ -26,10 +26,10 @@ internal static class AnalyzerExtensions
         public static void RegisterSyntaxTreeAction(this AnalysisContext context, Action<SyntaxTreeAnalysisContext, StyleCopSettings> action)
         {
             context.RegisterSyntaxTreeAction(
-                context =>
+                c =>
```
sharwell, line 29: ❗ All context parameters should be named `context`. This prevents accidental references to an invalid context from a containing scope.

sharwell (changes requested):
Commit 337c1a02a8def13727f51a86011d4bc2b3e21ca0 needs to be removed from this pull request

StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
bjornhellander: SyntaxTreeAnalysisContext and SyntaxNodeAnalysisContext do not have the methods to access the cached values through the provider, so I need to be able to access the outer context (see 94afcb29368b72931e4d151aa7fdd5cde3431e5b). Would you prefer having both variables called context and storing the first one in a separate variable before it gets hidden by the inner context? Or did you have something else in mind?

bjornhellander: I see. Well that would be a real bummer :-) The cache actually sounds completely useless on AnalysisContext if that is true, but I might misunderstand things. Will try to confirm your understanding.

bjornhellander: What I meant was that all members in AnalysisContext are about registeration and configuration, except the TryGetValue method. Therefore I can see no way of getting a SourceText instance from that specific context. If I can not get hold of a SourceText instance through it, then I can not call TryGetValue either, provided that you are right about not being allowed to store contexts for later. That is why I wrote that the TryGetValue method is useless on that specific context class. I hope that was clearer.

bjornhellander: That aside, I definitely agree that your suggestion would work *if* we for example always went through RegisterCompilationStartAction to register the actions, since CompilationStartAnalysisContext actually has an AnalyzerOptions property that we can get the settings from, but it would affect a lot of analyzers, so I felt a bit hesitant about doing that initially. I can create a small, incomplete PR for this strategy, as a proof of concept.

manfred-brands: You can get AnalyzerOptions using the Options property from other contexts using `context.Options.AnalyzerOptionsConfigProvider.GetOptions`

bjornhellander: @manfred-brands If you are talking about AnalysisContext, there are no properties on it

manfred-brands: I was talking about the context passed in to the Syntax/Operation analyzer method: [SyntaxNodeAnalysisContext.Options](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnostics.syntaxnodeanalysiscontext.options?view=roslyn-dotnet-4.2.0)

bjornhellander:
Created PR #3523 with another approach which does not capture context instances. Considering this as on hold for the time being.

bjornhellander:
@sharwell I posted a question (https://github.com/dotnet/roslyn/issues/61823) about if it is ok to store contexts and got the the impression that it would be ok in this case. Do you have any comments on either of my two pull requests (this and #3523)?

bjornhellander:
Closing since the idea was not ok

PR #3529, opened 2022-06-19
Update SA1515 to not let one range of trivia affect another

### LINKED ISSUE #3481: Anomalous SA1515 with prior inline comment
With versions > 1.0.2 including 1.2.0-beta.406 I'm seeing anomalous behavior with SA1515 Single Line Comment Must Be Preceded By Blank Line.

In the example below I'm seeing SA1515 reported on the line `// Inline Comment 2`. Generally, there is an exception for this rule when comment follows documentation comments.

What's interesting, if either `// Inline Comment 1` or the constructor are removed from the class, the warning goes away. See the attached repo project's `Class2` and `Class3`.

The project does generate a documentation file:
`<GenerateDocumentationFile>true</GenerateDocumentationFile>`

Thanks

```
/// <summary>
/// Class1.
/// </summary>
public class Class1 // Inline Comment 1
{
    /// <summary>
    /// Initializes a new instance of the <see cref="Class1"/> class.
    /// </summary>
    public Class1()
    {
    }

    /// <summary>
    /// Gets value.
    /// </summary>
    // Inline Comment 2
    public double Value { get; }
}
```
[SA1515Repo.zip](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/files/8507291/SA1515Repo.zip)

bjornhellander:
Fixes #3481

The problem as I see it was that the previous code kept state to be able to trigger on the following code:
```
a(); // Some comment
// Diagnostic wanted here
```
without triggering on this:
```
// Some comment
// Diagnostic not wanted here
```

But this state also leaked over to unrelated chunks of trivia. I replaced the state with more digging in the current trivia, to avoid the reported false positive.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515SingleLineCommentMustBePrecededByBlankLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515SingleLineCommentMustBePrecededByBlankLine.cs
@@ -108,21 +108,18 @@ namespace StyleCop.Analyzers.LayoutRules
         private static void HandleSyntaxTree(SyntaxTreeAnalysisContext context)
         {
             var syntaxRoot = context.Tree.GetRoot(context.CancellationToken);
-            var previousCommentNotOnOwnLine = false;
 
             foreach (var trivia in syntaxRoot.DescendantTrivia().Where(trivia => trivia.IsKind(SyntaxKind.SingleLineCommentTrivia)))
             {
                 if (trivia.FullSpan.Start == 0)
                 {
                     // skip the trivia if it is at the start of the file
-                    previousCommentNotOnOwnLine = false;
                     continue;
                 }
 
                 if (trivia.ToString().StartsWith("////", StringComparison.Ordinal))
                 {
                     // ignore commented out code
-                    previousCommentNotOnOwnLine = false;
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515SingleLineCommentMustBePrecededByBlankLine.cs
```diff
@@ -174,7 +166,8 @@ private static bool IsOnOwnLine<T>(T triviaList, int triviaIndex)
         {
             while (triviaIndex >= 0)
             {
-                if (triviaList[triviaIndex].IsKind(SyntaxKind.EndOfLineTrivia))
+                var currentTrivia = triviaList[triviaIndex];
+                if (currentTrivia.IsKind(SyntaxKind.EndOfLineTrivia))
```
sharwell, line 170:
```suggestion
                if (triviaList[triviaIndex].IsKind(SyntaxKind.EndOfLineTrivia))
```

sharwell (approved):

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515SingleLineCommentMustBePrecededByBlankLine.cs
bjornhellander: Updated

sharwell (approved):

PR #3535, opened 2022-07-17
Update SA1206 to handle c# 11 modifier "required"

### LINKED ISSUE #3527: SA1206 firing incorrectly for `required` in C# 11
Version:
[1.2.0-beta.435](https://www.nuget.org/packages/StyleCop.Analyzers/1.2.0-beta.435)

Repro:
```C#
internal struct SomeStruct
{
    public required int Prop { get; set; }
}
```

Expected:
No warnings

Actual:
```
SA1206: The 'required' modifier should appear before 'public'
```

bjornhellander:
Fixes #3527

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/OrderingRules/SA1206CSharp11CodeFixProviderUnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/OrderingRules/SA1206CSharp11CodeFixProviderUnitTests.cs
@@ -6,6 +6,7 @@ namespace StyleCop.Analyzers.Test.CSharp11.OrderingRules
     using System.Threading;
     using System.Threading.Tasks;
     using StyleCop.Analyzers.Test.CSharp10.OrderingRules;
+    using StyleCop.Analyzers.Test.Verifiers;
     using Xunit;
     using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
         StyleCop.Analyzers.OrderingRules.SA1206DeclarationKeywordsMustFollowOrder,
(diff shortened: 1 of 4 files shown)


bjornhellander:
Rebased and added the missing tests, @sharwell

KUTlime:
@bjornhellander @sharwell How is this PR going? I have to write `#pragma`s more than I wish to. 🤷‍♀️

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1206DeclarationKeywordsMustFollowOrder.cs
```diff
@@ -96,6 +96,12 @@ private static void CheckModifiersOrderAndReportDiagnostics(SyntaxNodeAnalysisCo
             foreach (var modifier in modifiers)
             {
                 var currentModifierType = GetModifierType(modifier);
+                if (currentModifierType == ModifierType.None)
+                {
+                    // Just ignore unknown modifiers, to avoid false positives before support for
+                    // new language features has been added.
+                    continue;
+                }
```
sharwell, line 104: 💭 I almost prefer the original behavior, otherwise support may never have been added.

bjornhellander: I think that "never" is unlikely, but it would almost certainly have taken longer to be discovered. I was just trying to think about the user experience and for my part I would definitely prefer false negatives over false positives, but I can revert that change if you want me to. Just say the word.

KUTlime: @sharwell Could you please make you call here? Thank you.

bjornhellander: Ok. I reverted this change now.

bjornhellander:
Anything else, @sharwell?

StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1206DeclarationKeywordsMustFollowOrder.cs
POnakS: Imo behavior proposed by @bjornhellander is much better. We have required keyword for >6 months. This means users of stylecop either do not use SA1206 or do not use required keyword over this period.

KUTlime: @POnakS We do use it (_required keyword_) but we have to constantly disable SA1206 by pragma.

KUTlime:
Ground control to @sharwell, are you alive? @bjornhellander requested a feedback. We are waiting for this.

sharwell:
Getting hammered recently. Team is running on a light staff after some organizational changes.

sharwell (approved):

KUTlime:
> Getting hammered recently. Team is running on a light staff after some organizational changes.

I recommend _soft_ drink next time. They are call _soft_ for a reason. 😉 #Sarcasm

PR #3544, opened 2022-07-29
Update SA1137 to correctly fix a violation where the closing brace is not on a separate line

### LINKED ISSUE #3296: Fix for SA1137	Elements should have the same indentation	incorrectly applies for closing brace that's not at the start of line
Consider the following code:

```c#
class C
{
    void M()
    {
        _ = new[]
        {
            42 };
    }
}
```

The closing brace of the array initializer produces:

> SA1137	Elements should have the same indentation

That's reasonable, but then the suggested fix is:

```diff
         {
-            42 };
+            42         };
     }
```

I.e. the fixer adds the required number of spaces before the closing brace, but it does *not* add a newline, resulting in incorrectly formatted code (which then causes SA1025	Code should not contain multiple whitespace characters in a row).

bjornhellander:
Fixes #3296

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
@@ -526,9 +526,22 @@ namespace StyleCop.Analyzers.ReadabilityRules
             SyntaxTrivia closeBraceIndentationTrivia = closeBraceToken.LeadingTrivia.LastOrDefault();
             string closeBraceIndentation = closeBraceIndentationTrivia.IsKind(SyntaxKind.WhitespaceTrivia) ? closeBraceIndentationTrivia.ToString() : string.Empty;
 
-            if (!string.Equals(openBraceIndentation, closeBraceIndentation, StringComparison.Ordinal))
+            if (closeBraceToken.IsOnlyPrecededByWhitespaceInLine())
             {
-                ReportDiagnostic(context, closeBraceToken, closeBraceIndentationTrivia, closeBraceIndentation, openBraceIndentation);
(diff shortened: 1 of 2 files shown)


bjornhellander:

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
```diff
@@ -526,9 +526,22 @@ private static void CheckBraces(SyntaxNodeAnalysisContext context, SyntaxToken o
...
+            }
+            else
+            {
+                // In this case, the closing brace does not seem to have any leading trivia,
+                // but the previous token can have trailing trivia.
+                SyntaxTrivia prevTokenTrailingTrivia = closeBraceToken.GetPreviousToken().TrailingTrivia.FirstOrDefault();
+                string prevTokenTrailingWhitespace = prevTokenTrailingTrivia.IsKind(SyntaxKind.WhitespaceTrivia) ? prevTokenTrailingTrivia.ToString() : string.Empty;
+                string expectedCloseBraceIndentation = Environment.NewLine + openBraceIndentation;
+
+                ReportDiagnostic(context, closeBraceToken, prevTokenTrailingTrivia, prevTokenTrailingWhitespace, expectedCloseBraceIndentation);
```
sharwell, line 544: It seems like SA1137 should not be reported here at all. We already get SA1500 here, and that has the correct fix.

sharwell (changes requested):

StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs
bjornhellander: I see what you mean. Didn't think of that. Alright, so I'm closing this PR.

PR #3548, opened 2022-08-10
Update SA1407 to also handle % as a multiplicative operator

### LINKED ISSUE #3545: SA1407 / IDE0047 conflict
There's currently no way to avoid both SA1407 or IDE0047 from getting reported for code without precedence issue.

For example:

The following line results in **SA1407** (Arithmetic expressions should declare precedence):
```cs
var x = 10 / 2 % 5;
```

If you add parentheses then **IDE0047** (Parentheses can be removed) is reported:
```cs
var x = (10 / 2) % 5;
```

I reported this as a [bug](https://github.com/dotnet/roslyn/issues/63024) in the .NET Compiler Platform (roslyn) analyzers, but it was rightfully considered as correct behavior.

Is it possible to update the implementation of **SA1407** to no longer report a diagnostic when there's no precedence issue?

bjornhellander:
Fixes #3545

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1407UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/MaintainabilityRules/SA1407UnitTests.cs
@@ -15,87 +15,107 @@ namespace StyleCop.Analyzers.Test.MaintainabilityRules
 
     public class SA1407UnitTests
     {
-        [Fact]
-        public async Task TestAdditionAndSubtractionAsync()
+        [Theory]
+        [InlineData("+", "+", "+")]
+        [InlineData("-", "+", "-")]
+        [InlineData("-", "-", "-")]
+        [InlineData("+", "-", "+")]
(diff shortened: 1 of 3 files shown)


sharwell:
@drieseng Based on the amount of time these two rules have conflicted and the minimal number of times it's been reported, this doesn't appear to be a particularly common coding pattern. I think my current recommendation would be to either rewrite the line to avoid having this expression in one line, or suppress one of the diagnostics at this location.

bjornhellander:
I tested IDE0047 on a new .NET 6 project with default settings:
**IDE0047 is not shown in the error window and there no squigglies shown either**. The only time it is shown is if I hover over one of the parenthesis.

Given that behaviour and the relative "un-commonness" of the remainder operator, I am not surprised at all that this has not been reported before. Are you concerned that this pull request might not do what it is intended to do, @sharwell, or why do you hesitate to merge this?

sharwell:
Resolving this one as **Won't Fix**. In the context of the examples presented in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3548#issuecomment-1246365645, the current behavior is desired, and is intentionally included in this form in tests.

PR #3580, opened 2022-11-27
Update Microsoft.CodeAnalysis.CSharp.Workspaces to version 4.4.0 for the c# 11 test project

bjornhellander:
This is the Roslyn version used in Visual Studio 17.4, with official support for c# 11.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/StyleCop.Analyzers.Test.CSharp11.csproj
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/StyleCop.Analyzers.Test.CSharp11.csproj
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


bjornhellander:
Failing the same way as #3535 with only an upgrade of Roslyn.
Also see https://github.com/dotnet/roslyn/issues/63013 for bug filed on Roslyn.

bjornhellander:
How do you want to handle this regression @sharwell? Would it be ok to temporarily accept this bad code fix behaviour as the expected behaviour just in the c#11 test project, to be able to move forward with other c#11 fixes?

sharwell (dismissed):

bjornhellander:
@sharwell, would it be ok to temporarily skip the three failing tests for c# 11, to be able to move forward with other c#11 fixes which also require a newer Roslyn?

bjornhellander:
@sharwell?

sharwell:
I'm going to try and get 4.4.1 packages published with a fix for the regression seen here.

sharwell (approved):
The test issue would not have been fixed by a 4.4.1 package. I filed dotnet/roslyn#66779 for the underlying issue, and applied a workaround to this pull request.

Issue #3581, opened 2022-11-27
SA1503 code fix indents incorrectly with custom indentation size


bjornhellander:
Consider the following file:
```
namespace TestNamespace
{
 public class TestClass
 {
  public void TestMethod(int i)
  {
   if (i == 0)
    return;
  }
 }
}
```

If this is included in a project with settings.indentation.indentationSize=1, then the code fix for SA1503 provides the following incorrect fix:
```
namespace TestNamespace
{
 internal class TestClass
 {
  public void TestMethod(int i)
  {
   if (i == 0)
            {
                return;
            }
        }
 }
}
```

StyleCop.Analyzers 1.2.0-beta.435.
Visual Studio 17.4 (Roslyn 4.4.0). Does not happen in earlier versions.
https://github.com/dotnet/roslyn/issues/63013 filed on earlier preview version of Roslyn, but the same applies for 4.4.0.
PR #3580, created to bump Roslyn version to 4.4.0 in the c#11 test project, fails because of the same problem.

PR #3590, opened 2022-12-18
Update SA1400 to recognize access modifier "file"

### LINKED ISSUE #3588: SA1400 false positive on access modifier "file"
The following code triggers SA1400, but should not:
```
file class TestClass
{
}
```

Tested with version 1.2.0-beta.435

bjornhellander:
Fixes #3588


bjornhellander:
This change requires a Roslyn version new enough to also trigger the indentation problem from #3581, so marked this as "on hold" for now.

sharwell (approved):

PR #3591, opened 2022-12-18
Update SA1206 to recognize modifier "file"

### LINKED ISSUE #3589: SA1206 does not offer code fix when access modifier "file" is in incorrect place
The following code triggers SA1206 (as it should), but no code fix is offered:
```
abstract file class TestClass
{
}
```

Tested with version 1.2.0-beta.435

bjornhellander:
Fixes #3589


bjornhellander:
This change requires a Roslyn version new enough to also trigger the indentation problem from https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3581, so marked this as "on hold" for now.

sharwell (approved):

PR #3607, opened 2023-02-25
Update SA1513 codefix to use the existing newline character sequence

### LINKED ISSUE #3360: Applying autofixes uses wrong line endings / should honour .editorconfig
`.editorconfig`:
```
[*.cs]
end_of_line = lf
```

`Example.cs`:
```c#
namespace Bug
{
    public static class Example
    {
        public static int MissingNewLine(bool condition)
        {
            if (condition)
            {
                return 42;
            } // SA1513: Closing brace should be followed by a blank line
            return 0;
        }
    }
}
```

Applying the automatic fix for SA1513 (and probably a bunch of others, but this is the one where I tracked down the problem) inserts a CRLF (carriage return + line feed) combination, but the file uses LF only. In Visual Studio, you can observe the state of line endings for the current file in the bottom right corner of the file-editor pane. It should initially say `LF`, when the automatic fix is applied it will switch to `MIXED` instead:
image

The next time the file is opened (but not immediately when the problem is introduced!), Visual Studio will complain about mixed line endings, and suggest to 'normalize' the line endings (using the wrong default option, of-course):
image

The root-cause is probably because [the fixer appears to use a hard-coded CRLF](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/82c2d90907131d939fe396e4191b3a867650615f/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/LayoutRules/SA1513CodeFixProvider.cs#L66) ([SyntaxFactory.CarriageReturnLineFeed](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntaxfactory.carriagereturnlinefeed?view=roslyn-dotnet-3.9.0)) rather than the currently configured line-end character.

bjornhellander:
Fixes #3360

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
@@ -9,6 +9,7 @@ namespace StyleCop.Analyzers.Test.LayoutRules
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.LayoutRules;
+    using StyleCop.Analyzers.Test.Helpers;
     using Xunit;
     using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
         StyleCop.Analyzers.LayoutRules.SA1513ClosingBraceMustBeFollowedByBlankLine,
@@ -1012,6 +1013,26 @@ public class TestClass
 ";
 
             await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        [WorkItem(3360, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3360")]
+        public async Task TestLineFeedEndOfLinesAsync()
+        {
+            var testCode = @"
+public class TestClass
+{
+}[|
+|]// Hello".ReplaceLineEndings("\n");
+
+            var fixedCode = @"
+public class TestClass
+{
+}
+
+// Hello".ReplaceLineEndings("\n");
+
+            await VerifyCSharpFixAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, fixedCode, CancellationToken.None).ConfigureAwait(false);
         }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
@@ -175,5 +175,35 @@ namespace StyleCop.Analyzers.Helpers
             triviaList = token.GetNextToken().LeadingTrivia;
             return triviaList.Count > 0 && triviaList.First().IsKind(SyntaxKind.WhitespaceTrivia);
         }
+
+        /// <summary>
+        /// Returns the closest end of line trivia preceding the <paramref name="token"/>.
+        /// This currently only looks immediately before the specified token.
+        /// </summary>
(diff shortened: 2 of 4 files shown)


bjornhellander:
Did you have something like this in mind, @sharwell?

I can imagine that some other fixers would like to use a new line which is *after* a specified token instead and some probably just have a span to go on. A simpler approach could be to just find the first newline in the document and use that.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
```diff
@@ -1013,5 +1013,29 @@ public void TestMethod(string extraSupport)
...
+
+        [Fact]
+        [WorkItem(3360, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3360")]
+        public async Task TestLineFeedEndOfLinesAsync()
+        {
+            var testCode =
+                "public class TestClass\n" +
+                "{\n" +
+                "}\n" +
+                "// Hello";
```
sharwell, line 1025: 💡 You can define a `ReplaceLineEndings` helper like this:
https://github.com/dotnet/roslyn/blob/f64507a8d6b6c3e01f8d1e05ce97552f3c37b341/src/Compilers/Test/Core/FX/StringExtensions.cs

This allows you to use verbatim strings like all the other tests, and just call the extension method to assign the line endings you want.

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
```diff
@@ -1013,5 +1013,29 @@ public void TestMethod(string extraSupport)
...
+            var fixedCode =
+                "public class TestClass\n" +
+                "{\n" +
+                "}\n" +
+                "\n" +
+                "// Hello";
+
+            DiagnosticResult[] expected =
+            {
+                Diagnostic().WithLocation(3, 2),
```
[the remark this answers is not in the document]
bjornhellander: Will do

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1513UnitTests.cs
bjornhellander: Absolutely. Would you like to keep the "fancy" conversions, or just assume CRLF in the input?

bjornhellander:
Updated, @sharwell

bjornhellander:
Any other comments, @sharwell?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
```diff
@@ -175,5 +175,35 @@ internal static bool IsFollowedByWhitespace(this SyntaxToken token)
...
+            var leadingTrivia = token.LeadingTrivia;
+            for (var i = leadingTrivia.Count - 1; i >= 0; i--)
+            {
+                if (leadingTrivia[i].IsKind(SyntaxKind.EndOfLineTrivia))
+                {
+                    return leadingTrivia[i];
+                }
+            }
+
+            var prevToken = token.GetPreviousToken();
```
sharwell, line 196: 💡 If there is any way to test this with a preceding `#region` directive, that would be good (preprocessor directives have embedded line endings that are not EndOfLineTrivia). Will leave this for future improvement though.

sharwell (approved):

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/TokenHelper.cs
bjornhellander: It looks like this rule never triggers when the closing brace is followed by a directive, so I don't see a way to test that case with this rule. But if this method will be used in other code fixers as well, then it seems reasonable to do.

Issue #3609, opened 2023-03-05
Convert SA1141 (Use Tuple Syntax) to an IOperation analyzer


sharwell:
I received a report of heavy CPU impact from this analyzer calling `GetSymbolInfo`. This would be resolved by converting the analyzer to an IOperation-based analyzer.

PR #3610, opened 2023-03-06
Optimize SA1141

sharwell:
* Avoid all work on compilations that don't support tuples (not a common case)
* Check identifier names before calling `GetSymbolInfo`

This is a partial resolution of the issue that led to #3609. Note that the new behavior for handling `ValueTuple.Create` will not work correctly in cases where users have a using alias directive like the following:

```cs
using VT = System.ValueTuple;
```


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp10/ReadabilityRules/SA1141CSharp10UnitTests.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
```
bjornhellander, line 1: Good catch with these missing test classes. I assume it was a mistake to not add them earlier. Would be good to have automated checks for things like this.

PR #3617, opened 2023-03-18
Update reading of file_header_template and stylecop.documentation.copyrightText to allow multiple lines

### LINKED ISSUE #3605: Support file_header_template
It seems to have problems with multi-line headers.

When you have something like:

```
file_header_template = this\nspans\nmultiple\nlines
```

the warning will be that it expects it to be all in one line.

bjornhellander:
Fixes #3605

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Settings/ObjectModel/AnalyzerConfigHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Settings/ObjectModel/AnalyzerConfigHelper.cs
@@ -51,6 +51,12 @@ namespace StyleCop.Analyzers.Settings.ObjectModel
             return null;
         }
 
+        internal static string TryGetMultiLineStringValue(AnalyzerConfigOptionsWrapper analyzerConfigOptions, string key, bool allowExplicitUnset = true)
+        {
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/Settings/ObjectModel/AnalyzerConfigHelper.cs
```diff
@@ -51,6 +51,12 @@ internal static string TryGetStringValue(AnalyzerConfigOptionsWrapper analyzerCo
             return null;
         }
 
+        internal static string TryGetMultiLineStringValue(AnalyzerConfigOptionsWrapper analyzerConfigOptions, string key, bool allowExplicitUnset = true)
+        {
+            var orgValue = TryGetStringValue(analyzerConfigOptions, key, allowExplicitUnset);
+            return orgValue?.Replace("\\r", "\r")?.Replace("\\n", "\n");
```
sharwell, line 57: Only the first `?.` needs to handle null, right?

```suggestion
            return orgValue?.Replace("\\r", "\r").Replace("\\n", "\n");
```

sharwell (dismissed):

sharwell (approved):

PR #3632, opened 2023-04-07
Update StyleCopTester project so it is able to load this solution

bjornhellander:
It did not start before due to assembly load exceptions. Unsure of what caused this.
I actually still get load exceptions occasionally, even with these changes, but a rebuild usually fixes it again.

I also added anther launch profile without the `/stats` option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ OpenCover.Symbols/
 .nuget/NuGet.exe
 build/nuget/
 *.log
+*.binlog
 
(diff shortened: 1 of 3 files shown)


sharwell (dismissed):

bjornhellander:
Fixed merge conflict

bjornhellander:
Closing this, since it's no longer enough to get it running. Current problems seem to be related to the resx generation.

PR #3639, opened 2023-04-15
Update NamingSettings and DocumentationSettings to keep one Regex instance instead of calling Regex.IsMatch

### LINKED ISSUE #3629: Regex.IsMatch(string, string) inside NamingSettings.ctor is causing 15.1% allocations
Building a large project, found Regex.IsMatch(string, string) inside NamingSettings.ctor is causing 15.1% allocations, mostly due to constructing regular expression repeatitively. At least the regular expressions should be reused.
image

bjornhellander:
Fixes #3629

Replaced the only remaining calls to the static Regex.IsMatch method.


sharwell (changes requested):
These both seem easy to just inline. Can we do that instead?

bjornhellander:
Removed use of Regex completely instead

sharwell (approved):

bjornhellander:
Thanks!

Issue #3634, opened 2023-04-10
Improve performance when handling settings in analyzers


bjornhellander:
I profiled using StyleCopTester and noticed that a lot of time was spent reading the settings file. Previously, a SourceTextValueProvider was used to cache the parsed json data, but that is no longer used.

The following optimizations looks reasonable:
1) Cache the parsed json representation of settings files.
2) Cache complete StyleCopSettings objects, with data from the settings json file and .editorconfig.

bjornhellander:
Closing since the remaining last part was handled in #3885

Issue #2975, opened 2019-07-02
Investigating analyzer impact on build times


peteroupc:
StyleCop Classic used to write information about files it analyzed, including analysis results, in an XML cache called `StyleCop.Cache`, to speed the time it takes to analyze a project if some of the source code files are unchanged in the meantime.  Although StyleCop Analyzers also has caches, they are limited to storing `using` aliases and copyright texts, but nothing like a results cache like StyleCop Classic used to have.

In my experience, running the StyleCop analyzer takes a considerable part of a project's build time even if it ultimately finds few or no violations, and I feel that having a results cache (like `StyleCop.Cache`) will speed up this time because StyleCop would then focus analysis on only those files that have changed, which is often not many files.

sharwell:
@peteroupc StyleCop Analyzers is not StyleCop Classic. If you find an indication of a performance problem, please file an issue and we will work to resolve it directly.

peteroupc:
I can't speak of the slow performance of StyleCop Analyzers except in general terms.  I know that StyleCop Analyzers is more than doubling build time of my projects but I don't know where to start in finding a specific "performance problem" of the kind you have in mind.

sharwell:
What was the build time prior to StyleCop Analyzers, and the build time after?

Do you have *any* other analyzers installed in your project?

peteroupc:
For one project of mine, the build time was&mdash;

- about 1:06 without StyleCop Analyzers, and
- about 2:54 with StyleCop Analyzers,

resulting in almost triple the build time.  This project used no other analyzers than StyleCop Analyzers.

sharwell:
@peteroupc The last time I measured it, the performance cost of StyleCop Analyzers was about 90% in the analyzer driver (enabled whenever any analyzer is installed) and about 10% StyleCop Analyzers itself. This project only has control over the performance of the latter. The easiest way to see if this is impacting you is to install another analyzer, e.g. FXCop Analyzers, and then re-run your comparison between StyleCop Analyzers installed or not. In this scenario, I would expect the difference between the two cases to be much less than 1:48.

The other thing that can impact performance is the number of diagnostics reported by the analyzers. This is the same as the number of diagnostics you see if the project has *no* suppressions. StyleCop Analyzers is optimized for cases where no diagnostics are reported. A few diagnostics wouldn't make much difference, but if are seeing and/or suppressing thousands of diagnostics, it would have a noticeable impact.

[...]

PR #3642, opened 2023-04-18
Update AnalyzersExtensions and SettingsHelper to use cached JsonValue objects where possible

bjornhellander:
Partial fix for #3634

The primary change is to use the SourceTextValueProvider cache in CompilationStartAnalysisContext to reduce the number of times the settings file is parsed. Code fix providers and some tests still use overloads of GetStyleCopSettings which does not use the cache.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
@@ -8,6 +8,7 @@ namespace StyleCop.Analyzers
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Immutable;
+    using System.Diagnostics.CodeAnalysis;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.Diagnostics;
     using StyleCop.Analyzers.Settings.ObjectModel;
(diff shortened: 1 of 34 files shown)


bjornhellander:
Seems to also be a partial fix for the SA1121 slowness reported in #2975.

bjornhellander:
Do you have time to review this, @sharwell?

I would also like to cache complete StyleCopSettings objects, since merging the json content with the analyzer options takes considerable time as well, but I am holding that off until this one is merged to make it a bit easier to review.

StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
```diff
@@ -23,72 +24,19 @@ internal static class AnalyzerExtensions
         /// </summary>
         /// <param name="context">The analysis context.</param>
         /// <param name="action">Action to be executed at completion of parsing of a document.</param>
-        public static void RegisterSyntaxTreeAction(this AnalysisContext context, Action<SyntaxTreeAnalysisContext, StyleCopSettings> action)
```
sharwell, line 26: 📝 So this is a bit of a strange situation. The `AnalysisContext.TryGetValue<TValue>` method is effectively static - it doesn't access any state of `AnalysisContext`. In theory, you could do something like this:

```csharp
delegate bool TryGetStyleCopSettingsFileJson(SourceText text, SourceTextValueProvider<Lazy<JsonValue>> valueProvider, out Lazy<JsonValue> value);

// This delegate can be captured and used in the callback
TryGetStyleCopSettingsFileJson tryGetValue = context.TryGetValue<Lazy<JsonValue>>;
```

sharwell: For cases where more than one callback is registered, it does seem like it would be more efficient to use an explicit start callback and only compute it once.

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
```diff
@@ -167,5 +228,39 @@ private static StyleCopSettings GetSettings(AnalyzerOptions options, SyntaxTree
...
+                    return new SettingsFile(additionalFile.Path, content);
+                }
+            }
+
+            return null;
+        }
+
+        private static Lazy<JsonValue> ParseJson(SourceText text)
+        {
+            return new Lazy<JsonValue>(() => JsonReader.Parse(text.ToString()));
```
sharwell, line 250: 💡 Would be easy to use [`SourceTextReader`](https://github.com/dotnet/roslyn-analyzers/commit/4369dd47748557d2e56161a783c749a020c7e83a) to avoid the call to `ToString()` here

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
```diff
@@ -86,7 +134,22 @@ internal static StyleCopSettings GetStyleCopSettings(this AnalyzerOptions option
         /// <returns>A <see cref="StyleCopSettings"/> instance that represents the StyleCop settings for the given context.</returns>
         internal static StyleCopSettings GetStyleCopSettings(this CompilationAnalysisContext context, SyntaxTree tree, DeserializationFailureBehavior failureBehavior, CancellationToken cancellationToken)
         {
-            return GetSettings(context.Options, tree, failureBehavior, cancellationToken);
+            var settingsFile = GetSettingsFile(context.Options, GetJsonValue, cancellationToken);
+            return GetSettings(context.Options, tree, settingsFile, failureBehavior);
+
+            Lazy<JsonValue> GetJsonValue(SourceText settingsText)
```
sharwell, line 140: 💡 It would be better to make this a `static` lambda in the call to `GetSettingsFile` so it can cache the delegate

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
```diff
@@ -25,23 +27,67 @@ internal static class SettingsHelper
...
+        /// <param name="context">The context that will be used to determine the StyleCop settings.</param>
+        /// <param name="cancellationToken">The cancellation token that the operation will observe.</param>
+        /// <returns>A <see cref="SettingsFile"/> instance which contains information about the the StyleCop settings file for the given context.
+        /// Null if no settings file was found.</returns>
+        [SuppressMessage("MicrosoftCodeAnalysisPerformance", "RS1012:Start action has no registered actions", Justification = "This is not a start action")]
+        internal static SettingsFile GetStyleCopSettingsFile(this CompilationStartAnalysisContext context, CancellationToken cancellationToken)
+        {
+            return GetSettingsFile(context.Options, GetJsonValue, cancellationToken);
+
+            Lazy<JsonValue> GetJsonValue(SourceText settingsText)
```
sharwell, line 46: 💡 It would be better to make this a `static` lambda in the call to `GetSettingsFile` so it can cache the delegate

sharwell (approved):

sharwell:
@bjornhellander The suggestions above can be implemented as part of the follow-up work

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
bjornhellander: You lost me :-) This local function captures the context. I hadn't heard of static lambdas before, but it seems the context makes it impossible for it to be static. I am probably misunderstanding you (or static lambdas). Could you elaborate, @sharwell?

sharwell: Suppose you need to capture a local variable, like this:

```cs
int specialValue = 5;
InvokeHelper(() => specialValue);

// helper
int InvokeHelper(Func<int> calculateValue)
  => -calculateValue();
```

You can avoid the delegate allocation by making a generic state argument, and then making the lambda `static`:

```cs
int specialValue = 5;
InvokeHelper(static specialValue => specialValue, specialValue);

// helper
int InvokeHelper<TArg>(Func<TArg, int> calculateValue, TArg arg)
  => -calculateValue(arg);
```

bjornhellander: Ah, got it! So if I change GetSettingsFile to accept a generic state/context parameter the way you suggest, would there be any difference between a static lambda and making the existing local function static? I just feel that a local function is better for readability.

sharwell: I'm not sure. We'd need to look at the IL output from the compiler to verify.

bjornhellander: I am not fluent in IL, but it looks like a branch around the allocation if it has already been performed. Similar code for both static cases.
https://sharplab.io/#v2:EYLgtghglgdgNAFxFANnAJiA1AHwAIBMAjALABQeAzAASHUDC1A3udW7TXgCzUBiA9vwAUsBNQAeASmqt2LMu0XUA+tQC81ACLDxcWkQBs1AJ7qAfCckBuWWwC+tmQvZVaPAEIQATiJhipTorySuyqGtpCutQAMtaOivEuhtSiMb5ixtJqFsY2zvaOjq4AKlr8ADzFZkKlAG4QKACuAKZ6eACslXpV1BAAxghQ/DCSjsEhtADsvQNDMEL1Tc1x+dQOZHZAA=

StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
bjornhellander: The big problem right now is re-creating the settings object over and over. Holding on to these two comments for later, when most of the work is done.

bjornhellander: > For cases where more than one callback is registered, it does seem like it would be more efficient to use an explicit start callback and only compute it once.

Handled in #3648

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
bjornhellander: Handled in #3648

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
bjornhellander: Handled in #3648

StyleCop.Analyzers/StyleCop.Analyzers/Settings/SettingsHelper.cs
bjornhellander: Handled in #3648

PR #3648, opened 2023-04-27
Cleanup in SettingsHelper and add caching of StyleCopSettings objects

bjornhellander:
Partial fix for #3634.
Handles some review comments from PR #3642.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/AnalyzerExtensionsTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/AnalyzerExtensionsTests.cs
@@ -74,7 +74,7 @@ class TypeName
                 context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
                 context.EnableConcurrentExecution();
 
-                context.RegisterCompilationStartAction(
+                context.RegisterCompilationStartActionWithSettings(
                     compilationStartContext =>
                     {
                         compilationStartContext.RegisterSyntaxNodeAction(
--- a/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs
@@ -8,7 +8,6 @@ namespace StyleCop.Analyzers
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Immutable;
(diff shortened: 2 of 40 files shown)


bjornhellander:
I had a look at https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3642#discussion_r1175637049 and started trying out some ways to optimize the case where more than one action is registered in the same analyzer. I thought I had a pretty trivial change, but a lot of tests for SA1516 fail and I haven't been abe to find the root cause yet. The actions are simply not called. And these are the only tests failing, at least for c# 6 which I have been running. I see that the register calls that are relevant in the failing cases is to a new method that I added, but it is used in 27 other places as well and the tests for those analyzers all pass.

My idea was to create a custom version of RegisterCompilationStartAction and have that one lookup the settings from the cache. That way, each analyzer wouldn't need to do it themselves as I interpreted your comment to suggest, but instead keep it hidden inside for example the RegisterSyntaxNodeAction extension method.

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/77555fa92263b672038f1cb58c3066549e18a06e/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs#L20-L28

The new register method:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/77555fa92263b672038f1cb58c3066549e18a06e/StyleCop.Analyzers/StyleCop.Analyzers/AnalyzerExtensions.cs#L100-L104

The first and last calls to RegisterSyntaxNodeAction in SA1516 tries to register the actions (without settings object as parameter) that never get called:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/77555fa92263b672038f1cb58c3066549e18a06e/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1516ElementsMustBeSeparatedByBlankLine.cs#L122-L134

@sharwell I pushed anyway in case you or anyone else can spot any mistakes in the last commit which could explain the failing SA1516 tests. Or just stop me if you think this is a bad idea.

bjornhellander:
Never mind. I found the mistake. Updates are coming.

bjornhellander:
Ok, so I continued with the actual StyleCopSettings caching as well. With these changes, the time spent in these analyzers are down another 30% in my benchmarks, down a total of 50% since I started looking at this. SA1121 is no longer in the top 10. Ready for review.

bjornhellander:
Do you have time to review this, @sharwell?

PR #3651, opened 2023-05-02
Make XmlCommentHelper faster

ninedan:
1) Use StringBuilder only when there are multiple lines
2) In white space normalization, return original string when no change is needed.
3) When changes are really needed, calculate exact size, allocate a char buffer for the changed text, then allocate string.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/ObjectPools/StringBuilderPool.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/ObjectPools/StringBuilderPool.cs
@@ -22,8 +22,11 @@ namespace StyleCop.Analyzers.Helpers.ObjectPools
 
         public static string ReturnAndFree(StringBuilder builder)
         {
-            SharedPools.Default<StringBuilder>();
-            return builder.ToString();
+            string result = builder.ToString();
+
+            StringBuilderPool.Free(builder);
+
+            return result;
         }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
@@ -317,6 +317,10 @@ namespace StyleCop.Analyzers.Helpers
         /// <summary>
         /// Append to StringBuilder and perform white space normalization.
         /// </summary>
+        /// <param name="builder">StringBuilder to append to.</param>
+        /// <param name="text">String to append.</param>
+        /// <param name="normalizeWhitespace">Normalize flag.</param>
+        /// <param name="lastWhitespace">last char is white space flag.</param>
         /// <returns>True if output is different.</returns>
         internal static bool AppendNormalize(this StringBuilder builder, string text, bool normalizeWhitespace, ref bool lastWhitespace)
         {
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
@@ -301,6 +301,7 @@ namespace StyleCop.Analyzers.Helpers
                     {
                         StringBuilderPool.Free(stringBuilder);
 
+                        // No change is needed, return original string.
                         return single;
                     }
                 }
@@ -313,6 +314,10 @@ namespace StyleCop.Analyzers.Helpers
             return StringBuilderPool.ReturnAndFree(stringBuilder);
         }
 
+        /// <summary>
+        /// Append to StringBuilder and perform white space normalization.
+        /// </summary>
+        /// <returns>True if output is different.</returns>
         internal static bool AppendNormalize(this StringBuilder builder, string text, bool normalizeWhitespace, ref bool lastWhitespace)
         {
             bool diff = false;
--- a/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
@@ -268,20 +268,112 @@ namespace StyleCop.Analyzers.Helpers
                 return null;
             }
 
-            StringBuilder stringBuilder = StringBuilderPool.Allocate();
+            string result = string.Empty;
+
+            StringBuilder stringBuilder = null;
 
             foreach (var item in textElement.TextTokens)
             {
-                stringBuilder.Append(item);
+                if (result.Length == 0)
+                {
+                    result = item.ToString();
+                }
+                else
+                {
+                    if (stringBuilder == null)
+                    {
+                        stringBuilder = StringBuilderPool.Allocate();
+                        stringBuilder.Append(result);
+                    }
+
+                    stringBuilder.Append(item.ToString());
+                }
             }
 
-            string result = StringBuilderPool.ReturnAndFree(stringBuilder);
+            if (stringBuilder != null)
+            {
+                result = StringBuilderPool.ReturnAndFree(stringBuilder);
+            }
+
             if (normalizeWhitespace)
             {
-                result = Regex.Replace(result, @"\s+", " ");
+                result = result.NormalizeWhiteSpace();
             }
 
             return result;
+        }
+
+        internal static string NormalizeWhiteSpace(this string text)
+        {
+            if (text == null)
+            {
+                return null;
+            }
+
+            int length = text.Length;
+
+            if (length == 0)
+            {
+                return string.Empty;
+            }
+
+            bool lastSpace = false;
+
+            bool diff = true;
+
+            foreach (char ch in text)
+            {
+                if (char.IsWhiteSpace(ch))
+                {
+                    if (lastSpace)
+                    {
+                        length--;
+                    }
+                    else
+                    {
+                        if (ch != ' ')
+                        {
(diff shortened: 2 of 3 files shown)


rachit-ranjan16 (approved):

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
```diff
@@ -268,22 +267,109 @@ internal static string GetText(XmlTextSyntax textElement, bool normalizeWhitespa
...
+            if (stringBuilder != null)
+            {
+                result = StringBuilderPool.ReturnAndFree(stringBuilder);
             }
 
-            string result = StringBuilderPool.ReturnAndFree(stringBuilder);
             if (normalizeWhitespace)
             {
-                result = Regex.Replace(result, @"\s+", " ");
+                result = result.NormalizeWhiteSpace();
```
danmoseley, line 299: just curious, @stephentoub on the latest .NET (of course this is broader targeted) would you expect regex replace using pre generated regex to have comparable performance to open coded?

I didn't measure, but fwiw 
```c#
        [GeneratedRegex(@"[\s^ ]+")]
        private static partial Regex GetRegex();
```
generates code in part like
```c#
			private bool TryFindNextPossibleStartingPosition(ReadOnlySpan<char> inputSpan)
			{
				int pos = runtextpos;
				if ((uint)pos < (uint)inputSpan.Length)
				{
					ReadOnlySpan<char> span = inputSpan.Slice(pos);
					for (int i = 0; i < span.Length; i++)
					{
						char ch;
						if (((ch = span[i]) < '\u0080') ? ((byte)("㸀\0\u0001\0\0䀀\0\0"[(int)ch >> 4] & (1 << (ch & 0xF))) != 0) : RegexRunner.CharInClass(ch, "\0\u0004\u0001 !^_d"))
						{
							runtextpos = pos + i;
							return true;
						}
					}
				}
				runtextpos = inputSpan.Length;
				return false;
```
the code below hides some of this inside `char.IsWhiteSpace`

danmoseley: and will return the same string if there are no non-space whitespace chars.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
```diff
@@ -268,22 +267,109 @@ internal static string GetText(XmlTextSyntax textElement, bool normalizeWhitespa
...
+                }
+                else
+                {
+                    lastSpace = false;
+                }
+            }
+
+            if (diff || (length != text.Length))
+            {
+                char[] buffer = new char[length];
```
danmoseley, line 344: as you probably know, if this code was targeting .NET Core (or conditionally compiled thus) you could eliminate this `char[]` allocation by using `string.Create( ...)`

danmoseley: also from IndexOfAnyWhitespace which would be vectorized.
https://github.com/dotnet/runtime/issues/77959

ninedan: Yes, but this project is targeting multiple runtimes.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
ninedan: After addressing the single line issue and regular expression issue, more optimizations could be having much less return.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/ObjectPools/StringBuilderPool.cs
ninedan, line 1: The original code has a bug: StringBuilder is not returned.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
stephentoub: > just curious, @stephentoub on the latest .NET (of course this is broader targeted) would you expect regex replace using pre generated regex to have comparable performance to open coded?

Yes.

Note that in .NET 8, the generated code now looks like this:
```C#
                private bool TryFindNextPossibleStartingPosition(ReadOnlySpan<char> inputSpan)
                {
                    int pos = base.runtextpos;
                    
                    // Empty matches aren't possible.
                    if ((uint)pos < (uint)inputSpan.Length)
                    {
                        // The pattern begins with a whitespace character.
                        // Find the next occurrence. If it can't be found, there's no match.
                        int i = inputSpan.Slice(pos).IndexOfAnyWhiteSpace();
                        if (i >= 0)
                        {
                            base.runtextpos = pos + i;
                            return true;
                        }
                    }
                    
                    // No match found.
                    base.runtextpos = inputSpan.Length;
                    return false;
                }
```
where that IndexOfAnyWhitespace is emitted as:
```C#
        /// <summary>Finds the next index of any character that matches a whitespace character.</summary>
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        internal static int IndexOfAnyWhiteSpace(this ReadOnlySpan<char> span)
        {
            int i = span.IndexOfAnyExcept(Utilities.s_asciiExceptWhiteSpace);
            if ((uint)i < (uint)span.Length)
            {
                if (char.IsAscii(span[i]))
                {
                    return i;
                }
        
                do
                {
                    if (char.IsWhiteSpace(span[i]))
                    {
                        return i;
                    }
                    i++;
                }
                while ((uint)i < (uint)span.Length);
            }
        
            return -1;
        }
        
        /// <summary>Supports searching for characters in or not in "\0\u0001\u0002\u0003\u0004\u0005\u0006\a\b\u000e\u000f\u0010\u0011\u0012\u0013\u0014\u0015\u0016\u0017\u0018\u0019\u001a\u001b\u001c\u001d\u001e\u001f!\"#$%&amp;'()*+,-./0123456789:;&lt;=&gt;?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~\u007f".</summary>
        internal static readonly IndexOfAnyValues<char> s_asciiExceptWhiteSpace = IndexOfAnyValues.Create("\0\u0001\u0002\u0003\u0004\u0005\u0006\a\b\u000e\u000f\u0010\u0011\u0012\u0013\u0014\u0015\u0016\u0017\u0018\u0019\u001a\u001b\u001c\u001d\u001e\u001f!\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~\u007f");
```
such that the search for the whitespace is vectorized.  Then the matching logic for finding all the contiguous whitespace currently looks like:
```C#
                    // Match a whitespace character atomically at least once.
                    {
                        int iteration = 0;
                        while ((uint)iteration < (uint)slice.Length && char.IsWhiteSpace(slice[iteration]))
                        {
                            iteration++;
                        }
                        
                        if (iteration == 0)
                        {
                            return false; // The input didn't match.
                        }
                        
                        slice = slice.Slice(iteration);
                        pos += iteration;
                    }
```
(though we're flirting with a change that'll result in that loop also changing to similarly be vectorized with IndexOfAnyExcept).

Replace itself also gets better in .NET 8.

ninedan:
Can someone please review the changes? Thanks.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
```diff
@@ -268,20 +267,101 @@ internal static string GetText(XmlTextSyntax textElement, bool normalizeWhitespa
...
+
+        /// <summary>
+        /// Append to StringBuilder and perform white space normalization.
+        /// </summary>
+        /// <param name="builder">StringBuilder to append to.</param>
+        /// <param name="text">String to append.</param>
+        /// <param name="normalizeWhitespace">Normalize flag.</param>
+        /// <param name="lastWhitespace">last char is white space flag.</param>
+        /// <returns>True if output is different.</returns>
+        internal static bool AppendNormalize(this StringBuilder builder, string text, bool normalizeWhitespace, ref bool lastWhitespace)
```
sharwell, line 325: ❔ Can we reasonably omit the `normalizeWhitespace` parameter, and update callers that passed false to call `Append` instead?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/XmlCommentHelper.cs
```diff
@@ -268,20 +267,101 @@ internal static string GetText(XmlTextSyntax textElement, bool normalizeWhitespa
...
+                }
+                else
+                {
+                    if (stringBuilder == null)
+                    {
+                        stringBuilder = StringBuilderPool.Allocate();
+                        stringBuilder.AppendNormalize(single, normalizeWhitespace, ref lastWhitespace);
+                    }
+
+                    stringBuilder.AppendNormalize(item.ToString(), normalizeWhitespace, ref lastWhitespace);
```
sharwell, line 290: ❔ Is it common for this line to be hit when `lastWhitespace` is true _and_ `item` is pure whitespace?

sharwell (approved):

PR #3670, opened 2023-06-17
Update SA1500 to also consider init accessors

### LINKED ISSUE #3667: SA1500 is not triggered on init accessors
```
class MyClass
{
    public int X
    {
        get
        { return 0; } // SA1500

        set
        { } // SA1500
    }

    public int Y
    {
        get
        { return 0; } // SA1500

        init
        { } // No SA1500!
    }
}
```

I expect SA1500 (Braces for multi line statements must not share line) to trigger on all accessors in the code above, but it is not triggered on the init accessor.

bjornhellander:
Fixes #3667

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/LayoutRules/SA1500CSharp9UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/LayoutRules/SA1500CSharp9UnitTests.cs
@@ -93,5 +93,50 @@ namespace StyleCop.Analyzers.Test.CSharp9.LayoutRules
                 FixedCode = testCode,
             }.RunAsync(CancellationToken.None).ConfigureAwait(false);
         }
+
+        [Fact]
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/LayoutRules/SA1500CSharp9UnitTests.cs
```diff
@@ -93,5 +93,50 @@ public async Task TestMultiLineRecordWithParameterAsync(string keyword)
...
+    }
+}";
+
+            var test = new CSharpTest
+            {
+                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
+                TestCode = testCode,
+                FixedCode = fixedCode,
+            };
+            await test.RunAsync().ConfigureAwait(false);
```
sharwell, line 139:
```suggestion
            await new CSharpTest
            {
                ReferenceAssemblies = ReferenceAssemblies.Net.Net50,
                TestCode = testCode,
                FixedCode = fixedCode,
            }.RunAsync(CancellationToken.None).ConfigureAwait(false);
```

sharwell (approved):

PR #3678, opened 2023-07-13
Update SA1121 and SA1404 to detect global using aliases, using GetImportScopes and compiling expression trees)

### LINKED ISSUE #3594: SA1121: Inconsistent behaviour with respect to global using aliases
If a global using alias is defined in another file, SA1121 does not produce a warning.
For example, the analyzer will not produce a warning in NoWarning.cs, but in AWarning.cs it does:

GlobalUsings.cs:

```
global using fp64 = System.Double;
```

NoWarning.cs:

```
public class TestWarning
{
    private double Method()
    {
        fp64 x = System.Math.PI;
        return x;
    }
}
```

AWarning.cs:

```
using MyMath = System.Math;

public class TestWarning2
{
    private double Method()
    {
        fp64 x = MyMath.PI;
        return x;
    }
}
```

bjornhellander:
Another attempt to fix #3594.
SA1404 also suffered from the same problem.


bjornhellander:
Tests will fail. That can be fixed and code cleaned up later on. I am more interested in the big picture right now.

Did you have something like this in mind, @sharwell? I haven't wrapped my head around your light-up code completely. I consider my own new code more or less unreadable, but it seems to work :-) I assume that some of it could be generated, but that is beyond me at the moment.

Note that GetImportScopes wasn't available until Roslyn 4.3.0, so with this approach the original issue isn't really resolved perfectly. Are you ok with these rules being broken for Roslyn versions between 4.0.0 (introduction of c# 10) and 4.3.0 (introduction of GetImportScopes)?

Regarding the failing tests: C# 10 tests fail because of the issue with GetImportScopes. What to do?
1) Upgrade to 4.3.0 in the c# 10 test project. It is still before the 4.4.0 which added c# 11 support, but might not be desirable.
2) Move the new tests to the c# 11 test project.
3) ...

bjornhellander:
I moved the tests to the c# 11 test project, because of reasons listed above. Ok?

sharwell:
> I moved the tests to the c# 11 test project, because of reasons listed above. Ok?

That seems fine

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/Lightup/IImportScopeWrapperCSharp11UnitTests.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// Tests calls where nullability rules are not obeyed
+#pragma warning disable CS8604 // Possible null reference argument.
```
sharwell, line 5: Just use `!` at the point where you intentionally break the rules in the test

bjornhellander: Updated

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/Lightup/IImportScopeWrapperCSharp11UnitTests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace StyleCop.Analyzers.Test.Lightup
+{
+    using System;
+    using System.Collections.Immutable;
+    using Microsoft.CodeAnalysis;
+    using StyleCop.Analyzers.Lightup;
+    using Xunit;
+
+    public class IImportScopeWrapperCSharp11UnitTests
```
bjornhellander, line 12: I think that I missed the existing lightup test pattern. Is the idea that there are tests for each language version? Simple tests for c# 6 that verify fallback behavior etc, inherit that one for c#7 through 10, and "cut the inheritance chain" with this new test class for c# 11?

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SyntaxTreeHelpers.cs
```diff
@@ -85,16 +85,28 @@ internal static bool ContainsUsingAlias(this SyntaxTree tree, ConcurrentDictiona
...
         {
+            // Check for "local" using aliases
             var nodes = tree.GetRoot().DescendantNodes(node => node.IsKind(SyntaxKind.CompilationUnit) || node.IsKind(SyntaxKind.NamespaceDeclaration) || node.IsKind(SyntaxKindEx.FileScopedNamespaceDeclaration));
+            if (nodes.OfType<UsingDirectiveSyntax>().Any(x => x.Alias != null))
+            {
+                return true;
+            }
+
+            // Check for global using aliases
+            if (SemanticModelExtensions.SupportsGetImportScopes)
```
bjornhellander, line 103: In a previous version, I ran either the new or the old code, so I needed this SupportsGetImportScopes method to know which path to take. That turned out not to work. As it is written now, this can be simplified by removing the Supports method, since the fallback of GetImportScopes returns an empty array.

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/SemanticModelExtensions.cs
```diff
@@ -0,0 +1,162 @@
...
+        static SemanticModelExtensions()
+        {
+            CreateGetImportScopesAccessor(out var getImportScopesAccessor, out var supportsGetImportScopes);
+            GetImportScopesAccessor = getImportScopesAccessor;
+            SupportsGetImportScopes = supportsGetImportScopes;
+        }
+
+        public static bool SupportsGetImportScopes { get; }
+
+        public static ImmutableArray<IImportScopeWrapper> GetImportScopes(this SemanticModel semanticModel, int position, CancellationToken cancellationToken = default)
```
bjornhellander, line 28: I didn't add any lightup tests for this class, but relied on the SA1121 and SA1404 tests instead. Does that feel good enough?

bjornhellander:
@sharwell I had a look at this again and I now suspect that I misunderstood the "pattern" of the existing lightup tests. I also realized that I over-complicated the SemanticModelExtensions class a bit. But the main thing was the tests. If you could have look at the comments I added when you have time.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SyntaxTreeHelpers.cs
bjornhellander: Fixed this

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/Lightup/IImportScopeWrapperCSharp11UnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(2)]
+        public void TestCompatibleInstance(int numberOfAliasSymbols)
+        {
+            var obj = CreateImportScope(numberOfAliasSymbols);
+            Assert.True(IImportScopeWrapper.IsInstance(obj));
+            var wrapper = IImportScopeWrapper.FromObject(obj);
+            Assert.Equal(obj.Aliases, wrapper.Aliases);
```
bjornhellander, line 25: I updated this test to verify the actual alias symbols returned from the wrapper. Used the Moq package to be able to create IAliasSymbol instances more easily. Ok?

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp11/Lightup/IImportScopeWrapperCSharp11UnitTests.cs
bjornhellander: I updated this, but didn't manage to make it play nicely together with the DerivedTestGenerator the way I described it above. I instead ended up with inheriting all the way from the c# 7 project and up. Does it look ok anyway?

bjornhellander:
Any more comments on this one, @sharwell? I added comments for some concerns that I have myself.

bjornhellander:
Rebased and squashed.
Any more comments, @sharwell?

bjornhellander:
@sharwell?

sharwell:
Sorry for the delays, will try to get to this one soon.

The occasional pings are welcome.

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/SemanticModelExtensions.cs
sharwell: ➡️ Updated this including new tests

sharwell (dismissed):

PR #3694, opened 2023-08-31
Update SA1202 to support interfaces (C# 8) and records (C# 9, 10)

sharwell:


StyleCop.Analyzers/StyleCop.Analyzers/OrderingRules/SA1202ElementsMustBeOrderedByAccess.cs
```diff
@@ -53,9 +53,6 @@ internal class SA1202ElementsMustBeOrderedByAccess : DiagnosticAnalyzer
         private static readonly DiagnosticDescriptor Descriptor =
             new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, AnalyzerCategory.OrderingRules, DiagnosticSeverity.Warning, AnalyzerConstants.EnabledByDefault, Description, HelpLink);
 
-        private static readonly ImmutableArray<SyntaxKind> TypeDeclarationKinds =
-            ImmutableArray.Create(SyntaxKind.ClassDeclaration, SyntaxKind.StructDeclaration);
-
         private static readonly ImmutableHashSet<SyntaxKind> MemberKinds = ImmutableHashSet.Create(
```
bjornhellander, line 56: Looks like this would also need an update to correctly order nested records and record structs, right?

sharwell: I will verify

sharwell: Updated to include records in `MemberKinds`.

bjornhellander: Great!

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/MemberOrderHelper.cs
```diff
@@ -158,6 +158,10 @@ internal static AccessLevel GetAccessLevelForOrdering(SyntaxNode member, SyntaxT
                     {
                         accessibility = AccessLevel.Internal;
                     }
+                    else if (member.Parent.IsKind(SyntaxKind.InterfaceDeclaration))
```
bjornhellander, line 161: Can't seem to add a comment higher up in this file, so adding it here instead.

How about the TypeMemberOrder property in this file and the code in lines 48 and 58? I don't recognize the behaviour that this code seems to imply, but that might be related to how I normally have StyleCop configured. Looks very much like records are missing here.

sharwell: Yes, this revealed another issue which is now fixed.

bjornhellander: Nothing else from me 👍

Issue #3518, opened 2022-06-03
SA1642 does not include records


xLEGiON:
It should be possible to create a record with the following XML documentation header for the constructor:

```csharp
/// <summary>
/// Initializes a new instance of the <see cref="Customer"/> record.
/// </summary>
public Customer()
{
}
```

It should also be possible to end the summary with "record class".

Similarly, record structs should allow the summary to end with "struct" or "record struct".

xLEGiON:
Personally, I like the XML documentation to match the declaring syntax but I guess it's a matter of preference.

My thoughts were that they are sufficiently distinct since the C# language reference lists "record" and "class" separately under "reference types", even if records are just a special type of class underneath.

I understand if there are concerns about adding complexity. I don't know if some sort of configuration setting would be a sufficient compromise? If not, then no problem.

AArnott:
I agree that `record` is *mostly* an implementation detail of the class or struct.

At the moment, SA1642 requires the docs to end with `class` even for `record struct` types, so there is a significant defect at least there, that should be fixed as it's cramping my accurate docs.

bjornhellander:
I can fix @AArnott's case, if you agree that it should be fixed, @sharwell. I also see that the code fix does not work properly for generic records and records structs. Given the current implementation, fixing the code fix would require adding lightup support for RecordDeclarationSyntax, to get hold of the type parameter list. I could add that manually, but it would be nice to get it through the currently missing CSharp9.md etc files instead.

sharwell:
@bjornhellander Lightup support already exists:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/b209e4ef9cf281963192ed77ae846ecaa3e1fe2c/StyleCop.Analyzers/StyleCop.Analyzers/Lightup/.generated/StyleCop.Analyzers.CodeGeneration/StyleCop.Analyzers.CodeGeneration.SyntaxLightupGenerator/RecordDeclarationSyntaxWrapper.g.cs

bjornhellander:
So it does. Missed it completely. Thanks!

PR #3696, opened 2023-09-16
Update SA1642 and its code fix to handle record structs correctly

bjornhellander:
Fixes the problems described by @AArnott and myself in #3518. "record" etc is still not expected in documentation.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
@@ -5,10 +5,13 @@
 
 namespace StyleCop.Analyzers.Test.DocumentationRules
 {
+    using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.DocumentationRules;
+    using StyleCop.Analyzers.Lightup;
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1642UnitTests.cs
```diff
@@ -20,9 +23,28 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
     [UseCulture("en-US")]
     public class SA1642UnitTests
     {
+        public static IEnumerable<object[]> AllTypeKinds
```
sharwell, line 26: This is already defined here:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/890236e1bef6f943554cfdc3b13e2cf2c8651863/StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/CommonMemberData.cs#L14-L32

You can use it like this:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/890236e1bef6f943554cfdc3b13e2cf2c8651863/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1502/SA1502UnitTests.Methods.cs#L28-L29

bjornhellander: Missed it 👍
Updated.

bjornhellander:
Do you have any other comments, @sharwell?

PR #3697, opened 2023-09-16
Update documentation for SA1308 to also mention prefix "t_"

### LINKED ISSUE #3695: SA1308 does not allow "t_" Prefix.
**Using Version**: StyleCop.Analyzers 1.2.0-beta.507

image

I receive a warning for a thread static field with the prefix "t_". 

The [rule description](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1308.md) states the cause as "A field name in C# is prefixed with m_ or s_."

It seems there was a past issue for this: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/688. Perhaps this is a regression bug?

bjornhellander:
Updates documentation to match actual behaviour.
Fixes #3695.

### DIFF
--- a/documentation/SA1308.md
+++ b/documentation/SA1308.md
@@ -17,11 +17,11 @@
 
 ## Cause
 
-A field name in C# is prefixed with *m_* or *s_*.
+A field name in C# is prefixed with *m_*, *s_*, or *t_*.
 
 ## Rule description
 
-A violation of this rule occurs when a field name is prefixed by *m_* or *s_*.
+A violation of this rule occurs when a field name is prefixed by *m_*, *s_*, or *t_*.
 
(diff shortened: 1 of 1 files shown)


documentation/SA1308.md
```diff
@@ -17,11 +17,11 @@
 
 ## Cause
 
-A field name in C# is prefixed with *m_* or *s_*.
+A field name in C# is prefixed with *m_*, *s_* or *t_*.
```
sharwell, line 20:
```suggestion
A field name in C# is prefixed with *m_*, *s_*, or *t_*.
```

documentation/SA1308.md
```diff
@@ -17,11 +17,11 @@
...
 
 ## Rule description
 
-A violation of this rule occurs when a field name is prefixed by *m_* or *s_*.
+A violation of this rule occurs when a field name is prefixed by *m_*, *s_* or *t_*.
```
sharwell, line 24:
```suggestion
A violation of this rule occurs when a field name is prefixed by *m_*, *s_*, or *t_*.
```

sharwell (changes requested):

documentation/SA1308.md
bjornhellander: Updated

sharwell (approved):

PR #3714, opened 2023-10-18
Temporary workaround for unstable SA1134 code fix

bjornhellander:
Would this be ok to get rid on random failures of the test VerifyMultipleAttributesOnSameLineForMembersAsync?
This happens quite often unfortunately. Unsure if this helps, but I interpret the documentation like it should.


sharwell:
I'm planning to take #2853 instead of this one

PR #3719, opened 2023-11-03
Allow inheritdoc for class constructors with base types

### LINKED ISSUE #2575: SA1648 - inheritdoc on constructors
SA1648 is being reported when `<inheritdoc />` is used on constructors within a subclass of the same signature of the parent. Although there may not be a solid standard for `inheritdoc`, Sandcastle's implementation seems to be taking on that role and calls out support for constructors (ReSharper as well):

[The inheritdoc element is valid at the root level (i.e. the same level as summary elements) on types, interfaces, virtual members, interface member implementations, and constructors.](https://ewsoftware.github.io/XMLCommentsGuide/html/86453FFB-B978-4A2A-9EB5-70E118CA8073.htm)

I suggest that StyleCopAnalyzers supports this usage.

```
/// <summary>
/// A class
/// </summary>
public class A
{
    /// <summary>
    /// Initializes a new instance of the <see cref="A"/> class.
    /// </summary>
    /// <remarks>Initializes hardware...</remarks>
    public A()
    {
    }
}

/// <summary>
/// B class
/// </summary>
public class B : A
{

    /// <summary>
    /// Initializes a new instance of the <see cref="B" /> class.
    /// </summary>
    /// <inheritdoc />
    public B()
    {
    }
}
```

MartyIX:
Fixes #2575

I'm not sure if the fix is correct or not. Hopefully it is.


bjornhellander:
My understanding is that the signature has to match a constructor from a base class for the `inheritdoc` element to be ok.

bjornhellander:
I am not aware of any nice utility methods for this.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1648InheritDocMustBeUsedWithInheritingClass.cs
```diff
@@ -159,6 +159,56 @@ private static void HandleMemberDeclaration(SyntaxNodeAnalysisContext context)
                 }
             }
 
+            if (memberSyntax is ConstructorDeclarationSyntax constructorDeclarationSyntax)
+            {
+                ISymbol symbol = context.SemanticModel.GetDeclaredSymbol(constructorDeclarationSyntax, context.CancellationToken);
```
bjornhellander, line 164: Looks like the symbol is already available in variable "declaredSymbol" from 10+ lines above, so I don't think you need to get it again

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1648InheritDocMustBeUsedWithInheritingClass.cs
```diff
@@ -159,6 +159,56 @@ private static void HandleMemberDeclaration(SyntaxNodeAnalysisContext context)
...
+
+                if (symbol is IMethodSymbol constructorMethodSymbol && constructorMethodSymbol.ContainingType is INamedTypeSymbol enclosingNamedTypeSymbol)
+                {
+                    INamedTypeSymbol baseType = enclosingNamedTypeSymbol.BaseType;
+
+                    if (baseType.SpecialType != SpecialType.System_Object)
```
bjornhellander, line 170: Do we really want to skip classes that inherits from Object? Maybe not the best documentation, but it seems to work:

image

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-#nullable disable
+#nullable enable
```
bjornhellander, line 4: Null reference types are already on in the project file, so you can just delete this line instead

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -18,6 +18,40 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
     /// </summary>
     public class SA1648UnitTests
     {
+        [Fact]
```
bjornhellander, line 21: Focus here and let the tests drive your changes in the analyzer. I assume that we need many more tests to cover all situations when it should and should not fire.

I would especially have added some tests where the base is class is not in the compiled assembly. The easiest route might be to always inherit from an external base classes (or none at all).

How about structs and records?

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -18,6 +18,40 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
...
+
+class Test : Base
+{
+    /// <inheritdoc />
+    public Test() { }
+}";
+
+            await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+
+            testCode = @"class Base
```
bjornhellander, line 38: Split it into two tests, or write it as a Theory instead of a Fact

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1648InheritDocMustBeUsedWithInheritingClass.cs
```diff
@@ -159,6 +159,56 @@ private static void HandleMemberDeclaration(SyntaxNodeAnalysisContext context)
                 }
             }
 
+            if (memberSyntax is ConstructorDeclarationSyntax constructorDeclarationSyntax)
+            {
+                ISymbol symbol = context.SemanticModel.GetDeclaredSymbol(constructorDeclarationSyntax, context.CancellationToken);
+
+                if (symbol is IMethodSymbol constructorMethodSymbol && constructorMethodSymbol.ContainingType is INamedTypeSymbol enclosingNamedTypeSymbol)
```
bjornhellander, line 166: There is too much code here for my taste. Move most of this to a (or multiple) helper method to make it easier to read. Possibly in NamedTypeExtensions, which already has similar functionality, but since this class would be the only caller right now, putting it here might be preferable.

Candidates for extraction:
- ConstructorSignatureMatchesBaseType(IMethodSymbol)
- bool HasMatchingSignatures(IMethodSymbol, IMethodSymbol)

bjornhellander:
Apart from my question about inheriting from Object, I think the analyzer changes look correct. I would just like to see more tests and improved readability. 👍

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
sharwell: > The extension Fine Code Coverage is free and easy to set up and use if you currently don't have any easy way of checking coverage. Aim for 100% coverage for your new code.

You can also send a PR, and click the link to the Codecov Report. By default, it will filter the view down to just the files which changed in the PR so you can see the coverage of code you added/changed.

bjornhellander: > Do you know about an analyzer where this is done (for me to better understand it)? Or could you tell a me good candidate for such base class?
> 
> I mean I think I understand what you suggest but I'm not sure if there is a convention how this is being tackled in this library. I would like to avoid reinventing the wheel if possible.

Sorry, I forgot this. The "core" assemblies are available to you automaatically in the test code, so you can for example use ArgumentException, which has multiple constructors with and without parameters. Or List or any other core type that suits your needs.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -18,6 +16,68 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
...
+        [Fact]
+        public async Task TestConstructorInheritsExplicitlyFromSystemObjectAsync()
+        {
+            var testCode = @"class Test : System.Object
+{
+    /// <inheritdoc />
+    public Test() { }
+}";
+
+            await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
```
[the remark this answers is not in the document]
bjornhellander: I will create a separate issue for this instead of discussing it here

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1648InheritDocMustBeUsedWithInheritingClass.cs
```diff
@@ -149,10 +147,32 @@ private static void HandleMemberDeclaration(SyntaxNodeAnalysisContext context)
...
 
             ISymbol declaredSymbol = context.SemanticModel.GetDeclaredSymbol(memberSyntax, context.CancellationToken);
+
+            if (memberSyntax is ConstructorDeclarationSyntax constructorDeclarationSyntax && declaredSymbol is IMethodSymbol constructorMethodSymbol)
+            {
+                if (constructorMethodSymbol.ContainingType is INamedTypeSymbol enclosingNamedTypeSymbol)
+                {
+                    INamedTypeSymbol baseType = enclosingNamedTypeSymbol.BaseType;
+
+                    if (baseType.SpecialType == SpecialType.System_Object)
```
bjornhellander, line 157: Is this special case needed? It seems everything would work fine without it, but I might be wrong. Personally, I would remove it if it's not needed.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -18,6 +17,87 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
...
+    public Test(string s, int b)
+        : base(s, b) { }
+}
+";
+
+            await VerifyCSharpDiagnosticAsync(testCode.Replace("$KEYWORD$", keyword), DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [MemberData(nameof(CommonMemberData.ReferenceTypeDeclarationKeywords), MemberType = typeof(CommonMemberData))]
```
bjornhellander, line 61: Was there a reason for not testing structs and record structs here?

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -18,6 +17,87 @@ namespace StyleCop.Analyzers.Test.DocumentationRules
     /// </summary>
     public class SA1648UnitTests
```
bjornhellander, line 18: I don't see any tests for when it **should** trigger, for example if there is a constructor in the base class but it has the wrong number of parameters, or if it has the same number of arguments but one of them is of the wrong type.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1648InheritDocMustBeUsedWithInheritingClass.cs
```diff
@@ -149,10 +147,32 @@ private static void HandleMemberDeclaration(SyntaxNodeAnalysisContext context)
             Location location;
 
             ISymbol declaredSymbol = context.SemanticModel.GetDeclaredSymbol(memberSyntax, context.CancellationToken);
+
+            if (memberSyntax is ConstructorDeclarationSyntax constructorDeclarationSyntax && declaredSymbol is IMethodSymbol constructorMethodSymbol)
+            {
+                if (constructorMethodSymbol.ContainingType is INamedTypeSymbol enclosingNamedTypeSymbol)
```
bjornhellander, line 153: ContainingType is defined as INamedTypeSymbol, so this can be simplified.

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
bjornhellander: Ok. No, I don't think so. I just didn't realize there were complications.

azure-pipelines.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       Release:
         BuildConfiguration: Release
         _debugArg: ''
+  timeoutInMinutes: 90
```
bjornhellander, line 22: The timeout problem has been solved in another way and it is on master now, so I would revert this and rebase instead. And maybe squash as well so we don't get 20 new commits for this change.

bjornhellander:
When you are done with everything mentioned so far, ping sharwell. 👍

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -1,14 +1,14 @@
...
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-#nullable disable
-
 namespace StyleCop.Analyzers.Test.DocumentationRules
 {
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.CodeAnalysis.Testing;
+    using Microsoft.VisualBasic.Devices;
```
sharwell, line 9: 😕 Is this needed?

StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1648UnitTests.cs
```diff
@@ -90,7 +215,6 @@ public async Task TestTypeWithEmptyBaseListAndCrefAttributeAsync(string declarat
         }
 
         [Theory]
-        [InlineData("Test() { }")]
```
sharwell, line 93: 📝 Let's add this back (so we have coverage of constructors), but change the signature (so it doesn't have a matching base type constructor)

For example:

```cs
[InlineData("Test(int ignored) { }")]
```

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1648InheritDocMustBeUsedWithInheritingClass.cs
```diff
@@ -206,15 +218,57 @@ private static void HandleMemberDeclaration(SyntaxNodeAnalysisContext context)
...
+                    {
+                        success = false;
+                        break;
+                    }
+                }
+
+                if (success)
+                {
+                    found = true;
+                    break;
```
sharwell, line 256: Can just return true here?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1648InheritDocMustBeUsedWithInheritingClass.cs
```diff
@@ -206,15 +218,57 @@ private static void HandleMemberDeclaration(SyntaxNodeAnalysisContext context)
...
+        /// </summary>
+        /// <returns><see langword="true"/> if any base type constructor's signature matches the signature of <paramref name="constructorMethodSymbol"/>, <see langword="false"/> otherwise.</returns>
+        private static bool HasMatchingSignature(ImmutableArray<IMethodSymbol> baseConstructorSymbols, IMethodSymbol constructorMethodSymbol)
+        {
+            bool found = false;
+
+            foreach (IMethodSymbol baseConstructorMethod in baseConstructorSymbols)
+            {
+                // Constructors must have the same number of parameters.
+                if (constructorMethodSymbol.Parameters.Count() != baseConstructorMethod.Parameters.Count())
```
sharwell, line 233:
```suggestion
                if (constructorMethodSymbol.Parameters.Length != baseConstructorMethod.Parameters.Length)
```

sharwell (approved):

Issue #3687, opened 2023-08-10
SA1010 false positive when list syntax is used in a field initializer


AArnott:
The following line produces a SA1010 warning claiming that the open square bracket should not be preceded by a space:

```cs
    protected static readonly string[] DefaultMetadataPaths = [MetadataPath, WdkMetadataPath];
```

This is using 1.2.0.507.

MiYanni:
I hit this when telling VS to make fixes for [Use collection initializers (IDE0028)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0028). It will automatically create statements that trigger SA1010. For example:
```C#
List<string> foo = [];
```

dciborow:
Hit this in a few places.

```csharp
ctors = [.. ctors.OrderByDescending(info => info.GetParameters().Length)];
```

```csharp
object result = TypesManager.InvokeGenericStaticMethod(typeof(ServiceLocator), nameof(ResolveMock), [type]);
```

```csharp
if (request.OtherEmails != null)
{
    emails = [.. request.OtherEmails.Split([';', ','])];
}
```

```csharp
        public static readonly string[] AllBlastRadiusStrings = [SingleNode, SingleCustomer, SingleScaleUnitSlicePartition, SingleScaleUnit, SingleRack, SingleCluster, SingleAZ, SingleRegion, MultipleRacks, MultipleClusters, MultipleAZs, MultipleRegions, Global, Other];
```

bjornhellander:
I assume that .NET 8 and C# 12 will be released very soon now. This fix is probably trivial, but what do you think about adding a new test project, @sharwell? Should it wait until you are done with your updates to the test projects (https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3675#issuecomment-1740054572), or can it be done now? If you would like to wait, do you have any estimate of when you will be done?

DavidArno:
.NET 8 and C# 12 were released yesterday and so this is now an issue with production code. For now, I've had to disable the SA1010 rule for our code until this can be fixed. Please can this be fixed as quickly as possible.

commonsensesoftware:
I'm unsure if all the various cases are being tracked or accounted for, but I thought I would report that SA1010 also fires during coalesce, coalesce assign, and ternary operations.

```c#
List<string>? list = default;

// fires on coalesce
List<string> otherList = list ?? [];

// fires on ternary
[...]

PR #3729, opened 2023-11-15
Redesign SA1010's handling of preceding whitespace to explicitly state expected style in all known cases and ignore all unknown cases.

bjornhellander:
Partial fix for #3687.

This attempts to improve the implementation of SA1010 by explicitly stating the expected style in all known situations and ignoring any unknown situations, instead of only listing situations where preceding whitespace is ok and assuming that all other situations should not have whitespace. My hope is that this will result in fewer false positives as the language evolves.

This "should" fix the reported issue with collection expressions, but if this change is approved, tests for collection expressions still need to be added when a c# 12 test project exists.


sharwell:
Given that the warnings appearing after a language change are the primary method we use to find and update the product to account for these changes, I'm hesitant to make a change that would hide these in the future.

symbiogenesis:
@sharwell so you seem to think we should have a whitelist rather than a blacklist. So the appropriate fix to get this merged is to set the default case to true?

bjornhellander:
Changing the default case would mean that this PR has zero affect functionally. Specifically, the linked issue would no longer be fixed by it. I will re-write this instead, now that we can add tests for c# 12 features.

Issue #3460, opened 2022-02-18
False positive on switch expression with await 


karol-gro:
Stylecop version: 1.2.0-beta.406

I get false positive issue on following code:
```csharp
// this gives SA1119 
await (someBool switch
{
	true => DoOneThing(),
	false => DoAnother(),
});

// after applying suggestions we got not compiling code (await is applied on bool, so parentheses are required to change order of operations)
await someBool switch
{
	true => DoOneThing(),
	false => DoAnother(),
};
```
Of course in this case one should use `if`, but this is minimized version of issue found in our code

Issue #3200, opened 2020-09-19
False positive on SA1119 for switch expression


DAud-IcI:
I found a C# 8 related bug. If you want to directly access a member on the result of a switch expression it has to be parenthesized otherwise you get a plain syntax error. For example SA1119 triggers on these parenteses:

```
var translatedString = (approvalStatus switch
{
    UserApprovalStatus.Approved => T["Approved"],
    UserApprovalStatus.Denied => T["Denied"],
    UserApprovalStatus.Pending => T["Pending"],
    _ => null,
})?.Value;
```

Seems like the switch expression behaves like a binary operator so the parentheses are necessary and removing them generates a bunch of syntax errors.

PR #3733, opened 2023-11-25
Update SA1119 to allow parenthesized switch expressions followed by an invocation

### LINKED ISSUE #3730: SA1119 false positive for parenthesis around `x switch {}` expression
https://github.com/n0099/open-tbm/blob/af3fe85e196e0ffaa0fa881e86b2c23999e0f736/c%23/imagePipeline/src/Ocr/PaddleOcrRecognizerAndDetector.cs#L25-L34
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression

bjornhellander:
Fixes #3730


sharwell (changes requested):
Need to determine if this situation is already covered by #3200 and/or #3460

sharwell (approved):

PR #3734, opened 2023-11-25
Add c# 12 test project

bjornhellander:
I took the liberty of adding a test project for c# 12. I assume that this will have to wait until your test changes are done @sharwell , but this can be rebased or re-done easily. It was mostly mechanical work anyway. 4 (?) test cases failed on c# 12, so I updated those.


StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
```diff
@@ -1010,5 +983,48 @@ public void Test()
...
+                Diagnostic().WithLocation(10, 25),
+                Diagnostic().WithLocation(11, 30),
+                Diagnostic().WithLocation(12, 30),
+                Diagnostic(CS1065).WithLocation(12, 53),
+                Diagnostic().WithLocation(13, 30),
+                Diagnostic(CS7014).WithLocation(13, 47),
+                Diagnostic().WithLocation(14, 30),
+                Diagnostic(CS1670).WithLocation(14, 47),
+                Diagnostic().WithLocation(15, 25),
+                Diagnostic(CS1669).WithLocation(15, 42),
```
sharwell, line 1002: 💡 Let's move all the `Diagnostic()` items back into the test itself, and only have the `CS*` diagnostics in this virtual method.

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
```diff
@@ -1010,5 +983,48 @@ public void Test()
...
+            return new[]
+            {
+                Diagnostic().WithLocation(12, 30),
+                Diagnostic(CS1065).WithLocation(12, 53),
+                Diagnostic().WithLocation(13, 30),
+                Diagnostic(CS7014).WithLocation(13, 47),
+                Diagnostic().WithLocation(14, 30),
+                Diagnostic(CS1670).WithLocation(14, 47),
+                Diagnostic().WithLocation(15, 25),
+                Diagnostic(CS1669).WithLocation(15, 42),
```
sharwell, line 1017: 💡 Same applies for `Diagnostic()` here

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
```diff
@@ -1010,5 +983,48 @@ public void Test()
...
+            };
+        }
+
+        protected virtual DiagnosticResult[] GetExpectedResultVerifyInvalidCodeConstructions()
+        {
+            return new[]
+            {
+                Diagnostic().WithSpan(4, 50, 4, 58),
+                DiagnosticResult.CompilerError("CS1660").WithMessage("Cannot convert anonymous method to type 'EventHandler[]' because it is not a delegate type").WithSpan(4, 50, 4, 62),
```
sharwell, line 1026: 💡 Same here

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
bjornhellander: Sure. I thought of that too briefly, but ended up not doing it. 👍

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
bjornhellander: I noticed that the only thing differing was the span, so I simplified to only check the start location. Ok simplification?

StyleCop.Analyzers/StyleCop.Analyzers.Test/ReadabilityRules/SA1130UnitTests.cs
bjornhellander: Added them as annotations in the tested code instead. This removed the need for two methods before and after the fix. Ok simplification?

sharwell (approved):

sharwell:
It may be possible to split the test executions into separate jobs that run in parallel, and then merge the result in a cleanup step. For now I just hit retry on the build.

PR #3741, opened 2023-12-03
Update SA1008 to not crash if there is no previous token

### LINKED ISSUE #2354: SA1008 threw a NullReferenceException
Using StyleCopAnalyzers 1.0 in VS2017 for a .NET 4.5 project, C#6. While refactoring I got the following error. At some point during my refactorings the file referenced in the error was removed, don't know if that is related or if this error occurred before that.  

```
Severity	Code	Description	Project	File	Line	Suppression State	Detail Description
Warning	AD0001	Analyzer 'StyleCop.Analyzers.SpacingRules.SA1008OpeningParenthesisMustBeSpacedCorrectly' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.	<project>	1	Active	Analyzer 'StyleCop.Analyzers.SpacingRules.SA1008OpeningParenthesisMustBeSpacedCorrectly' threw the following exception:
'Exception occurred with following context:
Compilation: <project>
SyntaxTree: <source path>

System.NullReferenceException: Object reference not set to an instance of an object.
   at StyleCop.Analyzers.Helpers.LocationHelpers.GetLineSpan(SyntaxToken token)
   at StyleCop.Analyzers.SpacingRules.SA1008OpeningParenthesisMustBeSpacedCorrectly.HandleOpenParenToken(SyntaxTreeAnalysisContext context, SyntaxToken token)
   at StyleCop.Analyzers.SpacingRules.SA1008OpeningParenthesisMustBeSpacedCorrectly.HandleSyntaxTree(SyntaxTreeAnalysisContext context)
   at StyleCop.Analyzers.AnalyzerExtensions.<>c__DisplayClass2_0.<RegisterSyntaxTreeActionHonorExclusions>b__0(SyntaxTreeAnalysisContext c)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c__DisplayClass41_0.<ExecuteSyntaxTreeActionsCore>b__0()
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock(DiagnosticAnalyzer analyzer, Action analyze, Nullable`1 info)
-----
'.
```

bjornhellander:
Fixes #2354

Protected all similar methods in LocationHelpers against null reference exception.
I found one situation where this happens in SA1008 so I added a unit test for that as well.

I also went through all uses of the SyntaxTree properties in SyntaxToken, SyntaxTrivia and SyntaxNodeOrToken and tried to make sure that they could not cause a null reference exceptions.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1008CSharp9UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1008CSharp9UnitTests.cs
@@ -5,8 +5,10 @@
 
 namespace StyleCop.Analyzers.Test.CSharp9.SpacingRules
 {
+    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.Test.CSharp8.SpacingRules;
     using Xunit;
@@ -80,6 +82,33 @@ class C
 }";
 
             await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [InlineData("")]
+        [InlineData(" ")]
+        [InlineData("\n")]
+        [InlineData("\n ")]
+        [WorkItem(2354, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2354")]
+        public async Task TestDeconstructionInTopLevelProgramAsync(string prefix)
+        {
+            var testCode = $@"{prefix}{{|#0:(|}} var a, var b) = (1, 2);";
+            var fixedCode = $@"{prefix}(var a, var b) = (1, 2);";
+
+            await new CSharpTest()
+            {
+                TestState =
+                {
+                    OutputKind = OutputKind.ConsoleApplication,
+                    Sources = { testCode },
+                },
+                ExpectedDiagnostics =
+                {
+                    // /0/Test0.cs(1,1): warning SA1008: Opening parenthesis should not be followed by a space.
+                    Diagnostic(DescriptorNotFollowed).WithLocation(0),
+                },
+                FixedCode = fixedCode,
+            }.RunAsync(CancellationToken.None).ConfigureAwait(false);
         }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
@@ -142,7 +142,11 @@ namespace StyleCop.Analyzers.SpacingRules
             var leadingTriviaList = TriviaHelper.MergeTriviaLists(prevToken.TrailingTrivia, token.LeadingTrivia);
 
             var isFirstOnLine = false;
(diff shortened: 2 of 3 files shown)


bjornhellander:
Ready for review

bjornhellander:
Updated to only handle SA1008.

StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1008OpeningParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -100,6 +100,9 @@ private static void HandleOpenParenToken(SyntaxTreeAnalysisContext context, Synt
             bool precededByKeyword;
             switch (prevToken.Kind())
             {
+            case SyntaxKind.None:
+                return;
```
sharwell, line 104: 📝 This won't correctly handle the following top-level program:

```cs
 ( var a, var b) = (3, 3);
System.Console.WriteLine($"{a}");
```

bjornhellander: Good catch! Didn't think of that.
Alright, I will have a look.

bjornhellander: I see now that you added both a leading and a trailing space in your example. What was your intention with the leading space?

sharwell: It should behave the same way as this:

```cs
void Method()
{
     ( var a, var b) = (3, 3);
    System.Console.WriteLine($"{a}");
}
```

bjornhellander: As far as I can understand, this is a "don't care" case, since it is the first token on the line

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/SpacingRules/SA1008CSharp9UnitTests.cs
```diff
@@ -81,5 +83,35 @@ void Method((int, int) c)
...
+            {
+                TestState =
+                {
+                    OutputKind = OutputKind.ConsoleApplication,
+                    Sources = { testCode },
+                },
+                ExpectedDiagnostics =
+                {
+                    // /0/Test0.cs(1,1): warning SA1008: Opening parenthesis should not be followed by a space.
+                    Diagnostic(DescriptorNotFollowed).WithLocation(1 + lineOffset, 1 + columnOffset),
```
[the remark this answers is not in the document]
bjornhellander: Updated

bjornhellander:
Test execution seems to fail without a reason in the previous build. I noticed that both OpenCover and xunit.runner.console are quite old (from 2016 and 2018). What do you think about updating them?

sharwell:
> Test execution seems to fail without a reason in the previous build.

I think this is a different issue that I can likely work around in the build system.

bjornhellander:
> > Test execution seems to fail without a reason in the previous build.
> 
> I think this is a different issue that I can likely work around in the build system.

Absolutely, I didn't mean in this PR. It looked like the test runner just died before reporting the outcome in the c# 6 debug run and in the c# 12 debug run, all tests passed but opencover.console.exe exited with code 1 without saying why. Just looked weird, so upgrading them might be a good thing. Unless you already had ideas about it.

sharwell (approved):

PR #3744, opened 2023-12-05
Increase build timeout to 90 minutes

sharwell:


sharwell:
This is no longer necessary because the pipeline now runs in less than 30 minutes

PR #3748, opened 2023-12-08
Update SA1008 to allow space before the opening parenthesis of a using alias definition of a tuple type

### LINKED ISSUE #3743: SA1008 False Positive if a tuple type is assigned an alias via `using`.
```
    using WeaveTemplateItem = (string path, Microsoft.CodeAnalysis.Text.SourceText text, bool useSourceGeneration, bool? configFileExists);
```

image

This seems like it should be an extension of the current logic that would allow an assignment like this:

```
    var x = (a + b) * c;
```

To more than just assignment expressions.

bjornhellander:
Fixes #3743

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/SpacingRules/SA1008CSharp12UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/SpacingRules/SA1008CSharp12UnitTests.cs
@@ -3,9 +3,30 @@
 
 namespace StyleCop.Analyzers.Test.CSharp12.SpacingRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
     using StyleCop.Analyzers.Test.CSharp11.SpacingRules;
+    using Xunit;
+
+    using static StyleCop.Analyzers.SpacingRules.SA1008OpeningParenthesisMustBeSpacedCorrectly;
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/SpacingRules/SA1008CSharp12UnitTests.cs
```diff
@@ -3,9 +3,26 @@
...
+        StyleCop.Analyzers.SpacingRules.TokenSpacingCodeFixProvider>;
 
     public partial class SA1008CSharp12UnitTests : SA1008CSharp11UnitTests
     {
+        [Fact]
+        [WorkItem(3743, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3743")]
+        public async Task TestTupleUsingAliasAsync()
+        {
+            const string testCode = @"
+using TestAlias = (string X, bool Y);";
```
sharwell, line 23: 💡 This should be tested similar to others where it also tests that a diagnostic is reported if there is no space, and the code fix inserts the missing space.

bjornhellander: Updated

sharwell (approved):

PR #3750, opened 2023-12-09
Update SA1009 to require a space after the closing parenthesis if it is followed by ++ or -- from a prefix unary expression

### LINKED ISSUE #3731: SA1009 is triggered for the increment in prefix notation
1.2.0-beta.507

The prefix notation will trigger the check

image

The postfix notation will not

image

bjornhellander:
Fixes #3731

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs
@@ -181,7 +181,9 @@ namespace StyleCop.Analyzers.SpacingRules
 
             case SyntaxKind.PlusPlusToken:
             case SyntaxKind.MinusMinusToken:
-                precedesStickyCharacter = true;
+                precedesStickyCharacter =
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1009ClosingParenthesisMustBeSpacedCorrectly.cs
```diff
@@ -204,6 +204,13 @@ private static void HandleCloseParenToken(SyntaxTreeAnalysisContext context, Syn
                 precedesStickyCharacter = true;
                 break;
 
+            case SyntaxKind.IfStatement:
+            case SyntaxKind.WhileStatement:
+            case SyntaxKind.ForStatement:
+            case SyntaxKind.ForEachStatement:
+                precedesStickyCharacter = false;
+                break;
```
sharwell, line 212: 💭 I think I would prefer this to be relocated to the handling of `PlusPlusToken`/`MinusMinusToken` above, and the new implementation in that section should be similar to the way `PlusToken` and `MinusToken` and handled further above.

bjornhellander: Updated. Did you have something like that in mind?

sharwell (approved):

PR #3751, opened 2023-12-09
[WIP] Update SA1600 to not require documentation for explicitly implemented members

### LINKED ISSUE #3717: SA1600 should not trigger for explicit interface implementations
Given the following declarations:

```cs
public interface IFoo
{
	/// <summary>
	/// My decoding method.
	/// </summary>
	static abstract bool TryDecode();
}

public class Foo : IFoo
{
	static bool IFoo.TryDecode() => true;
}
```

SA1600 flags the method on the `Foo` class.

This is unreasonable, since there is no way at all to invoke this method except by the interface, which is already documented. Adding `/// <inheritdoc />` to the implementation adds nothing. 

Per https://github.com/dotnet/roslyn/issues/70456, Roslyn does not offer a code fix to add the inheritdoc in the explicit interface implementation case, and this reasoning came up in the discussion.

bjornhellander:
Fixes #3717.

Good thing: Only the SA1600 code is touched.

Bad thing: This change leaves some old code uncovered by tests. The code is regarding ExplicitInterfaceSpecifier and is located inside the various GetDeclaredAccessibility methods in AccessLevelHelper. Could be because of missing tests for some other analyzers, or that the now uncovered code was actually unnecessary for all other analyzers that call these methods, however strange that sounds. If this change is accepted I can write another issue to investigating what to do about it.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/DocumentationRules/SA1600UnitTests.cs
@@ -90,7 +90,7 @@ using System;
         public async Task TestMethodWithoutDocumentationAsync()
         {
             await this.TestMethodDeclarationDocumentationAsync(string.Empty, false, false, false).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


bjornhellander:
No, I don't think this is right. Closing and starting discussion in issue first.

PR #3753, opened 2023-12-17
Change default value of test property CodeActionValidationMode to SemanticStructure instead of None and update so that tests still pass

### LINKED ISSUE #2785: Avoid using CodeFixValidationMode.None in tests
🔗 Issue filed in response to https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/2751#discussion_r215483408

* Tests should validate the semantic structure by default
* The few tests which fail validation should be updated to pass semantic structure validation or use a localized configuration of `CodeFixValidationMode.None`

bjornhellander:
Fixes #2785

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SA1129CodeFixProvider.cs
@@ -214,6 +214,11 @@ namespace StyleCop.Analyzers.ReadabilityRules
         /// <returns>A new member access expression.</returns>
         private static SyntaxNode ConstructMemberAccessSyntax(TypeSyntax typeSyntax, string memberName)
         {
+            // NOTE: This creates the correct source code when applied, but these are not necessarily the syntax
+            // nodes that the compiler would create from that source code. For example, the type syntax can
+            // contain QualifiedName nodes, whereas the compiler would have created SimpleMemberAccessExpression instead.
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/StyleCop.Analyzers.CodeFixes.csproj
```diff
@@ -36,6 +36,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="1.2.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="1.2.1" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
bjornhellander, line 39: Is this ok?

sharwell: I'd be concerned about the runtime impact of this on VS 2015. It seems like we could use `out` parameters instead pretty easily.

bjornhellander: Absolutely. I will update.

bjornhellander: Updated

sharwell (approved):

PR #3760, opened 2023-12-20
Fix SA1131 to not treat "complex" expressions as a literal

### LINKED ISSUE #3759: Unexpected SA1131
I updated this repository to 1.2.0-beta.553 and got lots of unexpected SA1131.
Triggers for example on these cases:
`x + 1 < y`
`a() < y`

The common denominator seems to be that the left side or the operator is not just a variable, so this is probably a very common problem.

bjornhellander:
Fixes #3759.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1131UseReadableConditions.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1131UseReadableConditions.cs
@@ -86,7 +86,11 @@ namespace StyleCop.Analyzers.ReadabilityRules
             switch (symbol)
             {
             case IFieldSymbol fieldSymbol when fieldSymbol.IsStatic && fieldSymbol.IsReadOnly:
-            case IMethodSymbol:
+                return true;
+
+            // NOTE: Without the when clause, this would also treat e.g unary or binary expressions as literals,
+            // since GetSymbolInfo returns the operator method in those cases.
+            case IMethodSymbol when expression is IdentifierNameSyntax:
                 return true;
 
             default:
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1131UseReadableConditions.cs
```diff
@@ -86,7 +86,11 @@ private static bool IsLiteral(ExpressionSyntax expression, SemanticModel semanti
             switch (symbol)
             {
             case IFieldSymbol fieldSymbol when fieldSymbol.IsStatic && fieldSymbol.IsReadOnly:
-            case IMethodSymbol:
+                return true;
+
+            // NOTE: Without the when clause, this would also treat e.g unary or binary expressions as literals,
+            // since GetSymbolInfo returns the operator method in those cases.
+            case IMethodSymbol when expression is IdentifierNameSyntax:
```
sharwell, line 93: 📝 I think this needs to be `NameSyntax` instead. Also the tests should include some examples involving qualified methods (e.g. `TestClass.Method1`) and generic methods (e.g. `Method3<int>`, `Enum.TryParse<DateTimeKind>`)

bjornhellander: Updated. Did not add test for Enum.TryParse since I couldn't get it to compile (due to multiple overloads) in time.

I have to leave now. Feel free to revert commit dfd5ed2e454d71a563dac50ca7bed57a37d31e9f instead if you would like. I can redo that one later on if you see more problems with this and you would like to make a new beta quickly.

sharwell (approved):

PR #3762, opened 2023-12-23
Fix csharp_using_directive_placement setting retrieval

kvpt:
Hi,

I use this setting in my editorconfig file
> csharp_using_directive_placement = outside_namespace

But the StyleCop Analyzer was not using it.

So I analyzed the source code and found that the method that read the setting expects that the setting is followed by the severity level.
This is not the case for me and so the setting is not retrieved.

Currently the code that set the setting not use the severity level so I just return an empty string in this case.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Settings/SettingsCSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Settings/SettingsCSharp8UnitTests.cs
@@ -5,6 +5,7 @@
 
 namespace StyleCop.Analyzers.Test.CSharp8.Settings
 {
+    using System;
     using System.Collections.Immutable;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


sharwell:
Can you add some new tests to this class?
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/1b5a75a06ed5a7eeae062bcef429363c074d6f8b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Settings/SettingsCSharp8UnitTests.cs

kvpt:
It's done

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Settings/SettingsCSharp8UnitTests.cs
```diff
@@ -172,6 +172,22 @@ public async Task VerifyInvalidReferenceBehaviorFromEditorConfigAsync()
             Assert.Equal("[InvalidReference]", styleCopSettings.DocumentationRules.GetCopyrightText("unused"));
         }
 
+        [Fact]
+        public async Task VerifyEditorConfigSettingsReadCorrectlyDirectivePlacementWithoutSeverityLevelAsync()
```
sharwell, line 176:
```suggestion
        [Theory]
        [InlineData("outside_namespace")]
        [InlineData("inside_namespace")]
        public async Task VerifyEditorConfigSettingsReadCorrectlyDirectivePlacementWithoutSeverityLevelAsync(string placement)
```

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Settings/SettingsCSharp8UnitTests.cs
```diff
@@ -172,6 +172,22 @@ public async Task VerifyInvalidReferenceBehaviorFromEditorConfigAsync()
...
+        {
+            var settings = @"root = true
+
+[*]
+csharp_using_directive_placement = outside_namespace
+";
```
sharwell, line 182:
```suggestion
            var expected = placement switch
            {
                "outside_namespace" => UsingDirectivesPlacement.OutsideNamespace,
                "inside_namespace" => UsingDirectivesPlacement.InsideNamespace,
                _ => throw new InvalidOperationException(),
            };
            var settings = $@"root = true

[*]
csharp_using_directive_placement = {placement}
";
```

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/Settings/SettingsCSharp8UnitTests.cs
```diff
@@ -172,6 +172,22 @@ public async Task VerifyInvalidReferenceBehaviorFromEditorConfigAsync()
...
+            var context = await this.CreateAnalysisContextFromEditorConfigAsync(settings).ConfigureAwait(false);
+
+            var styleCopSettings = context.GetStyleCopSettings(CancellationToken.None);
+
+            Assert.NotNull(styleCopSettings.OrderingRules);
+            Assert.Equal(UsingDirectivesPlacement.OutsideNamespace, styleCopSettings.OrderingRules.UsingDirectivesPlacement);
```
sharwell, line 188:
```suggestion
            Assert.Equal(expected, styleCopSettings.OrderingRules.UsingDirectivesPlacement);
```

sharwell:
Thanks for the test! I converted it to a `[Theory]` so it covers both settings.

sharwell (approved):

PR #3765, opened 2024-01-02
Update SA1025 documentation to reflect handling of multiple spaces preceding a symbol

### LINKED ISSUE #3684: SA1025 possible bug
Consider the following code:
```csharp
public enum MyEnum
{
    FirstValue,
    SecondValue,
    ThirdValue,
    FourthValue,
    FifthValue,
}

[Flags]
public enum MyEnums
{
    FirstValue  = 1 << MyEnum.FirstValue,
    SecondValue = 1 << MyEnum.SecondValue,
    ThirdValue  = 1 << MyEnum.ThirdValue,
    FourthValue = 1 << MyEnum.FourthValue,
    FifthValue  = 1 << MyEnum.FifthValue,
}
```

For readability I like to have all the `=` aligned in the flags enum. This triggers rule SA1025 due to having multiple whitespace characters in a row, however if I understand the documentation correctly this should be an exception.
Rule description for reference:
A violation of this rule occurs whenever the code contains multiple whitespace characters in a row, **unless** the characters come at the beginning or end of a line of code, following a comma or semicolon **or preceding a symbol**.

If it's not a bug but working as intended, I'd like to request that an exception be added for this case.

bjornhellander:
Fixes #3684 by updating documentation.

### DIFF
--- a/documentation/KnownChanges.md
+++ b/documentation/KnownChanges.md
@@ -123,8 +123,14 @@ StyleCop Classic allowed multiple spaces to precede a comment placed at the end 
 int x;    // comment
 ```
 
-StyleCop Analyzers does not currently make an exception to the SA1025 rule for this case, although this decision is
-still under review.
+It also allowed multiple spaces preceding a symbol, such as in the second line of the following code:
+
+```csharp
+int xyz = 1;
+int w   = 1;
+```
+
+StyleCop Analyzers does not currently make an exception to the SA1025 rule for these cases.
 
 ## Readability Rules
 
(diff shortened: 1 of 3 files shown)


documentation/KnownChanges.md
```diff
@@ -123,8 +123,14 @@ StyleCop Classic allowed multiple spaces to precede a comment placed at the end
 int x;    // comment
 ```
 
-StyleCop Analyzers does not currently make an exception to the SA1025 rule for this case, although this decision is
-still under review.
+It also allowed multiple spaces preceding a symbol, such as in the second line of the following code:
+
+```csharp
+int xyz = 1;
+int w   = 1;
```
sharwell, line 130: ❔ Are we sure that the implementation of StyleCop Classic allowed this case?

sharwell (approved):
Looks good but we should make sure to verify the StyleCop Classic behavior matches its documentation (we saw some deviations during the initial port).

documentation/KnownChanges.md
sharwell: ➡️ It does appear that it excluded operator symbols from CS1025:
https://github.com/StyleCop/StyleCop/blob/7da86d484c6487c5fff1ae84cbd41c03c5c212d8/Project/Src/StyleCop.CSharp.Rules/SpacingRules.cs#L1969C100-L1969C100

PR #3767, opened 2024-01-03
Update to use a custom fixer for SA1629

### LINKED ISSUE #2957: SA1629 code fix sometimes inserts periods just before last character
SA1629 will flag docs such as these as needing to end with a period:

```cs
        /// <typeparam name="S">The service interface to retrieve</typeparam>
        /// <param name="provider">The IServiceProvider implementation</param>
        /// <returns>A reference to the service</returns>
```

The (bulk) code fix provider produced this:

```cs
        /// <typeparam name="S">The service interface to retriev.e</typeparam>
        /// <param name="provider">The IServiceProvider implementation.</param>
        /// <returns>A reference to the servic.e</returns>
```

Note this doesn't always repro. When I undo just a few lines to their period-less state and reapply the code fix, the period goes to the right place. It only happens when there are a massive number (possibly hundreds) of fixes applied in bulk.

bjornhellander:
Hopefully fixes #2957.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1629CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1629CodeFixProvider.cs
@@ -5,6 +5,7 @@
 
 namespace StyleCop.Analyzers.DocumentationRules
 {
+    using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Composition;
     using System.Threading;
@@ -29,7 +30,7 @@ namespace StyleCop.Analyzers.DocumentationRules
         /// <inheritdoc/>
         public override FixAllProvider GetFixAllProvider()
         {
-            return CustomFixAllProviders.BatchFixer;
+            return FixAll.Instance;
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/DocumentationRules/SA1629CodeFixProvider.cs
```diff
@@ -54,10 +55,46 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                foreach (var diagnostic in diagnostics)
+                {
+                    changes.Add(GetTextChange(diagnostic));
+                }
+
+                changes.Sort((left, right) => left.Span.Start.CompareTo(right.Span.Start));
+
+                var text = await document.GetTextAsync().ConfigureAwait(false);
+                var tree = await document.GetSyntaxTreeAsync().ConfigureAwait(false);
+                return await tree.WithChangedText(text.WithChanges(changes)).GetRootAsync().ConfigureAwait(false);
```
sharwell, line 96: ❓  Should each of these lines have a cancellation token passed (from `fixAllContext`)?

bjornhellander: That's probably good. Updated.

sharwell (approved):

bjornhellander:
Is there anything else I can do here, @sharwell?

PR #3768, opened 2024-01-03
Update SA1515 to not require a blank line before a comment at the start of a collection expression

### LINKED ISSUE #3766: SA1515 should not trigger for C# 12 collection expressions
```cs
string[] elements =
[
  // Hydrogen <-- SA1515 Single-line comment should be preceded by blank line
  "H",

  // Helium
  "He",
]
```

This doesn't seem right to me, as the same code using the curly bracket syntax doesn't trigger the warning.

https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions

1.2.0-beta.556

bjornhellander:
Fixes #3766


bjornhellander:
I found some uncovered old code: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/7dae8e97270e0052200c90df7f58b0d789acb100/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1515SingleLineCommentMustBePrecededByBlankLine.cs#L262
I am wondering if this is something that needs to be handled for the opening bracket as well.

PR #3777, opened 2024-01-10
Fix SA1611 not triggering on class primary constructors

### LINKED ISSUE #3770: SA1611 is not raised for class primary constructor parameters that are missing documentation
For a traditional constructor in the below example, SA1611 is raised on the constructor parameter `myString` to indicate that it does not have a <param> entry in the constructor documentation:

```cs
/// <summary>
/// Class.
/// </summary>
public class Class
{
    /// <summary>
    /// Initializes a new instance of the <see cref="Class"/> class.
    /// </summary>
    public Class(string myString) { }
}
```

However, when making a similar mistake using a primary constructor, SA1611 is not reported:

```cs
/// <summary>
/// Class 2.
/// </summary>
public class ClassTwo(string myString) { }
```

I would expect SA1611 to be reported unless the parameter is appropriately documented beneath the class summary:

```cs
/// <summary>
/// Class 2.
/// </summary>
/// <param name="myString">My string.</param>
public class ClassTwo(string myString) { }
```

Edit: I am on version `1.2.0-beta.556`

SapiensAnatis:
Fixes #3770

`SA1611ElementParametersMustBeDocumented` will return early for class primary constructors with missing parameter documentation, since `GetParameters(node)` fails to acquire the parameter list of the constructor. Adds an additional check for the `ParameterList` of a `ClassDeclarationSyntax` to `GetParameters` to allow this diagnostic to be reported correctly.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/DocumentationRules/SA1611CSharp12UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/DocumentationRules/SA1611CSharp12UnitTests.cs
@@ -3,9 +3,109 @@
 
 namespace StyleCop.Analyzers.Test.CSharp12.DocumentationRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.Test.CSharp11.DocumentationRules;
+    using Xunit;
+    using static StyleCop.Analyzers.Test.Verifiers.CustomDiagnosticVerifier<StyleCop.Analyzers.DocumentationRules.SA1611ElementParametersMustBeDocumented>;
 
     public partial class SA1611CSharp12UnitTests : SA1611CSharp11UnitTests
     {
+        public static TheoryData<string> NonRecordDeclarationKeywords { get; } = new TheoryData<string>() { "class", "struct" };
+
+        [Theory]
+        [MemberData(nameof(NonRecordDeclarationKeywords))]
+        [WorkItem(3770, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3770")]
+        public async Task TestPrimaryConstructorMissingParametersAsync(string keyword)
+        {
+            var testCode = $@"
+/// <summary>
+/// Type.
+/// </summary>
+public {keyword} C(int {{|#0:param1|}}, string {{|#1:param2|}}) {{ }}";
+
+            DiagnosticResult[] expectedResults = new[]
+            {
+                Diagnostic().WithLocation(0).WithArguments("param1"),
+                Diagnostic().WithLocation(1).WithArguments("param2"),
+            };
+
+            await VerifyCSharpDiagnosticAsync(testCode, expectedResults, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [MemberData(nameof(NonRecordDeclarationKeywords))]
+        [WorkItem(3770, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3770")]
+        public async Task TestPrimaryConstructorPartiallyMissingParametersAsync(string keyword)
+        {
+            var testCode = $@"
+/// <summary>
+/// Type.
+/// </summary>
+/// <param name=""param1"">Parameter one.</param>
+public {keyword} C(int param1, string {{|#0:param2|}}) {{ }}";
+
+            await VerifyCSharpDiagnosticAsync(testCode, new[] { Diagnostic().WithLocation(0).WithArguments("param2") }, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Theory]
+        [MemberData(nameof(NonRecordDeclarationKeywords))]
+        [WorkItem(3770, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3770")]
+        public async Task TestPrimaryConstructorNoMissingParametersAsync(string keyword)
+        {
+            var testCode = $@"
+/// <summary>
+/// Type.
+/// </summary>
+/// <param name=""param1"">Parameter one.</param>
+/// <param name=""param2"">Parameter two.</param>
+public {keyword} C(int param1, string param2) {{ }}";
+
+            await VerifyCSharpDiagnosticAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, CancellationToken.None).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/DocumentationRules/SA1611CSharp12UnitTests.cs
```diff
@@ -3,9 +3,61 @@
...
+            var testCode = @"
+/// <summary>
+/// Class.
+/// </summary>
+public class C(int param1, string param2) { }";
+
+            DiagnosticResult[] expectedResults = new[]
+            {
+                Diagnostic().WithLocation(5, 20).WithArguments("param1"),
+                Diagnostic().WithLocation(5, 35).WithArguments("param2"),
```
sharwell, line 28: These should use the markup syntax for location:

```suggestion
                Diagnostic().WithLocation(0).WithArguments("param1"),
                Diagnostic().WithLocation(1).WithArguments("param2"),
```

sharwell (changes requested):
Can you add a test to SA1611CSharp9UnitTests covering records (showing that records do not report a diagnostic in this case)?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1611ElementParametersMustBeDocumented.cs
```diff
@@ -106,7 +107,8 @@ private static IEnumerable<ParameterSyntax> GetParameters(SyntaxNode node)
         {
             return (node as BaseMethodDeclarationSyntax)?.ParameterList?.Parameters
                 ?? (node as IndexerDeclarationSyntax)?.ParameterList?.Parameters
-                ?? (node as DelegateDeclarationSyntax)?.ParameterList?.Parameters;
+                ?? (node as DelegateDeclarationSyntax)?.ParameterList?.Parameters
+                ?? (node as ClassDeclarationSyntax)?.ParameterList()?.Parameters;
```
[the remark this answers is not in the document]
sharwell: I believe this should use `TypeDeclarationSyntax` instead of `ClassDeclarationSyntax`

SapiensAnatis:
I'll continue looking at this at the weekend, and and aim to make records with missing documentation raise SA1600.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/DocumentationRules/SA1611CSharp12UnitTests.cs
```diff
@@ -3,9 +3,109 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.Test.CSharp11.DocumentationRules;
+    using Xunit;
+    using static StyleCop.Analyzers.Test.Verifiers.CustomDiagnosticVerifier<StyleCop.Analyzers.DocumentationRules.SA1611ElementParametersMustBeDocumented>;
 
     public partial class SA1611CSharp12UnitTests : SA1611CSharp11UnitTests
     {
+        [Fact]
```
bjornhellander, line 15: The tests for class and struct seem to be identical except the keyword, which is natural. You can declare a Theory instead of a Fact and cover both class and struct with less test code.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1611ElementParametersMustBeDocumented.cs
```diff
@@ -84,6 +86,12 @@ protected override void HandleCompleteDocumentation(SyntaxNodeAnalysisContext co
             }
 
             var node = context.Node;
+            if (node.IsKind(SyntaxKindEx.RecordDeclaration))
```
bjornhellander, line 89: record structs have a separate SyntaxKind, so I suspect this will not behave as expected. I suggest you add a test for record structs as well, just to be sure.

bjornhellander: The test classes are inherited (and extended, when needed) from one language version to the next, so if you keep it in the c# 9 test project, it will still execute for c# 10 but in that situation it will run for record struct and record class as well, since the property will return all three in that project.

SapiensAnatis: Ok, I've made the CSharp9 test use the `CommonMemberData`.

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/DocumentationRules/SA1611CSharp12UnitTests.cs
SapiensAnatis: I've changed all of my tests to use the markup syntax for location -- thanks for the explanation as I'm not intricately familiar with analyzer testing 🙂

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1611ElementParametersMustBeDocumented.cs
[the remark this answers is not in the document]
SapiensAnatis: I've changed this over to `TypeDeclarationSyntax` and added an explicit early return for record types in the include and regular documentation analysis methods.

SapiensAnatis:
I've tidied up my changes relating to `class` and `struct` primary constructors according to the above review comments -- thanks both for your time! but did not manage to make much progress in making `record` / `record struct` declarations with missing param tags raise SA1600.

It feels like some refactoring is warranted there -- essentially I think I'd want the logic from `ElementDocumentationBase.HandleDeclaration` to be able to handle the `inheritdoc` and `include` cases but of course would not want to copy and paste it straight over. 

I'm unfamiliar with the codebase so would appreciate some guidance here -- would it be best to lift that logic up into the `XmlCommentHelper`? Or could we reconfigure `SA1600ElementsMustBeDocumented` to inherit from `ElementDocumentationBase`?

It does feel like it would be a fairly complicated change, so perhaps it would be best to keep this pull request focused on class and struct primary constructors, and then tackle record primary constructors separately.

bjornhellander:
Leaving SA1600 for another pull request seems perfectly reasonable to me

sharwell (approved):

PR #3782, opened 2024-01-21
Update SA1316 to not trigger in tuple types which are part of a member's signature, if that member is an override

### LINKED ISSUE #3781: SA1316: Inactionable diagnostic in method overrides
When a virtual method declared in a reference assembly returns a tuple with named values that are not PascalCase, overriding them in a compilation that includes stylecop analyzers warns that I must change them to pascal case. Yet doing so produces [CS8139](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs8139?f1url=%3FappId%3Droslyn%26k%3Dk(CS8139)). 

SA1316 should _not_ activate on method overrides. 

StyleCop.Analyzers version: 1.2.0.556

bjornhellander:
Partly fixes #3781.
Handles originally reported case (overrides), but not the general case of implementing an interface.


bjornhellander:

bjornhellander:
Rebased and updated to also handle members implemented from an interface, making it a complete fix for the linked issue.
Also moved a string allocation to make it not occur as often.

Ready for review.

Issue #3780, opened 2024-01-16
SA1600 is not raised for record primary constructors that are missing param tags


SapiensAnatis:
Today, the following code will not cause any Stylecop diagnostics to be raised:

```csharp
/// <summary>
/// Record.
/// </summary>
public record MyRecord(int Parameter);
```

This is in spite of the missing `<param>` tag for `Parameter`. Stylecop should report a diagnostic unless this is included:

```csharp
/// <summary>
/// Record.
/// </summary>
/// <param name="Parameter">Parameter.</param>
public record MyRecord(int Parameter);
```

Because `Parameter` declares a property as well as a constructor parameter, it was agreed in #3770 that it should report SA1600 instead of the usual SA1611 for missing `<param>` tags.

sharwell:
I think there are some versions of Roslyn where callbacks execute twice, resulting in duplicate diagnostics. You can include the diagnostic twice in the expected results and I can investigate further.

PR #3783, opened 2024-01-22
Draft: Make SA1600 trigger on undocumented record primary ctor parameters

SapiensAnatis:
Add analysis of XML <param> tags to SA1600 so that it is raised on record primary constructor parameters. Closes https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3780.

This is in draft as I had a few questions:

- All of my diagnostics are being reported twice in the C# 9 unit tests. I've baked these into the expectation, which is fine for C# 9 and C# 10, but causes C# 11 and C# 12 to fail after reporting the correct number of diagnostics.
- I've added a diagnostic verifier method that passes custom XML files to the C# 9 unit test file, rather than the base unit tests, because the base SA1600 tests pass in `this.LanguageVersion` to the verification. I wasn't quite sure how (or if) I should adapt the custom verifier method to accept this.
- The logic in `SA1600ElementsMustBeDocumented.GetDocumentedParameters` is more or less entirely copied from `ElementDocumentationBase.HandleDeclaration`. Would it be worth extracting this into a common helper, or are they sufficiently unique to leave separate?

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
@@ -6,8 +6,12 @@
 namespace StyleCop.Analyzers.DocumentationRules
 {
     using System;
+    using System.Collections;
+    using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Linq;
+    using System.Reflection.Metadata;
+    using System.Xml.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -143,11 +147,29 @@ namespace StyleCop.Analyzers.DocumentationRules
 
                 Accessibility declaredAccessibility = declaration.GetDeclaredAccessibility(context.SemanticModel, context.CancellationToken);
                 Accessibility effectiveAccessibility = declaration.GetEffectiveAccessibility(context.SemanticModel, context.CancellationToken);
-                if (NeedsComment(settings.DocumentationRules, declaration.Kind(), declaration.Parent.Kind(), declaredAccessibility, effectiveAccessibility))
+                if (!NeedsComment(settings.DocumentationRules, declaration.Kind(), declaration.Parent.Kind(), declaredAccessibility, effectiveAccessibility))
                 {
-                    if (!XmlCommentHelper.HasDocumentation(declaration))
+                    return;
+                }
+
+                if (!XmlCommentHelper.HasDocumentation(declaration))
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation()));
+                }
+
+                if (context.Node is TypeDeclarationSyntax typeDeclaration && RecordDeclarationSyntaxWrapper.IsInstance(typeDeclaration))
+                {
+                    RecordDeclarationSyntaxWrapper recordDeclaration = typeDeclaration;
+
+                    string[] documentedParameterNames = GetDocumentedParameters(context, typeDeclaration, out bool hasInheritdoc);
+                    IEnumerable<ParameterSyntax> parameters = recordDeclaration.ParameterList?.Parameters ?? Enumerable.Empty<ParameterSyntax>();
+
+                    foreach (var parameter in parameters)
                     {
-                        context.ReportDiagnostic(Diagnostic.Create(Descriptor, declaration.Identifier.GetLocation()));
+                        if (!hasInheritdoc && !documentedParameterNames.Contains(parameter.Identifier.ValueText))
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(Descriptor, parameter.Identifier.GetLocation()));
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


bjornhellander:
You can add a virtual method in the test class to handle different behavior in different Roslyn versions.
See for example TestPositionalRecord1Async. I think it handles the exact same case.

bjornhellander:
> I've added a diagnostic verifier method that passes custom XML files to the C# 9 unit test file, rather than the base unit tests, because the base SA1600 tests pass in this.LanguageVersion to the verification. I wasn't quite sure how (or if) I should adapt the custom verifier method to accept this.

Sorry, but I didn't understand what you meant. Note that the verifier class contains a number of overloads of VerifyCSharpDiagnosticAsync, with different parameters, with and without language version. Why did you need to add this method?

SapiensAnatis:
> Why did you need to add this method?

I added it to pass in the xml files so that I could test the included documentation case, following the example of SA1611. Is there a different way to test this?

bjornhellander:
> > Why did you need to add this method?
> 
> I added it to pass in the xml files so that I could test the included documentation case, following the example of SA1611. Is there a different way to test this?

Sorry! I completely misread your original comment and that's why I didn't understand it. I focused on your mention of LanguageVersion and got confused. Never mind. Adding it in that file seems fine to me. It can be moved later on if needed.

I will try to look at this again tonight, unless you have already received another response by then.

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/RecordDeclarationSyntaxWrapper.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace StyleCop.Analyzers.Lightup;
+
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+internal partial struct RecordDeclarationSyntaxWrapper : ISyntaxWrapper<TypeDeclarationSyntax>
+{
+    public static implicit operator RecordDeclarationSyntaxWrapper(TypeDeclarationSyntax node)
```
bjornhellander, line 10: This corresponds to a conversion from a base type to a sub type. Adding an implicit conversion operator for that is not a good thing in my book.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
```diff
@@ -341,6 +363,61 @@ public static void HandleEventFieldDeclaration(SyntaxNodeAnalysisContext context
...
+
+                if (hasIncludedDocumentation)
+                {
+                    var declaredSymbol = context.SemanticModel.GetDeclaredSymbol(typeDeclarationSyntax, context.CancellationToken);
+                    var rawDocumentation = declaredSymbol?.GetDocumentationCommentXml(expandIncludes: true, cancellationToken: context.CancellationToken);
+                    var includedDocumentation = XElement.Parse(rawDocumentation ?? "<doc></doc>", LoadOptions.None);
+
+                    if (includedDocumentation.Nodes().OfType<XElement>().Any(element => element.Name == XmlCommentHelper.InheritdocXmlTag))
+                    {
+                        hasInheritdoc = true;
```
bjornhellander, line 394: This code is not executed in tests

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
```diff
@@ -341,6 +363,61 @@ public static void HandleEventFieldDeclaration(SyntaxNodeAnalysisContext context
...
+                {
+                    IEnumerable<XmlNodeSyntax> xmlNodes = documentation.Content.GetXmlElements(XmlCommentHelper.ParamXmlTag);
+                    return xmlNodes.Select(XmlCommentHelper.GetFirstAttributeOrDefault<XmlNameAttributeSyntax>)
+                        .Where(x => x != null)
+                        .Select(x => x.Identifier.Identifier.ValueText)
+                        .ToArray();
+                }
+            }
+
+            private static class EmptyArray<T>
```
bjornhellander, line 417: I would move it to the Helpers folder so it can be used in other places as well if needed

bjornhellander:
> The logic in SA1600ElementsMustBeDocumented.GetDocumentedParameters is more or less entirely copied from ElementDocumentationBase.HandleDeclaration. Would it be worth extracting this into a common helper, or are they sufficiently unique to leave separate?

I think it's sufficiently complex to unify if it can be done without too much work, but I would wait for @sharwell's opinion on that.

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/RecordDeclarationSyntaxWrapper.cs
SapiensAnatis: I see your point, I am happy to make it an explicit operator. With the way it is being used I could also do something like:

```cs
bool RecordDeclarationSyntaxWrapper.TryWrap(TypeDeclarationSyntax node, out RecordDeclarationSyntaxWrapper result)
```

similar to `int.TryParse` and friends. I'm not sure if that's a bit out there though

SapiensAnatis:
Just a quick question: is there any issue with me responding to review feedback now, knowing my pipelines will fail before the double diagnostic problem is resolved? I am a bit clueless there and @sharwell kindly offered to look into it in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3780.

I'm not familiar with Azure Pipelines and am unsure if you guys are being charged per run or anything like that :sweat_smile:

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/RecordDeclarationSyntaxWrapper.cs
bjornhellander: Just to be clear: The explicit conversion operator already exists, so you can just remove this file and add a cast where you get an error.

bjornhellander:
> Just a quick question: is there any issue with me responding to review feedback now, knowing my pipelines will fail before the double diagnostic problem is resolved? I am a bit clueless there and @sharwell kindly offered to look into it in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3780.
> 
> I'm not familiar with Azure Pipelines and am unsure if you guys are being charged per run or anything like that 😅

I think that this will do it:
> You can add a virtual method in the test class to handle different behavior in different Roslyn versions.
> See for example TestPositionalRecord1Async. I think it handles the exact same case.

I just checked your tests and it looks to me like this should work. So instead of adding the expected diagnostics inside the test method, extract that code to a method per test (`GetExpectedResult<TestName>`) and override that one in every test project where the compiler behavior changes. In your case it should be the c# 11 project. Have a look at `TestPositionalRecord1Async` and update the failing ones the same way. Hopefully that will work.

SapiensAnatis:
Ah, ok, thanks @bjornhellander. I wasn't sure if it was a case of needing to fix the analyzer or the tests, but it sounds like this is Roslyn behaviour so updating the tests is probably all that we can do. I will make those changes when I next take a look at this and hopefully have something ready for review.

StyleCop.Analyzers/StyleCop.Analyzers/Lightup/RecordDeclarationSyntaxWrapper.cs
SapiensAnatis: Ah, understood. Removed this file and used the explicit operator.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
SapiensAnatis: It handles the case where your XML comment includes an XML file, which itself specifies an `<inheritdoc />`.

I've added a test that should cover this case.

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1600ElementsMustBeDocumented.cs
SapiensAnatis: I did a bit of refactoring and was no longer using this, so I've removed it

SapiensAnatis:
Regarding unifying `ElementDocumentationBase.HandleDeclaration` and the new code: I think it will be quite difficult to do unless SA1600 is adapted to inherit from `ElementDocumentationBase`, mainly because any changes to the signature of `HandleDeclaration` would require updates in quite a few analyzers. I wasn't sure whether adding that subclass would be advisable.

For now, I've added an extension method to `XmlCommentHelper` which returns a parsed `XmlComment` class that contains data from the XML comment. I was taking notes from the `XmlFileHeader` class here, but I like this as returning an object feels a bit less clunky than detecting `inheritdoc` tags using a `out bool isInheritdoc` parameter. 

I also felt this was a bit better than pasting into a private method; a 'domain model' class to represent XML comments that are parsed either from the Roslyn syntax tree or from `XElements` from an included file feels like a nice abstraction, which could also be extended, should any other analyzers not inheriting from `ElementDocumentationBase` want to perform analysis of XML comments in the future.

PR #3800, opened 2024-02-09
Update SA1110 to also check the parameter list in primary constructors

### LINKED ISSUE #3784: SA1110 does not trigger on primary constructors
I expect the following code to trigger SA1110 (opening parenthesis must be on declaration line), but it does not:
```
public record MyType
    (int X, int y)
{
}
```

Tested with version 1.2.0-beta.556

bjornhellander:
Fixes #3784

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/ReadabilityRules/SA1110CSharp9UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/ReadabilityRules/SA1110CSharp9UnitTests.cs
@@ -3,9 +3,66 @@
 
 namespace StyleCop.Analyzers.Test.CSharp9.ReadabilityRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.Test.CSharp8.ReadabilityRules;
+    using StyleCop.Analyzers.Test.Helpers;
+    using Xunit;
+    using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
+        StyleCop.Analyzers.ReadabilityRules.SA1110OpeningParenthesisMustBeOnDeclarationLine,
+        StyleCop.Analyzers.SpacingRules.TokenSpacingCodeFixProvider>;
 
     public partial class SA1110CSharp9UnitTests : SA1110CSharp8UnitTests
     {
+        [Theory]
+        [MemberData(nameof(CommonMemberData.TypeKeywordsWhichSupportPrimaryConstructors), MemberType = typeof(CommonMemberData))]
+        [WorkItem(3784, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3784")]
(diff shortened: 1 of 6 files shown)


bjornhellander:

sharwell (approved):

StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp9/ReadabilityRules/SA1110CSharp9UnitTests.cs
```diff
@@ -3,9 +3,66 @@
...
+        {
+            var testCode = $@"
+{typeKeyword} Foo
+    {{|#0:(|}}int x)
+{{
+}}";
+
+            var fixedCode = $@"
+{typeKeyword} Foo(
+    int x)
```
sharwell, line 51: 💡 Can you also add tests for inheritance cases?

bjornhellander: Just record declarations that inherit from another record? Could you describe some more what you are after?

sharwell: Yes, reference type records and classes with primary constructors allow a base type to be specified, and arguments can be provided to the base class.

```cs
record BaseRecord(int A);

record DerivedRecord(int A, int B) : BaseRecord(A);
```

```cs
class BaseClass(int a);

class DerivedClass(int a, int b) : BaseClass(int a);
```

bjornhellander: So the tests that you wanted was regarding the placement of the opening parenthesis **in the call to the base constructor**?

sharwell: Yes, that's correct. I believe SA1110 handles this for the equivalent `base(...)` call in constructors.

bjornhellander: Fixed

StyleCop.Analyzers/StyleCop.Analyzers.Test/Helpers/CommonMemberData.cs
```diff
@@ -137,5 +137,13 @@ public static IEnumerable<object[]> TypeKeywordsWhichSupportPrimaryConstructors
                 }
             }
         }
+
+        public static IEnumerable<object[]> ReferenceTypeKeywordsWhichSupportPrimaryConstructors
```
sharwell, line 141: 💡 Consider rewriting this the other way around so `TypeKeywordsWhichSupportPrimaryConstructors` calls `ReferenceTypeKeywordsWhichSupportPrimaryConstructors`.

sharwell (approved):

PR #3813, opened 2024-03-06
Update SA1012 to forbid space before a property pattern, when it is preceeded by an opening parenthesis. Previously only handled positional patterns correctly.

### LINKED ISSUE #3812: SA1008, SA1012: false positives for property pattern inside parentheses
Using `StyleCop.Analyzers` version `1.2.0-beta.556`

```csharp
public static class Helpers
{
    public static bool HasValidLength(string s)
    {
        // ⚠ SA1008 Opening parenthesis should not be preceded by a space.
        // ⚠ SA1012 Opening brace should be preceded by a space
        //              ↓↓
        return s is not ({ Length: < 5 } or { Length: > 10 });
    }
}
```

This is similar to the problem fixed in #3141.

bjornhellander:
Fixes #3812

Removed a condition which made us forbid a blank before a property pattern only in the case of a positional pattern.


bjornhellander:
Ready for review

sharwell (approved):

PR #3820, opened 2024-03-20
Update SA1513 to not require a blank line before linq 'into' keyword

DoobieAsDave:
No blank line needed before LINQ 'into' keyword when using anonymous object creation syntax. This PR fixes the issue.


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/LayoutRules/SA1513CSharp12UnitTests.cs
```diff
@@ -32,5 +32,76 @@ public class Foo
 
             await VerifyCSharpFixAsync(testCode, DiagnosticResult.EmptyDiagnosticResults, testCode, CancellationToken.None).ConfigureAwait(false);
         }
+
+        [Fact]
+        public async Task TestJoinIntoClauseSyntaxQueryExpressionAsync()
```
sharwell, line 37: 💡 Since this only needs features from C# 6, it should move all the way up to **SA1513CSharpUnitTests.cs**.

PR #3823, opened 2024-04-06
Update SA1003 to handle the null-coalescing assignment operator, the unsigned right-shift operator, the unsigned right-shift assignment operator and the null-forgiving operator

### LINKED ISSUE #3822: SA1003 should validate null-coalescing assignment operator
The rule [SA1003](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1003.md) should validate spaces around the null-coalescing assignment operator (`??=`)

The following code raise no warnings with stylecop 1.2.0-beta.556:
```csharp
value??=string.Empty;
```

While the expected style like with any other operator should be:
```csharp
value ??= string.Empty;
```

bjornhellander:
Fixes #3822. Also found some more operators that were not being handled by SA1003.


bjornhellander:
Ready for review

sharwell (approved):

PR #3832, opened 2024-04-27
Update some tests to not use a fixed language version

bjornhellander:
I found some more tests where a specific language version was used in all test projects.


bjornhellander:
Ready for review

sharwell (approved):

PR #3833, opened 2024-04-28
Update SA1612 to also check primary constructor parameters

### LINKED ISSUE #3786: SA1612 does not trigger on primary constructor
I expect the following code to trigger SA1612 (element parameter documentation must match element parameters), but it does not:

```
/// <summary>
/// Stuff.
/// </summary>
/// <param name="Y">Foo.</param>
/// <param name="X">Bar.</param>
public record Class8(int X, int Y)
{
}
```

Tested with version 1.2.0-beta.556

bjornhellander:
Fixes #3786


bjornhellander:
Ready for review

sharwell (approved):

PR #3836, opened 2024-05-11
Update SA1004 to allow doc comments to start with the in, out and ref keywords

### LINKED ISSUE #3817: SA1004 false-positive when first token on XML doc line is out keyword from wrapped params in <see cref="...."/>
Example:
```cs
/// <summary>
/// This is a summary.
/// </summary>
/// <remarks>
/// <para>
/// Description of some remarks that refer to a method: <see cref="SomeMethod(int, int,
/// out string)"/>.</para>
/// </remarks>
public class SomeClass
{
    public void SomeMethod(int x, int y, out string z) => z = string.Empty;
}
```
image

bjornhellander:
Fixes #3817

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1004UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1004UnitTests.cs
@@ -1,13 +1,14 @@
 ﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-#nullable disable
-
 namespace StyleCop.Analyzers.Test.SpacingRules
 {
+    using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.Testing;
+    using StyleCop.Analyzers.Lightup;
     using StyleCop.Analyzers.SpacingRules;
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1004UnitTests.cs
```diff
@@ -213,5 +228,33 @@ private void Method1(int x, int y)
...
+
+        [Theory]
+        [MemberData(nameof(ParameterModifiers))]
+        [WorkItem(3817, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3817")]
+        public async Task TestParameterModifierFirstOnLineAsync(string keyword)
+        {
+            string testCode = $@"
+/// <summary>
+/// Description of some remarks that refer to a method: <see cref=""SomeMethod(int, int,
+/// {keyword} string)""/>.
```
sharwell, line 240: Do we also need to check this wrapping?

```
/// <see cref="SomeMethod(int, int, ref
/// readonly string)"/>
```

sharwell: To me it seems similar to the current report. I think most users would not wrap a comment in the middle of a `cref` attribute, but automatic word wrapping tools can do it.

bjornhellander: Ok. Added.

sharwell (approved):

PR #3839, opened 2024-05-15
Do not warn if a comma follows a preprocessor directive

### LINKED ISSUE #3816: Do not warn SA1001 (Commas should be spaced correctly) if it follows a preprocessor directive
Example where I don't think I should need to suppress the diagnostic:
```cs
partial struct Money : IFormattable
#if !NETSTANDARD
#pragma warning disable SA1001 // Commas should be spaced correctly
    , ISpanFormattable
#pragma warning restore SA1001
#endif
```

MattFromRVA:
Fixes #3816

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
@@ -349,6 +349,27 @@ public class TestClass
             await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
         }
 
+        [Fact]
+        [WorkItem(3816, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3816")]
+        public async Task TestCommaFollowingPreprocessorDirectiveAsync()
+        {
+            var testCode = @"
+interface IFormattable {}
+interface ISpanFormattable {}
+
+partial struct Money : IFormattable
+#if !NETSTANDARD
+    , ISpanFormattable
+#endif
+{
+}
+";
+
+            var expected = DiagnosticResult.EmptyDiagnosticResults;
+
+            await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
+
         private Task TestCommaInStatementOrDeclAsync(string originalStatement, DiagnosticResult expected, string fixedStatement)
         {
             return this.TestCommaInStatementOrDeclAsync(originalStatement, new[] { expected }, fixedStatement);
--- a/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1001CommasMustBeSpacedCorrectly.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/SpacingRules/SA1001CommasMustBeSpacedCorrectly.cs
@@ -7,6 +7,7 @@ namespace StyleCop.Analyzers.SpacingRules
 {
     using System;
     using System.Collections.Immutable;
+    using System.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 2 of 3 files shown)


bjornhellander (changes requested):

MattFromRVA:
Thanks for the feedback @bjornhellander! My changes are ready for review. I believe the pipeline errors are not related to my code changes?

bjornhellander (approved):
Looks good to me. And no, the failing tests do not look like your fault.

bjornhellander:
@sharwell, do you have any additional comments?

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
```diff
@@ -349,6 +350,209 @@ public void TestMethod()
...
+
+partial struct Money : IFormattable
+    , ISpanFormattable
+{
+}
+";
+
+            var expected = new[]
+            {
+                DiagnosticResult.CompilerWarning("SA1001").WithSpan(6, 5, 6, 6).WithArguments(" not", "preceded"),
```
sharwell, line 451: I would change the diagnostic construction to use `Diagnostic()` and `WithLocation(int)`.

```suggestion
                Diagnostic().WithLocation(0).WithArguments(" not", "preceded"),
```

Then update the `testCode` to indicate the diagnostic location by wrapping it in `{|#0:` and `|}`.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
```diff
@@ -349,6 +350,209 @@ public void TestMethod()
...
+            };
+
+            await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
```
sharwell, line 454: 💡 All tests that have one or more diagnostics should also define the `fixedCode` that the code fix produces.

sharwell:
Two small things to update but otherwise looks good.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
MattFromRVA: I added the `fixedCode` to the tests but realized that the code fix provider does not remove the whitespace, causing the tests to fail. Please let me know if I overlooked something.

Since this appears to be a separate issue from skipping the warning, I am happy to open a new issue and submit a PR to address it.

bjornhellander: I tried the code from that particular test case and it seems like it can be fixed just fine.

But I understand if the other tests (all having directives before the comma) are not handled by the code fix, since it would have to move the comma to before the directives. Currently it only replaces tokens and can therefore not handle this case.

A separate issue seems fine to me, if we intend to change this behavior at all. Does not feel obvious to me at all.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
```diff
@@ -8,6 +8,7 @@ namespace StyleCop.Analyzers.Test.SpacingRules
     using System;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft;
```
bjornhellander, line 11: If you expand the test case that can verify the fix as well, you can remove this using statement at the same time.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
MattFromRVA: Ok I see what I did wrong here. Test has been updated.

StyleCop.Analyzers/StyleCop.Analyzers.Test/SpacingRules/SA1001UnitTests.cs
```diff
@@ -349,6 +349,120 @@ public void TestMethod()
...
+{
+}
+";
+
+            var expected = DiagnosticResult.EmptyDiagnosticResults;
+
+            await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
```
sharwell:
```suggestion
        [Fact]
        [WorkItem(3816, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3816")]
```

sharwell (approved):

PR #3842, opened 2024-05-17
Update `stylecop.json` schema to have correct default for `headerDecoration`.

alexrp:
The default is documented to be the empty string, and since it's typed as `string`, `null` is not actually a valid value and will fail to validate.


sharwell:
The default values need to also be updated in these two locations:

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/17b613ddbc2f0f8313727953048cea30c36b7482/StyleCop.Analyzers/StyleCop.Analyzers/Settings/ObjectModel/DocumentationSettings.cs#L119

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/17b613ddbc2f0f8313727953048cea30c36b7482/StyleCop.Analyzers/StyleCop.Analyzers/Settings/ObjectModel/DocumentationSettings.cs#L270

sharwell (approved):

PR #3859, opened 2024-06-06
Correct SA1629 to add a period if the documentation ends in an xml entity. Previously replaced the semicolon in the xml entity.

### LINKED ISSUE #3802: `SA1629` eats ending semicolon
``` csharp
/// <summary>
/// Indicates that this property is a dictionary containing SAML attribute values.
/// </summary>
/// <remarks>This attribute should only be placed on a Dictionary&lt;string, string?&gt;</remarks>
[AttributeUsage(AttributeTargets.Property)]
public sealed class SamlAttributesAttribute : System.Attribute;
```

The fixer for `SA1629` will replace the `;` with `.` instead of appending it, thereby breaking the XML syntax.

bjornhellander:
Fixes #3802.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
@@ -9,6 +9,7 @@ namespace StyleCop.Analyzers.DocumentationRules
     using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Linq;
+    using System.Text.RegularExpressions;
     using System.Xml.Linq;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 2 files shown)


StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
```diff
@@ -131,8 +134,9 @@ private static void HandleSectionOrBlockXmlElement(SyntaxNodeAnalysisContext con
                             {
                                 int spanStart = textToken.SpanStart + textWithoutTrailingWhitespace.Length;
                                 ImmutableDictionary<string, string> properties = null;
-                                if (textWithoutTrailingWhitespace.EndsWith(",", StringComparison.Ordinal)
-                                    || textWithoutTrailingWhitespace.EndsWith(";", StringComparison.Ordinal))
+                                if (textWithoutTrailingWhitespace.EndsWith(",", StringComparison.Ordinal) ||
+                                    (textWithoutTrailingWhitespace.EndsWith(";", StringComparison.Ordinal) &&
```
sharwell, line 138: 💡 Keep `||` and `&&` at the beginning of the line

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
```diff
@@ -131,8 +134,9 @@ private static void HandleSectionOrBlockXmlElement(SyntaxNodeAnalysisContext con
...
+                                     !XmlEntityRegex.IsMatch(textWithoutTrailingWhitespace)))
```
sharwell, line 139: 💭 Can we not use `textToken` here?

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
bjornhellander: Reverted

StyleCop.Analyzers/StyleCop.Analyzers/DocumentationRules/SA1629DocumentationTextMustEndWithAPeriod.cs
bjornhellander: Updated to check syntax kind instead

sharwell (approved):

PR #3863, opened 2024-06-26
Fix SA1514: Do not report when documenting types declared in the global namespace

### LINKED ISSUE #3849: SA1514 is falsely reported when documenting types declared in the global namespace
```cs
/// <summary>
/// X.
/// </summary>
public class K
{
}
```
`SA1514: Element documentation header should be preceded by blank line` is reported on the first `///` in the file. Adding the new line resolves the diagnostic but instead raises `SA1517: Code should not contain blank lines at start of file`, leading to an infinite loop. Furthermore, the class has been intentionally declared in the global namespace, so there is no space for a namespace declaration

MattFromRVA:
Fixes #3849

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1514UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1514UnitTests.cs
@@ -1151,5 +1151,23 @@ public class TestClass
 
             await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
         }
+
+        [Fact]
+        [WorkItem(3849, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3849")]
+        public async Task TestClassInGlobalNamespaceAsync()
+        {
+            var testCode = @"
+/// <summary>
+/// X.
+/// </summary>
+public class TestClass
+{
+}
+";
+
+            var expected = DiagnosticResult.EmptyDiagnosticResults;
+
+            await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
     }
 }
--- a/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1514ElementDocumentationHeaderMustBePrecededByBlankLine.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1514ElementDocumentationHeaderMustBePrecededByBlankLine.cs
@@ -170,6 +170,18 @@ namespace StyleCop.Analyzers.LayoutRules
                     // no leading blank line necessary at start of scope.
                     return;
                 }
+
+                // Logic to handle global namespace case
+                if (prevToken.IsKind(SyntaxKind.None))
+                {
+                    // Check if the node is the first non-trivia element in the file
+                    var firstToken = context.Node.SyntaxTree.GetRoot().GetFirstToken();
+                    if (firstToken == context.Node.GetFirstToken())
+                    {
+                        // Node is the first element in the global namespace
+                        return;
+                    }
+                }
             }
 
             context.ReportDiagnostic(Diagnostic.Create(Descriptor, GetDiagnosticLocation(documentationHeader)));


StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1514UnitTests.cs
```diff
@@ -1151,5 +1151,23 @@ public enum TestEnum
 
             await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
         }
+
+        [Fact]
+        [WorkItem(3849, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3849")]
+        public async Task TestClassInGlobalNamespaceAsync()
+        {
+            var testCode = @"
+/// <summary>
```
sharwell, line 1160: ❓ Do we have a test for this same case, but without a newline present at the start of the file?

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1514ElementDocumentationHeaderMustBePrecededByBlankLine.cs
```diff
@@ -170,6 +170,18 @@ private static void HandleDeclaration(SyntaxNodeAnalysisContext context)
                     // no leading blank line necessary at start of scope.
                     return;
                 }
+
+                // Logic to handle global namespace case
+                if (prevToken.IsKind(SyntaxKind.None))
+                {
+                    // Check if the node is the first non-trivia element in the file
+                    var firstToken = context.Node.SyntaxTree.GetRoot().GetFirstToken();
+                    if (firstToken == context.Node.GetFirstToken())
```
bjornhellander, line 179: Is this check actually needed? If there is no previous token (line 175), can this condition really be false? You have at least no test that does not enter this if statement.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1514ElementDocumentationHeaderMustBePrecededByBlankLine.cs
```diff
@@ -170,6 +170,18 @@ private static void HandleDeclaration(SyntaxNodeAnalysisContext context)
...
+                    {
+                        // Node is the first element in the global namespace
+                        return;
```
bjornhellander, line 182: I suspect that you skip too many cases as the code looks right now. I suggest you add tests for cases where it **should** trigger even though there is no token before the documentation. I am thinking e.g. a normal comment or a directive before the documentation. And if that reveals a need for a more subtle bail-out, consider similar cases where it should *not* trigger.

MattFromRVA: Yep. Great catch. I need to review and add some more tests. Thanks

bjornhellander (approved):

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1514UnitTests.cs
```diff
@@ -1151,5 +1151,199 @@ public enum TestEnum
...
+    }
+}
+";
+
+            var expected = DiagnosticResult.EmptyDiagnosticResults;
+
+            await VerifyCSharpDiagnosticAsync(testCode, expected, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
```
sharwell:
```suggestion
        [Fact]
        [WorkItem(3849, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3849")]
```

StyleCop.Analyzers/StyleCop.Analyzers.Test/LayoutRules/SA1514UnitTests.cs
```diff
@@ -1151,5 +1151,199 @@ public enum TestEnum
...
+
+            var expected = new[]
+            {
+                Diagnostic().WithLocation(0).WithArguments(" not", "preceded"),
+            };
+
+            await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
```
sharwell:
```suggestion
        [Fact]
        [WorkItem(3849, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3849")]
```

sharwell (approved):

PR #3864, opened 2024-06-26
Add test project for c# 13 (pre-release version)

bjornhellander:


bjornhellander:
@sharwell, I created this mostly to check if there were any regressions so far. None found. Some previous Roslyn problems were fixed and some tests which have syntax errors in them got slightly different behavior, but looks good so far. If you think it is too early to merge this now, I can update this now and then until the time is right.

I added LanguageVersionEx.CSharp13 the way I expect it will be defined eventually even though it was not in place yet.

What do you think?

sharwell:
Also need to incorporate this to https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/build/build-and-test.yml

bjornhellander:
Added that one and test.yml to the solution so they are easier to remember :-)

sharwell (approved):

sharwell (approved):

PR #3872, opened 2024-07-09
Add alternative rules for SA1116 and SA1117 with stricter handling of multiline parameters

### LINKED ISSUE #3183: Possible inconsistency of SA1117

mathew-tolj:
Fixes #3183 by adding alternative rules that can be used in place of SA1116 and SA1117 to use stricter handling of multiline parameters.

This is an alternative approach to #3869.

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SX1116CodeFixProvider.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.CodeFixes/ReadabilityRules/SX1116CodeFixProvider.cs
@@ -1,0 +1,86 @@
+﻿// Copyright (c) Tunnel Vision Laboratories, LLC. All Rights Reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#nullable disable
+
+namespace StyleCop.Analyzers.ReadabilityRules
+{
+    using System.Collections.Immutable;
+    using System.Composition;
+    using System.Threading;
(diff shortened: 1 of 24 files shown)


sharwell:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3869#issuecomment-2245347136 applies here too. Fortunately, this work is not lost, as the code used for this pull request could easily form the foundation of the new rule in a different repository.

PR #3883, opened 2024-08-05
Update SA1135UsingDirectivesMustBeQualified to not crash on a UsingDirectiveSyntax without a Name

### LINKED ISSUE #3882: SA1135 throws exception for type keywords
When adding a type alias using a keyword, the analyzer throws `AD0001`, when inside a namespace

```csharp
namespace ClassLibrary
{
    using Example = string;
}
```

and also when using value tuple values

```csharp
namespace ClassLibrary
{
    using Example = (System.Collections.IList, int);
}
```

bjornhellander:
Fixes #3882.


bjornhellander:
So far, this pull request only prevents the exception. I assume that the analyzer should also be updated to check all the new types of aliases that can be created now. I have started looking at that as well, but it seems to be less trivial. Would it be ok to save that for another pull request if I create a separate issue for that, @sharwell?

PR #3905, opened 2024-11-26
Fix codecov not displaying on GitHub

eXpl0it3r:
As the image was being loaded via HTTP, GitHub seems to refuse the proxying.

> Cannot proxy the given URL


eXpl0it3r:
These tests are really flaky, at least in the Azure Pipelines... 🫤

Issue #3904, opened 2024-11-22
SA1137 is not raised in C# 12 collection expressions


ar-schraml:
The error code [`SA1137: Elements should have the same indentation`](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1137.md)  is not raised in C# 12 collection expressions.

Example:
![image](https://github.com/user-attachments/assets/7bbd9ca4-7d6a-49c2-acfd-5d17e15f9ffb)
SA1137 is correctly raised inside an object initialization (_example1_), but it is missing in collection expressions (_example2_)

<details><summary>Example code</summary>
<p>

```c#
    private readonly List<string> example1 = new()
    {
        "a",
"b",
                "c",
    };

    private readonly List<string> example2 =
    [
        "a",
"b",
                "c",
    ];
```

</p>
</details>

Mivee:
Heyha, I'd like to give it a shot, but since its also my first open source contribution in general I dont know how long it will take.

PR #3910, opened 2024-12-14
SA1137 is not raised

Mivee:
Issue URL: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3904

Collectionexpressions are detected as variable declarations and not as InitializerExpression. 

Previously they where checked based on the declaration itself (so we only have 1 Item e.g. `var myVariable = [1,2,3])` . Now, for collection initializations we get the SyntaxNode from the variable and then validate each item of the collection on its own. 

Since this is my first open source contribution feedback is very welcome. 

I also found 2 points where I'm not sure how to proceed with: 
- `Microsoft.CodeAnalysis.CSharp` is installed with V.1.2.1 @Stylecop.Analyzers. V.4.7.0 introduced  the `CollectionExpressionSyntax` that would simply my `IsCollectionExpression()` to just a type check basiclly. Also the "workaround" with that id could be removed so imo it would be way cleaner, due to beeing 3 Major Versions behind i didnt wanted to update, since i basiclly dont know the project and also what could be affected by that upgrade. 
-  locally StyleCop.Analyzers.Test.CSharp8.SpacingRules.SA1009CSharp8UnitTests.TestMissingTokenAsync failed due to having a different Windows language. A possible solution could be https://stackoverflow.com/questions/5793000/is-this-a-good-approach-for-temporarily-changing-the-current-threads-culture

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/ReadabilityRules/SA1137CSharp12UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp12/ReadabilityRules/SA1137CSharp12UnitTests.cs
@@ -3,9 +3,61 @@
 
 namespace StyleCop.Analyzers.Test.CSharp12.ReadabilityRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.Test.CSharp11.ReadabilityRules;
+    using Xunit;
+    using static StyleCop.Analyzers.Test.Verifiers.StyleCopDiagnosticVerifier<
+        StyleCop.Analyzers.ReadabilityRules.SA1137ElementsShouldHaveTheSameIndentation>;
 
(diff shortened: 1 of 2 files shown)


bjornhellander:
I don't think this is the correct way of handling this, since collection expressions can be used in other places besides declarations.
Try adding another test with e.g. an assignment statement using a collection expression, to verify that.

This triggers the corresponding handling of initializer expressions:
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/StyleCop.Analyzers/StyleCop.Analyzers/ReadabilityRules/SA1137ElementsShouldHaveTheSameIndentation.cs#L79
Unless I am missing something, collection expressions should be handled in a similar way.

SyntaxKindEx contains CollectionExpression, but the Lightup folder in the analyzer project should also be updated to include support for collection expressions. This has not yet been added to the repo as far as I can see. @sharwell would need to do this, or at least describe what to do. This is to avoid the call to ChildNodes(), which seems to work fine, but is not the normal way of accessing syntax nodes in this repository.

PR #3915, opened 2025-02-06
Update SA1013 to allow whitespace before } when it starts a line inside interpolated string

### LINKED ISSUE #3914: SA1013 Allow } to start a line in interpolated string
Currently, SA1013 interferes in situations where I would like to split very long interpolated strings into multiple lines, or when tooling like Resharper does that when formatting code.
This can be seen in the following example code:
``` csharp
var message =
    $"This a very long message with id {
        id
    }, that is easier to read when split into multiple lines ({
        numberOfLines
    } lines). Some additional expression to make it even longer: {
        (foobarbaz * xyz) / (foobarbaz + xyz)
    }";
```
Here, SA1013 generates warnings on lines starting with `}` (i.e. lines 4, 6 and 8).
The warning is that `}` should not be preceded by whitespace.
My suggestion is to not generate these warning in this case.

morsiu:
Closes #3914


sharwell:
At this time, I'm not planning to make this change. The spacing enforced here is similar to the spacing for closing parentheses in argument lists (SA1009).

PR #3927, opened 2025-05-28
update SA1518 to ignore empty (0 bytes) files

### LINKED ISSUE #3926: SA1518 should not trigger on empty files
The "file has to end with newline" should not be triggered if a file does not contain anything.

if the file size is 0 bytes, that file does not contain lines of text (as for example defined by posix: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) and thus does not have to end in a newline.

the problem occurs for example in protoc-generated c#-code - for some reason the protobuf compiler creates some empty (0bytes) files and these then trigger the error.

Obviously, if these files would contain a line marking the file as autogenerated the check would be all right, because the file would not be empty anymore.

DWIAltonaAnalytics:
closes #3926


StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SyntaxTreeHelpers.cs
```diff
@@ -72,6 +72,14 @@ public static bool IsWhitespaceOnly(this SyntaxTree tree, CancellationToken canc
                 && TriviaHelper.IndexOfFirstNonWhitespaceTrivia(firstToken.LeadingTrivia) == -1;
         }
 
+        public static bool IsEmpty(this SyntaxTree tree, CancellationToken cancellationToken)
+        {
+            var root = tree.GetRoot(cancellationToken);
+            var firstToken = root.GetFirstToken(includeZeroWidth: true);
+
+            return firstToken.IsKind(SyntaxKind.EndOfFileToken) && firstToken.FullSpan.Length == 0;
```
bjornhellander, line 80: Tests for SA1518 can be improved by checking a file with a single space or something like that. `firstToken.FullSpan.Length == 0` can currently be removed without any tests failing,

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1518UseLineEndingsCorrectlyAtEndOfFile.cs
```diff
@@ -72,6 +72,12 @@ public override void Initialize(AnalysisContext context)
 
         private static void HandleSyntaxTree(SyntaxTreeAnalysisContext context, StyleCopSettings settings)
         {
+            if (context.Tree.IsEmpty(context.CancellationToken))
```
bjornhellander, line 75: Considering how rare this is, I would have added this check as the last thing before reporting the diagnostic, to save a couple of clock cycles when no diagnostics are reported.

StyleCop.Analyzers/StyleCop.Analyzers/Helpers/SyntaxTreeHelpers.cs
DWIAltonaAnalytics: I added two test cases with white space only "files".
Of those (depending on the setting) one would fail if the `IsEmpty` helper does not check if the token is empty.

StyleCop.Analyzers/StyleCop.Analyzers/LayoutRules/SA1518UseLineEndingsCorrectlyAtEndOfFile.cs
DWIAltonaAnalytics: agreed and done.

DWIAltonaAnalytics:
No Idea why the tests are hanging?

DWIAltonaAnalytics:
I am tempted to push a non-change commit to trigger the pipeline again :)

If  any of the owners read this, please restart the pipeline.

@bjornhellander can you approve, or does that require higher level of user rights?

bjornhellander:
@DWIAltonaAnalytics, no I can't approve pull requests. @sharwell is the one you need.

sharwell (approved):

PR #4034, opened 2025-11-26
Update to support default interface members

### LINKED ISSUE #3002: Review impact of default interface members on existing rules
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#default-interface-members

### LINKED ISSUE #3551: SA1600 triggers for private/internal interface members
[Interfaces may have private and internal members](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/types/interfaces):

> Interface members are public by default, and you can explicitly specify accessibility modifiers, such as public, protected, internal, private, protected internal, or private protected. A private member must have a default implementation.

SA1600 requires documentation of these members when `documentPrivateElements: false` and `documentInternalElements: false`. For example, `InternalMethod` and `PrivateMethod` here are triggering SA1600:

```csharp
/// <summary>
/// An interface with private members.
/// </summary>
internal interface IInterfaceWithPrivateMembers
{
    internal void InternalMethod()
    {
    }

    private void PrivateMethod()
    {
    }

    /// <summary>
    /// A public method.
    /// </summary>
    public void PublicMethod() => this.PrivateMethod();
}
```

Tested with 1.2.0-beta.435.

### LINKED ISSUE #3682: SA1400 FP for delegates in interface declaration
```csharp
public interface IInterface
{
    delegate void Foo(); // <- SA1400 Element should declare an access modifier
}
```

I think it's a FP since all members of interface are public be default and we do not add access modifiers for other members (i.e. method declarations)

### LINKED ISSUE #3717: SA1600 should not trigger for explicit interface implementations

sharwell:
Fixes #3551
Fixes #3682 
Fixes #3717

Closes #3002

### DIFF
--- a/documentation/SA1202.md
+++ b/documentation/SA1202.md
@@ -40,6 +40,12 @@ rule.
 Complying with a standard ordering scheme based on access level can increase the readability and maintainability of the
 file and make it easier to identify the public interface that is being exposed from a class.
 
+### Interfaces and default interface members
+
+Default interface members (including members with implementations and static interface members) follow the same access
+ordering as class members. Within an interface, `public` members should appear before `internal`, `protected internal`,
+`protected`, `private protected`, and `private` members.
+
 ## How to fix violations
 
 To fix an instance of this violation, order the elements in the file in the order described above.
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/OrderingRules/SA1202CSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/OrderingRules/SA1202CSharp8UnitTests.cs
@@ -5,7 +5,9 @@ namespace StyleCop.Analyzers.Test.CSharp8.OrderingRules
 {
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.CodeAnalysis.Testing;
     using StyleCop.Analyzers.Test.CSharp7.OrderingRules;
+    using StyleCop.Analyzers.Test.Helpers;
     using Xunit;
     using static StyleCop.Analyzers.Test.Verifiers.StyleCopCodeFixVerifier<
         StyleCop.Analyzers.OrderingRules.SA1202ElementsMustBeOrderedByAccess,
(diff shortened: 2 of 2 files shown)


documentation/SA1202.md
```diff
@@ -40,6 +40,12 @@ rule.
 Complying with a standard ordering scheme based on access level can increase the readability and maintainability of the
 file and make it easier to identify the public interface that is being exposed from a class.
 
+### Interfaces and default interface members
+
+Default interface members (including members with implementations and static interface members) follow the same access
+ordering as class members. Within an interface, `public` members should appear before `internal`, `protected internal`,
+`protected`, `private protected`, and `private` members.
```
AArnott: Interface member ordering is _not_ mere style, but is significant when the interface is used as a COM interface. 
Should accessibility modifiers and member type sorting rules automatically be disabled when an interface is annotated with attributes typically associated with COM?

documentation/SA1400.md
```diff
@@ -25,6 +25,8 @@ C# allows elements to be defined without an access modifier. Depending upon the
 
 This rule requires an access modifier to be explicitly defined for every element. This removes the need for the reader to make assumptions about the code, improving the readability of the code.
 
+Members declared in an `interface` are implicitly `public` and do not need an explicit access modifier. This includes default interface members introduced in C# 8 (instance members with default implementations, static interface members, and nested interface members such as delegates). SA1400 does not report diagnostics for these interface members.
```
AArnott: I agree with this point.
I wonder if we should have a rule to discourage adding redundant `public` access modifiers on interface members when they are all `public`. I've seen some of these crop up already, and IMO it is more confusing and maybe mis-educating other developers who read it into thinking it's important.

sharwell: ➡️ Would be good to file this as a separate request. It may already exist in the IDE/SDK code style rules.

PR #4036, opened 2025-11-28
Support static local functions from C# 8

### LINKED ISSUE #3005: Review impact of static local functions on existing rules
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#static-local-functions

sharwell:
Closes #3005

### DIFF
--- a/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/OrderingRules/SA1206CSharp8UnitTests.cs
+++ b/StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/OrderingRules/SA1206CSharp8UnitTests.cs
@@ -3,9 +3,70 @@
 
 namespace StyleCop.Analyzers.Test.CSharp8.OrderingRules
 {
+    using System.Threading;
+    using System.Threading.Tasks;
     using StyleCop.Analyzers.Test.CSharp7.OrderingRules;
+    using Xunit;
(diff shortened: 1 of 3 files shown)


StyleCop.Analyzers/StyleCop.Analyzers.Test.CSharp8/OrderingRules/SA1206CSharp8UnitTests.cs
```diff
@@ -3,9 +3,70 @@
...
+            var expected = Diagnostic().WithArguments("static", "async").WithLocation(0);
+
+            await VerifyCSharpFixAsync(testCode, expected, fixedCode, CancellationToken.None).ConfigureAwait(false);
+        }
+
+        [Fact]
+        [WorkItem(3005, "https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3005")]
+        public async Task TestStaticUnsafeLocalFunctionIncorrectOrderAsync()
+        {
+            var testCode = @"public class TestClass
```
AArnott: Have you considered using newer syntax for these new tests that adds syntax coloring and avoids any escaping requirements? e.g.

```cs
           var testCode = /* lang=c#-test */ """
               public class TestClass
               {
                 // ...
               }
               """;
```

sharwell: Yes, will convert over to that syntax after I finish work on the current language feature updates.

PR #4040, opened 2025-11-29
Add test project for c# 14

bjornhellander:

Issue #4028, opened 2025-11-26
Review impact of field-backed properties on existing rules


sharwell:
https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#the-field-keyword

EdwinEngelen:
This falsely triggers SA1500 and SA1513 on the closing brace.

Although this is a stupid example, usually there will be some expression in the getter and/or setter.

```cs
public bool Property1
{
    get { return field; }
    set { field = value; }
} = default;
```

PR #4077, opened 2025-12-12
Update for property initializer in C# 14

EdwinEngelen:
This falsely triggers SA1500 and SA1513 on the closing brace.
Although this is a stupid example, usually there will be some expression in the getter and/or setter.

```cs
public bool Property1
{
    get { return field; }
    set { field = value; }
} = default;
```

This problem has also been mentioned https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/4028#issuecomment-3642289439.
Depends on PR #4040.


sharwell:
Note that I probably won't be able to get to this until after the C# 14 audit is complete. So far I (recently) completed the C# 8 and 9 audits and am currently working on C# 10.

PR #4080, opened 2025-12-28
Update for null-forgiving operator at end of line

EdwinEngelen:
This falsely triggers SA1003 on the null-forgiving operator at the end of the line.

```cs
public void Foo(string? maybeNullString)
{
    var result = maybeNullString!
        .ToString();
}
```

Method chaining is a common practice and placing each method call on a separate line is clean. When a null-forgiving operator is applied, then it should be placed immediately after the expression that can be null. So that's at the end of the line, and not at the beginning of the next line.

Depends on PR #4077 (a stack of pull requests that together produce a NuGet package that solves my StyleCop problems).


sharwell:
Marking as draft for now due to dependency on #4077

PR #4083, opened 2026-04-07
Clarify how SA1518 is influenced by editorconfig in docs

dg-marbl:
To avoid confusion if using an .editorconfig and the default configuration of the analyzer configuration is not used. See in [code](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/513ba0fefdf66e86c030c00a73812322585358ae/StyleCop.Analyzers/StyleCop.Analyzers/Settings/ObjectModel/LayoutSettings.cs#L71)


dg-marbl:
Ups, I just saw that it will be already added here: #3374. I close the PR

